PR #5, opened 2023-05-29
Pico neo 3/4 support, microphone streaming, audio fixes, quest sdk update

Meister1593:
Updated ubuntu to 23 with gcc 13, fixed with a single import
Updated vaapi ffmpeg encoding to work with latest ffmpeg
Updated Oculus OpenXR Mobile SDK to v53.0
Updated monado to [e9475b13](https://gitlab.freedesktop.org/monado/monado/-/commit/e9475b13137db2cca8571576b5381a70fb8180a5) (by @Okabintaro)
Fixed UB in audio sewer (by @Okabintaro) causing buffer overflow
Fixed openxr non-comformance related to haptics using subactions afrer creation ([source](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XrHapticActionInfo)) causing spam on pico xr runtime (and no haptics at all)
Changed default bitrate to at least 50 mbps (10 mbps far too low to even see anything, and raising default bitrate seems reasonable)
Added support for Pico neo 3, Pico 4 with their latest openxr mobile sdk, 
Added support for microphone streaming by @galister
Added rudimentary input profile switches for supported headsets (also partially done by @galister)
Remove unnecessary systemd mentions/links (and simplify code) to allow building on non-systemd systems (by @galister)

Found one issue with VK_FORMAT_B8G8R8A8_UNORM when using monado async reprojection, made into todo, likely to be in one of the next PR's

Staying as draft now before testing on all headsets and thoroughly.


server/hostname.cpp
```diff
@@ -1,35 +1,13 @@
...
-			free(hostname);
-			sd_bus_unref(bus);
-			return s;
-		}
-	}
-
-	sd_bus_unref(bus);
-
-	return "Unknown";
+  char buf[256];
```
xytovl, line 8: Can we make the `gethostname`/dbus switch an ifdef?

PrettyHostname will allow non-latin characters and uppercase.

external/pico/libs/android.armeabi-v7a/libopenxr_loader.so
xytovl, line 1: I think we should move to git LFS, either in the pull request, or if it's complicated I can rewrite history after.

Meister1593: i will try moving to git lfs, if will be unsuccessful will say about that :+1:

Meister1593: hi, im back
there is one problem with git lfs though...
image

xytovl: I tried on a copy of your branch:
```
git lfs install
git rebase -i ...
git lfs track external/pico/libs/android.arm64-v8a/libopenxr_loader.so external/pico/libs/android.armeabi-v7a/libopenxr_loader.so
git commit --amend ...
```
It seems to work fine, what happens if you try to clone/pull https://github.com/Meumeu/WiVRn/tree/pico4 ?

Meister1593: libs seem to be there, but it is to be expected, i tried doing `git lfs migrate import --everything --include="*.so" --verbose`

xytovl: I don't know why your command failed initially. For the oculus libs, I can force push to master and rewrite all tags, or we just keep them here. Github won't let you rewrite history of an other repo.

Meister1593: Command did not fail, i couldn't push to my repo because of github thing

server/hostname.cpp
galister: there's now a separate `hostname.cpp` and `hostname-sd.cpp`, the latter will be included by cmake if systemd is available.

external/pico/libs/android.armeabi-v7a/libopenxr_loader.so
Meister1593: Sorry for long wait
I think all .so files are now in git lfs and tracked for future loaders as well

xytovl:
I see the PR is still marked as "draft", do we need anything else?

client/application.cpp
```diff
@@ -408,36 +409,81 @@ void application::initialize()
...
 	// TODO get it from application info
 	xr_actionset = xr::actionset(xr_instance, "actions", "Actions");
-	std::vector<XrActionSuggestedBinding> bindings;
 
-	for (const auto & [name, type]: oculus_touch)
-	{
-		std::string name_without_slashes = std::string(name).substr(1);
+	// TODO better way to handle bindings for different vendors/flavors of vendor
+	char property_chararr[PROP_VALUE_MAX + 1];
+	__system_property_get("ro.product.manufacturer", property_chararr);
```
xytovl, line 415: This, and the associated include, should be guarded by ifdefs. We try to support a PC version of the client, and this breaks compilation.

Meister1593: I added ifdefs

Meister1593: also made some fallback variable as well, just a placeholder for future

Meister1593:
I'm gonna also add slight monado upgrade (before some vulkan changes, mostly superficial changes)

Meister1593:
Forgot to ifdef import

Meister1593:
Okay it builds now locally and runs up to around determining bindings and failing somewhere further (which is i guess to be expected?)

Meister1593:
is it good for merge @xytovl ? builds are strangely failing but it builds locally fine

PR #15, opened 2023-11-16
Per device config

xytovl:


common/utils/wrap_lambda.h
Meumeu, line 1: Missing pragma once

client/hardware.h
Meumeu, line 1: Missing pragma once

PR #34, opened 2024-02-10
Add WIVRN_DUMP_VIDEO Runtime Configuration

damaestro:
- Enable dumping of the video stream to disk with runtime var

During testing it is valuable to be able to scribe the video stream to local disk. This might not be the best place to do the environment variable check and instead pass the result of the check into the function or some other configuration mechanism. I don't know how expensive the `std::getenv()` is and it would be in a critical path of performance.


xytovl:
I had https://github.com/Meumeu/WiVRn/commit/102f5ed6e9d1f33f719423a7ceccdf3fa348a350 on the encode-rewrite branch. I think it is a little better because we can configure the path, and it will overwrite if it already exists.

damaestro:
Tested this out cherry picked to the dev branch and it works as expected. Looks good to me.

PR #39, opened 2024-02-18
Only Sign APK if Secretes Exist

damaestro:
This allows for the github actions pipeline to successfully run in forks.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -115,8 +115,18 @@ jobs:
       run: >
         sudo apt-get update && sudo apt-get install glslang-tools librsvg2-bin
 
+    - name: Check for Secret availability
+      id: secret-check
+      shell: bash
+      run: |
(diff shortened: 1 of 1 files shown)


xytovl:
I think it would be better to check if secret is available: https://stackoverflow.com/questions/70249519/how-to-check-if-a-secret-variable-is-empty-in-if-conditional-github-actions

A fork could then setup secrets on their side and have a signed version as well.

.github/workflows/Build.yml
```diff
@@ -115,8 +115,18 @@ jobs:
       run: >
         sudo apt-get update && sudo apt-get install glslang-tools librsvg2-bin
 
+    - name: Check for Secret availability
+      id: secret-check
+      shell: bash
+      run: |
+        if [ "${{ secrets.APK_SIGNINGKEYPASSWORD }}" != '' && "${{ secrets.APK_KEYSTORE_BASE64 }}" != '']; then
```
xytovl, line 122: Missing whitespace before `]`, see https://github.com/Meumeu/WiVRn/actions/runs/7951605315/job/21705233593?pr=39

PR #43, opened 2024-02-28
skip ATW if only single stereo layer projected

mittorn:
Reprojection is already done on client, so may skip it on server, passing untouched eye poses from app

### DIFF
--- a/server/driver/wivrn_comp_target.cpp
+++ b/server/driver/wivrn_comp_target.cpp
@@ -526,12 +526,28 @@ static VkResult comp_wivrn_present(struct comp_target * ct,
 	view_info.display_time = cn->cnx->get_offset().to_headset(desired_present_time_ns).count();
 	for (int eye = 0; eye < 2; ++eye)
 	{
-		xrt_relation_chain xrc{};
-		xrt_space_relation result{};
-		m_relation_chain_push_pose_if_not_identity(&xrc, &cn->c->base.slot.poses[eye]);
-		m_relation_chain_resolve(&xrc, &result);
-		view_info.fov[eye] = xrt_cast(cn->c->base.slot.fovs[eye]);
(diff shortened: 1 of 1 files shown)


server/driver/wivrn_comp_target.cpp
```diff
@@ -526,12 +526,28 @@ static VkResult comp_wivrn_present(struct comp_target * ct,
...
+		view_info.fov[eye] = xrt_cast(slot.fovs[eye]);
+		view_info.pose[eye] = xrt_cast(slot.poses[eye]);
+		if (slot.layer_count == 1 && 
+			slot.layers[0].data.type == XRT_LAYER_STEREO_PROJECTION &&
+			slot.layers[0].data.stereo.r.fov.angle_left != 0.0f)
+		{
+			// only stereo projection is visible, may skip ATW
+			const auto &stereo = slot.layers[0].data.stereo;
+			view_info.pose[eye] = xrt_cast(eye ? stereo.r.pose : stereo.l.pose);
+			cn->c->debug.atw_off = true;
```
xytovl, line 539: This will be effective for the next image, there will be reprojection glitches on each single layer/multi layer switch.
if ATW is only effective for single layer images it should be always disabled.

xytovl: The only way I see to have the atw switch at the right place is to look at https://gitlab.freedesktop.org/monado/monado/-/blob/6876de1da6986544c6cd04b5bc5e703d6f492e00/src/xrt/compositor/main/comp_renderer.c?page=2#L1101, value is read in `dispatch_compute` and `comp_target_check_ready` is called on each frame before that.
Setting the ATW flag based on layer count in check_ready function should be right for every frame.

xytovl:
rebased and merge on dev with https://github.com/Meumeu/WiVRn/commit/999a1e0cf03c015ee38222ba4cb5af916ef726a8

PR #51, opened 2024-03-16
Enable PIE support, by default

damaestro:
Resolves https://github.com/Meumeu/WiVRn/issues/48

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,11 @@ endif()
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_CXX_STANDARD 20)
 
+# Default to PIC code
+include(CheckPIESupported)
+check_pie_supported()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -9,9 +9,20 @@ if(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.24)
     cmake_policy(SET CMP0135 NEW)
 endif()
 
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.14.0")
+	cmake_policy(SET CMP0083 NEW)
+	include(CheckPIESupported)
+	check_pie_supported()
+endif()
```
xytovl, line 16: file has
```cmake
cmake_minimum_required(VERSION 3.20)
```
According to https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings the policy will be set to `NEW`.

We should only need the other part of the diff.

PR #55, opened 2024-04-08
CMake: Add absolute runtime path option for XR manifest

Scrumplex:


Scrumplex:
I personally haven't tested this yet. I'll report back once I had some food! :D

xytovl:
Tested locally, no change when option is not set, which is fine enough for me.

Just add a description for the option, fix indentation (tabs in CMakeLists.txt, indent if content) and it's fine. Thanks for the contribution.

Scrumplex:
Made the changes you requested and tested the manifest with the option. LGTM!

PR #61, opened 2024-05-01
feat(client): resolution scale setting

galister:
I'm trying to add a simple select box to be able to adjust the resolution on the client.

Except for some reason an app restart is required. If connecting to a stream right after setting this, it'll create the stream swapchain with a res that's way off. Any tips on what that could be?

### DIFF
--- a/client/configuration.cpp
+++ b/client/configuration.cpp
@@ -117,6 +117,11 @@ configuration::configuration(const std::string & path)
 		preferred_refresh_rate = val.get_double();
 	}
 
+	if (auto val = root["resolution_scale"]; val.is_double())
+	{
+		resolution_scale = val.get_double();
+	}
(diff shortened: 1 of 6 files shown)


xytovl:
What do you mean by way off?

The swapchain should be 1.4 * resolution where resolution is the value provided in the info packet.

Starting from recommended width/height from headset we should have:
info_packet recommended eye width/height = override_view(model) * scale from your PR
video stream size = recommended * scale
swapchain = recommended * 1.4

galister:
ok, i think the issue then comes from the fact that i have to set server side scale to 0.5 in order for the projection to look acceptable, so the video stream is sized very differently from the swapchain

galister:
ok, this is good to go i think.

PR #79, opened 2024-06-20
add japanese translation

mitsuami-megane:
It builds but displays all �'s in the client. I think the font is missing the glyphs.

### DIFF
--- a/client/locale/ja/wivrn.po
+++ b/client/locale/ja/wivrn.po
@@ -1,0 +1,216 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: WiVRn 0.16\n"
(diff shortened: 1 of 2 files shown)


xytovl:
Thanks for the translations, there's some work in progress in https://github.com/Meumeu/WiVRn/tree/i18n to use a correct font and not require to manually set the glyph range.

mitsuami-megane:
i added my .po on top of i18n branch, but i'm just getting regular question marks now, not even �'s

Meumeu:
i18n branch should be working now, I've included your commit as well

mitsuami-megane:
confirmed, works great!

PR #82, opened 2024-06-27
update monado to 24.0.0

galister:
straightforward update, was testing this for the past couple of days with no issues.

mostly wanting it because of the added `MNDX_xdev_space` extension, which will be needed when experimenting with tracker support

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,7 +182,7 @@ FetchContent_Declare(imgui         EXCLUDE_FROM_ALL URL https://github.com/ocorn
 # Server dependencies
 FetchContent_Declare(monado
     GIT_REPOSITORY   https://gitlab.freedesktop.org/monado/monado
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -182,7 +182,7 @@ FetchContent_Declare(imgui         EXCLUDE_FROM_ALL URL https://github.com/ocorn
 # Server dependencies
 FetchContent_Declare(monado
     GIT_REPOSITORY   https://gitlab.freedesktop.org/monado/monado
-    GIT_TAG          598080453545c6bf313829e5780ffb7dde9b79dc
+    GIT_TAG          v24.0.0
```
xytovl, line 185: I think it is a bit better to use sha even for a tagged version, cmake won't have to access the remote if it already has the version

galister: good call, thanks.

PR #97, opened 2024-07-31
passthru EXT_eye_gaze_interaction

galister:
I thought it'd make more sense to do this first and then be able to calculate the foveation params on the server side.


server/driver/wivrn_session.cpp
```diff
@@ -141,7 +141,7 @@ xrt_result_t xrt::drivers::wivrn::wivrn_session::create_session(xrt::drivers::wi
 	auto & devices = *out_xsysd;
 	int n = 0;
 	if (self->hmd)
-		usysds->base.base.static_roles.head = devices->xdevs[n++] = self->hmd.get();
+		usysds->base.base.static_roles.head = usysds->base.base.static_roles.eyes = devices->xdevs[n++] = self->hmd.get();
```
xytovl, line 144: We should only set the role when tracking is supported

PR #98, opened 2024-08-01
update japanese locale

mitsuami-megane:


client/locale/ja/wivrn.po
```diff
@@ -211,6 +215,8 @@ msgid "Displayed"
 msgstr "表示"
 
 #: client/scenes/stream_stats.cpp:293
-#, c++-format
 msgid "Estimated motion to photons latency: {}ms"
 msgstr "「Motion-to-photon」推定遅延 {}ミリ秒"
+
+#~ msgid "Waiting for video stream"
+#~ msgstr "動画ストリームを待ち"
```
xytovl, line 222: This text was removed (there is no corresponding c++ line), you can drop it from .po

PR #103, opened 2024-08-10
expose battery on hmd xdev

galister:


server/driver/wivrn_session.cpp
```diff
@@ -304,6 +304,11 @@ void wivrn_session::operator()(from_headset::feedback && feedback)
 		dump_time("display", feedback.frame_index, o.from_headset(feedback.displayed), feedback.stream_index);
 }
 
+void wivrn_session::operator()(from_headset::battery && battery)
+{
+	hmd->update_battery(battery);
```
xytovl, line 309: We need some locking, this is called from network thread but reads are from monado thread

galister: just `std::lock_guard lock(mutex);` is good enough?

client/scenes/stream_tracking.cpp
```diff
@@ -220,6 +239,32 @@ void scenes::stream::tracking()
...
+							}
+
+							auto plugged_jint = intent.call<jni::Int>("getIntExtra", plugged_jstr, default_jint);
+							battery.charging = plugged_jint && plugged_jint.value > 0;
+						}
+
+						network_session->send_stream(battery);
+
+						next_battery_check = now + battery_check_interval;
+					}
```
xytovl, line 265: Is it possible to time/profile this block? It is unlikely to be an issue but the home-grown jni wrapper is likely slow.

We may also want to move that after the `busy_time` calculation or we may reduce polling rate if it takes more than 0.2ms.

server/driver/wivrn_session.cpp
xytovl: The comment should be in `update_battery`, use the mutex in `wivrn_hmd` and it's fine (for read too of course).

client/scenes/stream_tracking.cpp
galister: I'm getting an average of 0.8ms for the whole block. moved it after the busy time. let me see what else i can try.

xytovl: My guess is that `GetMethodID` inside `call` may be slow, a profiler would give more accurate data.

0.8ms every 5s is not bad but not entirely free. We should check if something can be done before merging.

common/wivrn_packets.h
```diff
@@ -218,7 +218,14 @@ struct feedback
 	uint8_t times_displayed;
 };
 
-using packets = std::variant<headset_info_packet, feedback, audio_data, handshake, tracking, hand_tracking, inputs, timesync_response>;
+struct battery
+{
+	float charge;
+	bool present;
+	bool charging;
+};
```
xytovl, line 226: Is there an API in Android to query controller battery as well? If we want to support it in the future, we will need a device id associated with the data and maybe bundle multiple statuses in one packet.

galister: the only way to do it is via the old oculusvr api afaik.

client/scenes/stream_tracking.cpp
galister: down to 0.5ms after caching the method id. i'm guessing it's the intent itself that takes long. should we look at moving this to a separate thread, or is it fine like this?

client/scenes/stream_tracking.cpp
```diff
@@ -138,6 +142,25 @@ void scenes::stream::tracking()
...
+	jni::string scale_jstr("scale");
+	jni::string plugged_jstr("plugged");
+	jni::Int default_jint(-1);
+
+	auto receiver_jobj = jni::object<"android/content/BroadcastReceiver">(NULL);
+	auto filter_jobj = jni::new_object<"android/content/IntentFilter">(filter_jstr);
+
+	auto register_receiver = jni::klass("android/content/Context")
+	                                 .method<jni::object<"android/content/Intent">>("registerReceiver", receiver_jobj, filter_jobj);
+	auto get_int_extra = jni::klass("android/content/Intent").method<jni::Int>("getIntExtra", level_jstr, default_jint);
```
galister, line 160: while this works, i think it would be more proper to pass the actual typenames, except i'm not sure how to do that.

client/scenes/stream_tracking.cpp
xytovl: I think it's ok like this. I'll try to profile it more accurately when I have time

xytovl:
If you did end-to-end testing and it's fine I can merge. I may try and optimize the remaining microseconds later.

galister:
yes, it's tested

PR #104, opened 2024-08-12
passthru FB_face_tracking2

galister:
FB_face_tracking2 support got merged into Monado just now.

We only implement the VISUAL tracking method, because that's the one that makes sense.

Tested with OscAvMgr + VRChat.

### DIFF
--- a/client/scenes/stream_tracking.cpp
+++ b/client/scenes/stream_tracking.cpp
@@ -197,12 +197,15 @@ void scenes::stream::tracking()
 			for (XrDuration Δt = 0; Δt <= prediction; Δt += std::max<XrDuration>(1, prediction))
 			{
 				from_headset::hand_tracking hands{};
+				from_headset::fb_face2 fb_face2{};
 
 				packet.production_timestamp = t0;
 				hands.production_timestamp = t0;
+				fb_face2.production_timestamp = t0;
 
 				packet.timestamp = t0 + Δt;
 				hands.timestamp = t0 + Δt;
+				fb_face2.timestamp = t0 + Δt;
 
 				try
 				{
@@ -240,6 +243,14 @@ void scenes::stream::tracking()
 						hands.joints = locate_hands(application::get_right_hand(), local_floor, hands.timestamp);
 						t.pause();
 						network_session->send_stream(hands);
+						t.resume();
+					}
+
+					if (application::get_fb_face_tracking2_supported())
+					{
+						application::get_fb_face_tracker2().get_weights(t0 + Δt, &fb_face2);
+						t.pause();
+						network_session->send_stream(fb_face2);
 						t.resume();
 					}
 
--- a/client/scenes/stream_tracking.cpp
+++ b/client/scenes/stream_tracking.cpp
@@ -246,36 +246,6 @@ void scenes::stream::tracking()
 					XrDuration busy_time = t.count();
 					// Target: polling between 1 and 5ms, with 20% busy time
 					tracking_period = std::clamp<XrDuration>(std::lerp(tracking_period, busy_time * 5, 0.2), 1'000'000, 5'000'000);
-
-#ifdef __ANDROID__
-					if (next_battery_check < now)
-					{
-						timer t2(instance);
-						from_headset::battery battery{};
-						auto intent = ctx.call<jni::object<"android/content/Intent">>(register_receiver, receiver_jobj, filter_jobj);
-						if (intent)
-						{
-							auto level_jint = intent.call<jni::Int>(get_int_extra, level_jstr, default_jint);
-							auto scale_jint = intent.call<jni::Int>(get_int_extra, scale_jstr, default_jint);
-
-							if (level_jint && level_jint.value >= 0 && scale_jint && scale_jint.value >= 0)
-							{
-								battery.present = true;
-								battery.charge = (float)(level_jint.value) / (float)(scale_jint.value);
-							}
-
-							auto plugged_jint = intent.call<jni::Int>(get_int_extra, plugged_jstr, default_jint);
-							battery.charging = plugged_jint && plugged_jint.value > 0;
-						}
-
-						network_session->send_stream(battery);
-
-						next_battery_check = now + battery_check_interval;
-						XrDuration battery_dur = t2.count();
-
-						spdlog::info("Battery check took: {}", battery_dur);
-					}
-#endif
 				}
 				catch (const std::system_error & e)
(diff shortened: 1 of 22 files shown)


common/wivrn_packets.h
```diff
@@ -185,6 +186,14 @@ struct hand_tracking
 	std::optional<std::array<pose, XR_HAND_JOINT_COUNT_EXT>> joints;
 };
 
+struct fb_face2
+{
+	std::array<float, 70> weights;
```
xytovl, line 191: We could use `XR_FACE_EXPRESSION2_COUNT_FB` if we make sure we have a recent enough OpenXR

client/scenes/stream_tracking.cpp
```diff
@@ -243,6 +244,12 @@ void scenes::stream::tracking()
 						t.resume();
 					}
 
+					if (application::get_fb_face_tracking2_supported())
+					{
+						application::get_fb_face_tracker2().get_weights(t0 + Δt, &fb_face2);
```
xytovl, line 249: Either we do the extrapolation + latest (Δt = 0 and Δt = prediction) and server side needs to use a history system to store/retrieve the values, or we only query one time.

What I understand from current code is that it will get 2 samples per loop, which might get reordered, then server only stores one.

client/scenes/stream_tracking.cpp
```diff
@@ -243,6 +244,12 @@ void scenes::stream::tracking()
...
+						network_session->send_stream(fb_face2);
```
xytovl, line 250: Timer should be paused when sending, I guess manually doing this everytime doesn't scale, we can make a macro so that future additions won't miss it.

Rationale: socket sometimes block (queue full? radio error?), and mess up time measurement. On some runtimes, locating spaces is heavy on CPU time and it is not possible to run the loop at 1kHz rate. Lowering the sampling rate helps with thermals and is a bit better. We don't want to lower the rate for occasional network spikes.

server/driver/wivrn_hmd.cpp
```diff
@@ -274,6 +292,16 @@ void wivrn_hmd::update_battery(const from_headset::battery & new_battery)
...
+void wivrn_hmd::update_fb_face2(const from_headset::fb_face2 & new_face)
+{
+	memcpy(&face.face_expression_set2_fb.weights, new_face.weights.data(), sizeof(float) * new_face.weights.size());
+	memcpy(&face.face_expression_set2_fb.confidences, new_face.confidences.data(), sizeof(float) * new_face.confidences.size());
+	face.face_expression_set2_fb.is_valid = new_face.is_valid;
+	face.face_expression_set2_fb.is_eye_following_blendshapes_valid = new_face.is_eye_following_blendshapes_valid;
+	face.face_expression_set2_fb.data_source = XRT_FACE_TRACKING_DATA_SOURCE2_VISUAL_FB;
+	face.face_expression_set2_fb.sample_time_ns = os_monotonic_get_ns();
+}
+
```
xytovl, line 304: This is not thread safe, and considering the comment on the prediction time, I think deriving from `history` could be better.

client/scenes/stream_tracking.cpp
galister: We only get 1 sample and store that. But maybe extrapolating on the server side would make sense.

galister:
I think the Wireshark pipeline's missing the openxr header.

Either way, I'll look at using history, but I've a busy week so it might take some time.

galister:
nvm, libopenxr-dev in jammy is just too old

client/scenes/stream_tracking.cpp
xytovl: `get_weights` is inside the `for (XrDuration Δt = 0; Δt <= prediction; Δt += std::max<XrDuration>(1, prediction))` loop, it will be executed twice

galister: so i guess the battery check also needs to be moved out?

galister:
Since last review:
- I moved the battery check outside the predict loop
- I rewrote the face tracker to use history
- Face tracker is now its own device
- It works and it's happy

However, I will need help with:
- Getting an up-to-date `openxr.h` into the Wireshark dissector
- Figure out why the XRT defines don't work in the Flatpak action

server/driver/weights_list.h
```diff
@@ -0,0 +1,59 @@
...
+		std::array<float, NumWeights> result = t < ta ? a : b;
+		for (size_t i = 0; i < NumWeights; i++)
+		{
+			auto vel = (b[i] - a[i]) / h;
+			float dt = (t - tb) / 1.e9;
+
+			result[i] = std::clamp(result[i] + vel * dt, 0.0f, 1.0f);
+		}
+		return result;
+	}
```
xytovl, line 53: It should be possible to not implement this by replacing  the `if (extrapolate)` with `if constexpr (extrapolate)`.

I don't know how much sense it makes to extrapolate expressions.

galister: it's not being used right now as the underlying history has extrapolate off, but I implemented it in case it would be used for something else that needs extrapolation

server/driver/wivrn_fb_face2_tracker.cpp
```diff
@@ -0,0 +1,150 @@
...
+		if (!is_valid)
+		{
+			return XRT_SUCCESS;
+		}
+
+		out_value->face_expression_set2_fb.is_eye_following_blendshapes_valid = is_eye_following_blendshapes_valid;
+
+		XrTime predicted_time = ((XrTime)now) + last_prediction_offset;
+
+		out_value->face_expression_set2_fb.sample_time_ns = predicted_time;
```
xytovl, line 114: The sample_time could be stored as data with the weights list, where interpolate/extrapolate would likely take the max value. No need to store it out of the history object.

I don't if there is a reason to split weights and confidences

galister: the idea was to keep weights_list generic enough so that it could be used for a different face tracking extension down the line.

xytovl:
> However, I will need help with:
> 
> * Getting an up-to-date `openxr.h` into the Wireshark dissector check

I have pushed a commit with the cmake rules for dissector

server/driver/wivrn_fb_face2_tracker.cpp
xytovl: If you think it makes sense to have 2 objects, it's not a big problem, but the sample time should be part of the data

PR #107, opened 2024-08-18
Add New APK Only Pipeline

damaestro:
- Automatically sync downstream if build succeeds on a commit to master
- Downstream build APK only and publish as a release
- Use deterministic name for programmatic access

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -16,6 +16,7 @@ jobs:
     # You can convert this to a matrix build if you need cross-platform coverage.
     # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
     runs-on: ubuntu-22.04
+    if: ${{ vars.APK_ONLY == '' }}
 
     steps:
     - uses: actions/checkout@v4
@@ -77,6 +78,7 @@ jobs:
   build-server-flatpak:
     name: Linux server (flatpak)
     runs-on: ubuntu-latest
+    if: ${{ vars.APK_ONLY == '' }}
     container:
       image: bilelmoussaoui/flatpak-github-actions:freedesktop-23.08
       options: --privileged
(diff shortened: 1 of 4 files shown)


damaestro:
This can be merged without the full setup completed as everything should skip when the correct github actions configuration is not present.

.github/workflows/Build.yml
```diff
@@ -228,19 +233,21 @@ jobs:
       with:
         name: apk-${{ matrix.hmd }}-${{ (github.event_name == 'pull_request' && matrix.buildtype == 'Release') && 'Release-unsigned' || matrix.buildtype }}
         path: apk/*.apk
+        retention-days: 30
```
damaestro, line 236: Added to save on storage costs.

.github/workflows/Build.yml
```diff
@@ -228,19 +233,21 @@ jobs:
...
 
     - name: Archive debug symbols
       if: ${{ matrix.buildtype != 'Release' }}
       uses: actions/upload-artifact@v4
       with:
         name: symbols-${{ matrix.hmd }}-${{ matrix.buildtype }}
         path: debug/**/*.so
+        retention-days: 30
```
damaestro, line 244: Added to save on storage costs.

damaestro:
@xytovl @Meumeu to finish the setup after this is merged we need to create the PAT that has specific granular permissions to call the sync and to create the release. You could use the legacy token type but I recommend the granular.

An org admin needs to go to https://github.com/settings/personal-access-tokens/new and create a new granular token with all of the needed permissions for the downstream repo:

* Owned by the `WiVRn` org
* Only select repositories -> `WiVRn/WiVRn-APK`

Repository permissions:

 **Read access** to metadata

 **Read and Write access** to actions, code, commit statuses, issues, merge queues, pull requests, repository custom properties, repository hooks, and workflows

Once that token is correctly configured, the `DOWNSTREAM_GITHUB_TOKEN` repository secret needs to be configured here https://github.com/WiVRn/WiVRn-APK/settings/secrets/actions and here https://github.com/WiVRn/WiVRn/settings/secrets/actions

The `APK_SIGNINGKEYPASSWORD` and `APK_KEYSTORE_BASE64` repository secrets also have to be configured to be able to create the signed APK.

PR #117, opened 2024-08-24
eye-tracked foveation

galister:
Some considerations:
- We can either apply the foveation early, for example in `comp_wivrn_flush` this means less delay on the foveation, but we must wait for the fence in-place.
- We can apply it at the end of `comp_wivrn_present`, which will effectively apply it for the next frame. The advantage here is that it's after the busy period, but there will be the added latency of one extra frame. In this case, the fence wait can be in `comp_wivrn_flush`.

During testing, waiting the extra frame was not noticeable for me, but I can imagine it might be for some people. I went with the first option, eventually.

I included a downstream patch that makes Monado only sample the red distortion image, so that we don't have to spend resources maintaining all 3.


galister:
Just tested on Pico 4 as well and behaves as expected without eye tracking.

patches/monado/0006-distortion-images.patch
```diff
@@ -0,0 +1,28 @@
...
+--- a/src/xrt/compositor/shaders/distortion.comp
++++ b/src/xrt/compositor/shaders/distortion.comp
+@@ -118,8 +118,8 @@ void main()
+ 	vec2 dist_uv = position_to_uv(extent, ix, iy);
+ 
+ 	vec2 r_uv = texture(distortion[iz + 0], dist_uv).xy;
+-	vec2 g_uv = texture(distortion[iz + 2], dist_uv).xy;
+-	vec2 b_uv = texture(distortion[iz + 4], dist_uv).xy;
++	vec2 g_uv = r_uv;
++	vec2 b_uv = r_uv;
```
xytovl, line 25: I have doubts the compiler is smart enough to do the `transform_uv` only once as well as `texture(source[iz], r_uv).r,`.

If we only read `r` channel, let's drop the g,b variables from the rest of the shader as well.

server/driver/wivrn_comp_target.cpp
```diff
@@ -596,6 +610,25 @@ void wivrn_comp_target::reset_encoders()
...
+	};
+
+	foveation_renderer->render_distortion_images(foveation, c->nr.distortion.images, c->nr.distortion.image_views);
+
+	wivrn_bundle->device.resetFences(*foveation_renderer->fence);
+	{
+		scoped_lock lock(c->base.vk.queue_mutex);
+		wivrn_bundle->queue.submit(submit_info, *foveation_renderer->fence);
+	}
+	auto res = wivrn_bundle->device.waitForFences(*foveation_renderer->fence, true, UINT64_MAX);
```
xytovl, line 629: Isn't there a way to include this in the command buffer used by monado? We introduce a GPU-CPU synchronization point

server/driver/wivrn_foveation.cpp
```diff
@@ -0,0 +1,279 @@
...
+		                                     .layerCount = 1},
+		        },
+		};
+		cmd_buf.pipelineBarrier(
+		        vk::PipelineStageFlagBits::eTopOfPipe,
+		        vk::PipelineStageFlagBits::eComputeShader,
+		        {},
+		        nullptr,
+		        nullptr,
+		        im_barriers0);
```
xytovl, line 164: I am not sure if the driver is smart enough to do this efficiently or if we should have barriers out of the loop, but with both images

server/driver/wivrn_foveation.h
```diff
@@ -0,0 +1,60 @@
...
+	std::array<xrt_vec2, 2> center_offset = {};
+	std::array<from_headset::tracking::view, 2> views = {};
+	xrt_quat last_gaze = {};
+
+public:
+	wivrn_foveation() {}
+
+	void set_initial_parameters(std::array<to_headset::foveation_parameter, 2> p);
+	void update_tracking(const from_headset::tracking &, const clock_offset &);
+	std::array<xrt_vec2, 2> get_center();
```
xytovl, line 39: Those functions will be called in network and monado threads, they need to be thread-safe

server/shaders/foveate.comp.glsl
```diff
@@ -0,0 +1,61 @@
...
+
+layout(push_constant) uniform PushConstants
+{
+    vec2 a;
+    vec2 b;
+    vec2 scale;
+    vec2 center;
+} pcs;
+
+const float one_over_dim = 1.0 / 127.0;
```
xytovl, line 33: Can we have the dimension as a specialization constant and sync it with monado's image dimensions?

server/driver/wivrn_foveation.cpp
```diff
@@ -0,0 +1,279 @@
...
+		        .a = {foveation.x.a, foveation.y.a},
+		        .b = {foveation.x.b, foveation.y.b},
+		        .scale = {foveation.x.scale, foveation.y.scale},
+		        .center = {foveation.x.center, foveation.y.center},
+		};
+
+		cmd_buf.bindPipeline(vk::PipelineBindPoint::eCompute, *pipeline);
+		cmd_buf.bindDescriptorSets(vk::PipelineBindPoint::eCompute, *layout, 0, ds[eye], {});
+		cmd_buf.pushConstants<FoveationParamsPcs>(*layout, vk::ShaderStageFlagBits::eCompute, 0, f_params);
+		cmd_buf.dispatch(16, 16, 1);
```
xytovl, line 178: The 16s here are 128 (from image size constant in monado's code) / 8 (from shader local size). Can we use the constant from monado?
For the 8 from shader, as long as it's under WiVRn's control I don't care if it's a magic value.

galister: i couldn't include `render/render_interface.h` because then it complains about:

`In included file: unknown type name 'PFN_vkCreateXcbSurfaceKHR'; did you mean 'PFN_vkDestroySurfaceKHR'?`

galister: in the shader I can do 
`float one_over_dim = 1.0 / float((gl_WorkGroupSize.x * gl_NumWorkGroups.x) - 1);`
but then it won't be a compile-time const

server/driver/wivrn_comp_target.cpp
galister: i ended up keeping the separate cmd buf, but dropped the fence

server/driver/wivrn_foveation.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+#include "driver/xrt_cast.h"
+#include "math/m_api.h"
+#include "utils/wivrn_vk_bundle.h"
+#include "wivrn_packets.h"
+#include "xrt/xrt_defines.h"
+#include <array>
+#include <cmath>
+
+#include <cmath>
```
xytovl, line 31: header is present twice

PR #119, opened 2024-09-02
only rely on history if clock_offset is stable

galister:
prevents extrapolation errors and abrupt jumps in tracking that often happen as a result of reconnecting or network lag spikes

### DIFF
--- a/server/driver/clock_offset.cpp
+++ b/server/driver/clock_offset.cpp
@@ -58,7 +58,7 @@ void clock_offset_estimator::add_sample(const xrt::drivers::wivrn::from_headset:
 	}
 	else
 	{
-		sample_interval = std::chrono::seconds(1);
+		sample_interval = std::chrono::milliseconds(100);
 		int64_t latency = 0;
 		for (const auto & s: samples)
 			latency += s.received - s.query;
@@ -88,12 +88,11 @@ void clock_offset_estimator::add_sample(const xrt::drivers::wivrn::from_headset:
 		x0 += (s.query + s.received) * 0.5;
 		y0 += s.response;
 	}
-	x0 /= n;
-	y0 /= n;
+	x0 *= inv_n;
+	y0 *= inv_n;
 
 	if (samples.size() < num_samples)
 	{
-		offset.a = 1;
 		offset.b = y0 - x0;
 		return;
 	}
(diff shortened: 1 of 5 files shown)


server/driver/history.h
```diff
@@ -47,6 +48,15 @@ class history
 		// Discard outdated data, packets could be reordered
 		if (not data.empty())
 		{
+			// keep only one sample if the clock_offset is unstable
+			if (!offset.is_stable())
+			{
+				U_LOG_D("not using history: clock_offset not stable");
+				data.clear();
+				data.emplace_back(sample, produced, t);
+				return;
```
galister, line 57: need to find a way so that the t and Δt samples don't get mixed

server/driver/clock_offset.cpp
```diff
@@ -146,3 +151,8 @@ XrTime clock_offset::to_headset(XrTime timestamp_ns) const
 {
 	return a * timestamp_ns + b;
 }
+
+bool clock_offset::is_stable() const
+{
+	return stable;
+}
```
xytovl, line 158: We already have an `operator bool()`.
We can change `clock_offset.cpp` to check for consistency before making it valid.

server/driver/history.h
```diff
@@ -47,6 +48,19 @@ class history
 		// Discard outdated data, packets could be reordered
 		if (not data.empty())
 		{
+			// keep only one sample if the clock_offset is unreliable
+			if (not offset)
```
xytovl, line 52: All the paths that lead to `add_sample` check if clock offset is valid in `wivrn_session`.

We should never reach this place, if we want to keep a single sample with unstable clock, we should remove the checks.

galister:
I've been trying a different approach - detecting in add_sample if the sample was produced in the future. 

Except it turns out that it's really common for produced_timestamp to be in the future. I would think this is a problem.

Upon first connection, the produced_timestamp is in the past as expected, right from the start.

Upon reconnect, this is what I noted:
- Starts with produced_timestamp in the future, but then it stabilizes right before the video first appears. 
  - This I would think where the offset estimator finished its first pass.
- Video first appears and at the same time, produced_timestamp is now for some reason hundreds of ms in the future
- Over the next seconds, it can keep going further into the future, then after 10~15s, top out and start coming back. I've seen it top out while being at 600-700ms into the future.
- From the moment of reconnect, it can take 30-60s for it to stabilize so that produced_timestamp is no longer in the future.

There were cases when produced_timestamp started in the past, then blasted off into future and came back again, within a similar time frame.

I think that the video stream kicking in could be throwing the offset estimator off.

Possible points for follow-up investigation:
- Temporarily turn off offset estimator for stream initialization phase
- Investigate the huge lag spike that happens on reconnect, but not on first connect
- Make offset estimator less susceptible to outlier values

PR #121, opened 2024-09-04
stage tracking for client + tweaks

galister:
The stage tracking is the larger commit, probably look at that one separately. the others are straightforward.

I changed it to re-center Monado stage instead of tracking origin offset. this preserves calibration between different tracking origins.

From my testing, recenter action seems indistinguishable from the previous method; you will be standing in the center and facing towards the game world's forward direction. Floor level is not touched.

I also switched it to monitor local space only, as to my knowledge, local only changes when recentering, while stage/local_floor may change when the floor level changes.

### DIFF
--- a/client/scenes/stream.h
+++ b/client/scenes/stream.h
@@ -78,10 +78,9 @@ private:
 	std::unique_ptr<wivrn_session> network_session;
 	std::atomic<bool> exiting = false;
 	std::thread network_thread;
-	std::mutex local_floor_mutex;
-	xr::space local_floor;
 	std::mutex tracking_control_mutex;
 	to_headset::tracking_control tracking_control;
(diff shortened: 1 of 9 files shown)


client/scenes/stream.h
```diff
@@ -78,10 +78,10 @@ class stream : public scene_impl<stream>, public std::enable_shared_from_this<st
 	std::unique_ptr<wivrn_session> network_session;
 	std::atomic<bool> exiting = false;
 	std::thread network_thread;
-	std::mutex local_floor_mutex;
-	xr::space local_floor;
 	std::mutex tracking_control_mutex;
 	to_headset::tracking_control tracking_control;
+	xr::space stage;
```
xytovl, line 83: As this space no longer changes, we can use `application::world_space`

PR #122, opened 2024-09-05
steamvr lh support

galister:


server/driver/wivrn_session.cpp
```diff
@@ -162,13 +163,44 @@ xrt_result_t xrt::drivers::wivrn::wivrn_session::create_session(xrt::drivers::wi
...
 		devices->xdevs[n++] = self->left_hand.get();
 		usysds->base.base.static_roles.hand_tracking.left = self->left_hand.get();
 	}
-	if (self->right_hand)
+	if (self->right_hand && !active_right_hand)
 	{
+		active_right_hand = self->right_hand.get();
 		devices->xdevs[n++] = self->right_hand.get();
 		usysds->base.base.static_roles.hand_tracking.right = self->right_hand.get();
 	}
```
galister, line 206: we will eventually want to expose hand tracking, except nowadays OC completely breaks down when using WiVRn hand tracking + lighthouse controller.

server/driver/wivrn_session.cpp
```diff
@@ -162,13 +167,44 @@ xrt_result_t xrt::drivers::wivrn::wivrn_session::create_session(xrt::drivers::wi
...
+	xrt_device * active_left_hand = nullptr;
+	xrt_device * active_right_hand = nullptr;
+
+#ifdef WIVRN_FEATURE_STEAMVR_LIGHTHOUSE
+	auto use_steamvr_lh = std::getenv("WIVRN_USE_STEAMVR_LH");
+	xrt_system_devices * lhdevs = NULL;
+
+	if (use_steamvr_lh && steamvr_lh_create_devices(&lhdevs) == XRT_SUCCESS)
+	{
+		U_LOG_W("Found %lu lighthouse devices:", lhdevs->xdev_count);
```
xytovl, line 179: This should be info level

galister: just removed it, Monado logs them as well

galister:
anything else for this?

xytovl:
I could not test the cmake part, if it works fine for you we can merge. Maybe squash the cmake commit with the previous one.

galister:
Tested and works.

Also here are the instructions for Envision (though envision refuses to work without systemd, so I couldn't test)

In WiVRn Profile, set:
- XR Service CMake Flags:
  - `WIVRN_FEATURE_STEAMVR_LIGHTHOUSE=ON`
- Environment Variables:
  - `WIVRN_USE_STEAMVR_LH=1`
  - `LH_DISCOVER_WAIT_MS=6000`

PR #129, opened 2024-09-12
update monado

galister:
- `EXT_hand_tracking` is now disabled by default, so added `XRT_FEATURE_OPENXR_HAND_TRACKING_EXT=ON`
- removed `0007-pass-requested-time-for-FB_face_tracking2.patch` (merged upstream)
- changed parts of `0008-Use-mipmaps-for-distortion-shader.patch` to avoid nesting. hopefully will be less prone to conflicts this way.
- changed server/driver timestamps `uint64_t` → `int64_t`

tested and seems to not explode


patches/monado/0008-Use-mipmaps-for-distortion-shader.patch
```diff
@@ -1794,32 +1763,32 @@ index 1b295a324..e29b79b03 100644
...
++		    &data,                         // data
++		    &world_poses[i],               // world_pose
++		    &eye_poses[i],                 // eye_pose
++		    &fovs[i],                      // fov
++		    rsci_rtr,                      // rtr
++		    &layer_viewport_data,          // layer_viewport_data
++		    &layer_norm_rect,              // layer_norm_rect
++		    rsci->image,                   // image
++		    rsci->srgb_view,               // srgb_view
++		    +scratch_view->info.mip_count, // mip_count
```
xytovl, line 1776: The `+` should not be there

PR #133, opened 2024-09-18
add wivrn+tcp scheme

galister:


client/main.cpp
```diff
@@ -67,6 +67,7 @@ void real_main()
 			        .manual = true,
 			        .visible = true,
 			        .compatible = true,
+			        .tcp_only = app.get_server_tcp_only(),
```
galister, line 70: not sure if this would be better to have under `service`

xytovl: It would be better, as we have to pass both to establish a connection

client/configuration.cpp
```diff
@@ -103,6 +103,7 @@ configuration::configuration(const std::string & path)
 		                .name = (std::string)i["pretty_name"],
 		                .hostname = (std::string)i["hostname"],
 		                .port = (int)i["port"].get_int64(),
+		                .tcp_only = i["tcp_only"].get_bool(),
```
xytovl, line 106: We should accept entires where the bool is not present in case of update from a previous version and server was stored because it had autoconnect property.

galister: do we actually want to store this, actually?

the client has no way of knowing if the server is tcp-only, other than from the intent. and the intent being there doesn't mean that the server only works with tcp

xytovl: Right, then don't store it and initialize to false.

xytovl: The address from intent gets stored, it is safer to store tcp_only property as well.

galister: updated to `is_bool() && get_bool()`

galister: does that work, or do i have to check the key?

xytovl: it should work, that's what we do for other fields

xytovl:
client/wivrn_discover.cpp creates a `wivrn_discover::service` doesn't initialize every field. We should set tcp_only to false by default.

PR #151, opened 2024-10-15
Create Italian Translation

albanobattistella:


client/locale/it.po
```diff
@@ -0,0 +1,237 @@
...
+msgstr "Visualizzato"
+
+#: client/scenes/stream_stats.cpp:300
+msgid "Predicted"
+msgstr "Previsto"
+
+#: client/scenes/stream_stats.cpp:318
+#, c++-format
+msgid "Estimated motion to photons latency: {}ms"
+msgstr "Latenza stimata del movimento dei fotoni: {}ms"
```
xytovl, line 237: It is the time between user motion to optical display. photons themselves don't have a latency.

Direi "Latenza stimata da movimento a fotoni: {}ms"

client/locale/it.po
Meumeu, line 1: This file must be called `client/locale/it/wivrn.po` instead of `client/locale/it.po`

PR #162, opened 2024-10-31
Add Vulkan Video encoder to dashboard settings

ImSapphire:


dashboard/settings.ui
```diff
@@ -57,7 +57,7 @@
        <item row="11" column="2">
         <widget class="QRadioButton" name="radio_manual_foveation">
          <property name="text">
-          <string comment="Foveation">Manual</string>
+          <string comment="Foveation">Manua&amp;l</string>
```
xytovl, line 60: this (and other similar changes) should not be part of the commit

Meumeu:
You should probably add a warning in `settings::on_encoder_settings_changed` like we did for av1, for example:
```
	if (encoder == 4 /* vulkan */)
	{
		status += tr("Vulkan Video encoder is experimental\n");
	}
```

PR #170, opened 2024-11-05
Spanish translation

Respuit:


xytovl:
Thank you for the contribution, I see that the mail in the commit and "last-translator" field are different, is that intended?

Respuit:
> Thank you for the contribution, I see that the mail in the commit and "last-translator" field are different, is that intended?

Ah yes, old email on the commit because of old account and new one on the "last-translator" field (I'm still moving things to the new one)

PR #177, opened 2024-11-11
Add initial controller offset values for Pico 4

### LINKED ISSUE #94: pico controller position offset
Only just started with wivrn and it is great, much smoother experience doing xr dev in godot using wivrn instead of alvr + steamvr.

One thing I noticed though is that the pico 4 controllers are offset compared to where they should be, the video should show it. It starts in wivrn and then I pres the "home" button and it shows where they are according to the pico OS home screen.

https://github.com/user-attachments/assets/4fb53c8f-b41e-40ce-9577-92ff3df8a0a7

Mhowser:
This should resolve #94 
we still need to adjust the distance between the controllers.

### DIFF
--- a/client/hardware.cpp
+++ b/client/hardware.cpp
@@ -265,6 +265,16 @@ std::pair<glm::vec3, glm::quat> controller_offset(std::string_view profile, xr::
 			default:
 				break;
 		}
+	else if (profile == "pico-4")
+		switch (space)
+		{
+			case xr::spaces::grip_left:
+			case xr::spaces::grip_right:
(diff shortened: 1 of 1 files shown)


xytovl:
just fix the formatting and it's perfect

Mhowser:
I'm not sure where the formatting problem is, and to be honest I don't know how to code in general, I just reused some the Oculus controller offsets and then filled in the stuff that is Pico related. I have edits allowed for maintainers on so I would appreciate if you could fix the formatting on my behalf :)

Mhowser:
Thank you!

PR #184, opened 2024-11-22
WIVRN_FORCE_EXTENSIONS for force initializing face & eye tracking

galister:
Addresses the incredibly niece use case of starting WiVRn with a non-face-tracked headset, then reconnecting with a face tracked headset later.


xytovl:
Can we make it generic: `WIVRN_FORCE_EXTENSIONS=XR_FB_face_tracking2`?

PR #200, opened 2024-12-04
support building wivrnctl with elogind

galister:
not sure if you consider this proper enough, it does build and work fine with everything else having WIVRN_USE_SYSTEMD off


galister:
ok that also does work

PR #202, opened 2024-12-05
do not attempt to call begin_feature

galister:


server/driver/wivrn_session.cpp
```diff
@@ -542,37 +542,27 @@ xrt_result_t wivrn_session::get_roles(xrt_system_roles * out_roles)
 
 xrt_result_t wivrn_session::feature_inc(xrt_device_feature_type type)
 {
-	if (type >= XRT_DEVICE_FEATURE_MAX_ENUM)
-		return XRT_ERROR_FEATURE_NOT_SUPPORTED;
-
-	// If it wasn't zero nothing to do.
-	if (!xrt_reference_inc_and_was_zero(&feature_use[type]))
```
xytovl, line 549: Can you also remove the now unused `feature_use` array?

galister:
without patch it crashes when launching oscavmgr on 1st try, with patch i couldn't reproduce in 15 attempts

PR #233, opened 2025-01-11
Replace resolution dropdown with a slider; Increase limit to 350%

Insprill:
As the title says. The increased range now matches SteamVR settings, and the slider allows for more granular control while avoiding a huge dropdown list.

The primary motivation for this is to play VTOL VR at 300% resolution, as I do on Windows with Steam Link.

### DIFF
--- a/client/scenes/lobby_gui.cpp
+++ b/client/scenes/lobby_gui.cpp
@@ -608,7 +608,7 @@ void scenes::lobby::gui_settings()
 		const auto width = stream_view.recommendedImageRectWidth;
 		const auto height = stream_view.recommendedImageRectHeight;
 		auto intScale = int(current * 100);
-		if (ImGui::SliderInt(_S("Resolution scale"), &intScale, 50, 500, fmt::format(_F("%d%% - {}x{} per eye"), int(width * current), int(height * current)).c_str()))
+		if (ImGui::SliderInt(_S("Resolution scale"), &intScale, 50, 350, fmt::format(_F("%d%% - {}x{} per eye"), int(width * current), int(height * current)).c_str()))
 		{
 			config.resolution_scale = intScale * 0.01;
 			config.save();
--- a/client/scenes/lobby_gui.cpp
+++ b/client/scenes/lobby_gui.cpp
@@ -607,17 +607,11 @@ void scenes::lobby::gui_settings()
 		const auto current = config.resolution_scale;
 		const auto width = stream_view.recommendedImageRectWidth;
 		const auto height = stream_view.recommendedImageRectHeight;
-		if (ImGui::BeginCombo(_S("Resolution scale"), fmt::format(_F("{}% - {}x{} per eye"), int(current * 100), int(width * current), int(height * current)).c_str()))
+		auto intScale = int(current * 100);
+		if (ImGui::SliderInt(_S("Resolution scale"), &intScale, 50, 500, fmt::format(_F("%d%% - {}x{} per eye"), int(width * current), int(height * current)).c_str()))
 		{
-			for (int scale = 50; scale <= 200; scale += 10)
-			{
-				if (ImGui::Selectable(fmt::format(_F("{}% - {}x{} per eye"), scale, (width * scale) / 100, (height * scale) / 100).c_str(), scale == current, ImGuiSelectableFlags_SelectOnRelease) and scale != current)
-				{
-					config.resolution_scale = scale * 0.01;
-					config.save();
-				}
-			}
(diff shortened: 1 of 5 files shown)


galister:
I've noticed that SteamVR uses arbitrary values for 100% and the scale values from Steam Link do not necessarily translate to the WiVRn client res.

At 300% client res we're talking about 6192x6624 per-eye on the Quest 3. Is your headset able to handle that?

What kind of actual headset performance are you getting with the 300% setting?

500% is definitely a value that I believe no actual headset will be able to do in the foreseeable future.

galister:
As a side note, I'm on Quest Pro and the headset is unable to maintain 90fps at 200% client res scale & 0.35 server scale.

Insprill:
I did forget about Steam's made-up resolution percentages. My 100% is headset resolution, so I thought that meant it would match up, but looking at the numbers, it does not.

When going this high in resolution, I don't think it's expected to get more than 72 fps. My Quest 3 can maintain 72fps up until around 290% (5985x6269) at 0.5 server scale. At 0.35 server scale, it can do around 340% (7017x7350). Tested with 3x VAAPI H264 encoders.

I think setting the limit to 400% would be fair, or we could limit it to stay under maxImageRectWidth/Height which is ~378% on the Quest 3 (it does work higher than that, just not well). Thoughts?

xytovl:
maxImageRectWidth/Height is supposed to be a hard limit, however at least HTC and Pico are able to render larger images. This is why it only shows a warning instead of setting a hard limit.

The 100% mark is defined in hardware.cpp to match the panel definition, the recommended width/height seem to be tuned for local rendering and are much lower resolution. Other systems may use the recommended size as their 100%.

The slider instead of dropdown looks like a good idea, the concern is on the range and how easy it will be to set too high values. Maybe a warning above a certain threshold?

Insprill:
The problem I see would be where to set that warning since the practical limit depends on how the server is configured, the target refresh rate, and which headset it's running on. What are the maximum image sizes for the HTC/Pico? If they're as high as the Quest 3's (8192x8192), I don't see a case where anyone could go higher than that and still have usable performance, so it'd make a reasonable limit.

xytovl:
Tested on HTC XR elite:
500% (9600x9600) fails to create encoders, maximum dimension is 4096
421% (8083x8083) crashes the headset app
303% (5817x5817) starts but is extremely slow
225% (320x4320) works, but is not very smooth.

The hard limit should be somewhere near 300%, the performance warning at about 200%.

Side note: picking accurate numbers with hand tracking is not easy.

client/scenes/lobby_gui.cpp
```diff
@@ -607,17 +607,11 @@ void scenes::lobby::gui_settings()
 		const auto current = config.resolution_scale;
 		const auto width = stream_view.recommendedImageRectWidth;
 		const auto height = stream_view.recommendedImageRectHeight;
-		if (ImGui::BeginCombo(_S("Resolution scale"), fmt::format(_F("{}% - {}x{} per eye"), int(current * 100), int(width * current), int(height * current)).c_str()))
+		auto intScale = int(current * 100);
+		if (ImGui::SliderInt(_S("Resolution scale"), &intScale, 50, 500, fmt::format(_F("%d%% - {}x{} per eye"), int(width * current), int(height * current)).c_str()))
```
xytovl, line 611: Can you update the existing translations for the "{}" -> "%d" change

Insprill: Do you mean here or in other places? The %d gets formatted by ImGui, our _F function doesn't touch it.

xytovl: In client/locale there are translations, checked by https://github.com/WiVRn/WiVRn/actions/runs/12719300523/job/35467616628?pr=233

You just have to replace the key (msgid) and values (msgstr) in wivrn.po files.

Insprill:
How about a hard limit of 350%? There could be cases where you'd want a higher resolution in something that isn't very interactive, like Bigscreen.

client/scenes/lobby_gui.cpp
Insprill: Ah, I see; I misunderstood. Fixed.

xytovl:
> How about a hard limit of 350%? I can get ~83ms of latency at 350% on the Quest 3 at 0.35x scale and 100Mbps. It's not great for active games, but it's fine for less interactive things like Bigscreen.

350 should be ok, Quest 3S for instance has the same GPU as 3, but lower resolution so there may be a use case.

PR #236, opened 2025-01-13
Add section about Steam Flatpak usage

losuler:
# Context

Unless I missed some mention of it, I didn't see any discussion about using WiVRn with the Flatpak version of Steam, which doesn't work out of the box. After spending some time testing and confirming each of these settings, I found that these steps are what are needed to get things working (at least with VRChat).
# Rationale

I wanted to avoid the combination of the `--filesystem=xdg-config/openxr:ro` permission and the symlink, by using the `--persist=.config/openxr` permission instead, but this was more annoying because it doesn't work if WiVRn is started before the Steam Flatpak (whereas the symlink doesn't rely on anything starting first).

Additionally, I originally wanted to just symlink `~/.config/openxr` and not `~/.config/openxr/1`, but this is seemingly incompatible with the `--filesystem=xdg-config/openxr:ro` permission.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,23 @@ If using Wine/Proton, it will probe for OpenVR at startup, so even for OpenXR ap
 
 When you start the server through flatpak, it will automatically configure the current OpenVR to use OpenComposite.
 
+### Steam Flatpak
+
+If you're using the Steam Flatpak, you'll need to grant read only access to the following paths:
+
+```bash
+flatpak override --user \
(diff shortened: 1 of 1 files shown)


saori-yuko:
what about `PRESSURE_VESSEL_FILESYSTEMS_RW=$XDG_RUNTIME_DIR/wivrn/comp_ipc:/var/lib/flatpak/app/io.github.wivrn.wivrn %command%` ?

PR #237, opened 2025-01-14
Forward systemName from client system properties

ImSapphire:
Used Beat Saber leaderboard mod BeatLeader for testing. Without this change, replays submitted when playing on WiVRn are displayed as 'Unknown headset with Oculus Touch controllers'. On my Quest 3 with this change, it properly displays as 'Quest 3 with Oculus Touch controllers', consistent with replays from Windows with Quest Link.

### DIFF
--- a/server/driver/wivrn_session.cpp
+++ b/server/driver/wivrn_session.cpp
@@ -220,6 +220,8 @@ wivrn::wivrn_session::wivrn_session(std::unique_ptr<wivrn_connection> connection
 		roles.right_profile = xdevs[roles.right]->name;
 	if (roles.gamepad >= 0)
 		roles.gamepad_profile = xdevs[roles.gamepad]->name;
+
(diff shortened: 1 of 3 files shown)


server/driver/wivrn_session.cpp
```diff
@@ -220,6 +220,8 @@ wivrn::wivrn_session::wivrn_session(std::unique_ptr<wivrn_connection> connection
 		roles.right_profile = xdevs[roles.right]->name;
 	if (roles.gamepad >= 0)
 		roles.gamepad_profile = xdevs[roles.gamepad]->name;
+
+	memcpy(xrt_system.base.properties.name, get_info().system_name.c_str(), get_info().system_name.length());
```
xytovl, line 224: Can we append "on WiVRn"? Take care of truncating to `XRT_MAX_SYSTEM_NAME_SIZE` and leaving a null character at the end.

PR #246, opened 2025-01-26
Typo on client screen when connected

shmink:
A very pedantic one here but just thought I'd add it anyway. On my headset when connected I would get a screen saying "Connection to fedora" (fedora being the name of my PC.

I'm assuming the intention was that it should say "Connected to fedora" instead.

### DIFF
--- a/client/scenes/lobby_gui.cpp
+++ b/client/scenes/lobby_gui.cpp
@@ -200,7 +200,7 @@ void scenes::lobby::gui_connecting(locked_notifiable<pin_request_data> & pin_req
 	if (server_name == "")
 		CenterTextH(fmt::format(_F("Connection")));
 	else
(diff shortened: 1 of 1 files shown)


xytovl:
The title is set before the connection succeeded, in case you have a firewall or if DNS issues, it will remain on that title.

If you want to write "connected", then you have to set 2 titles, and extend the "async_session" progress message.

When "next_scene" is present, connection is fully ready, with no video.
When "async_session" object is present, the connection is being established, either it is reaching for the server, waiting for the PIN or negociating initial data (which should be almost instant)

mitsuami-megane:
Would also need to change the keys in the locale files.

In the JA locale the text is translated with the meaning being closer to `Establishing connection to {}`

PR #248, opened 2025-01-29
Add option to use a palm offset

Vixeliz:
Since some devices don't have this and its required for certain games (ie rumble and bonelab), I added an option to instead use grip pose offset by a certain rotation. Right now its hard coded but it should definitely be changed to either be in the config or somewhere in code for common devices.

Sorry for any mistakes made im not super familiar with cpp so if anything needs to be changed let me know

### DIFF
--- a/server/driver/configuration.cpp
+++ b/server/driver/configuration.cpp
@@ -110,6 +110,11 @@ configuration configuration::read_user_configuration()
 				result.scale = *it;
 		}
 
+		if (auto it = json.find("grip-surface"); it != json.end())
+		{
+			result.grip_surface = *it;
+		}
+
 		if (auto it = json.find("bitrate"); it != json.end())
 			result.bitrate = *it;
 
--- a/server/driver/wivrn_controller.cpp
+++ b/server/driver/wivrn_controller.cpp
@@ -221,22 +221,17 @@ static std::ofstream & tracking_dump()
 
 static auto make_palm(int hand_id, bool palm_pose)
 {
-	std::optional<std::array<float, 3>> grip_surface = configuration::read_user_configuration().grip_surface;
-	if (palm_pose && !grip_surface)
-	{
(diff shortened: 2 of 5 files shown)


server/driver/wivrn_controller.cpp
```diff
@@ -448,8 +452,18 @@ void wivrn_controller::update_tracking(const from_headset::tracking & tracking,
 		cnx->set_enabled(aim.device, false);
 	if (not grip.update_tracking(tracking, offset))
 		cnx->set_enabled(grip.device, false);
-	if (not palm.update_tracking(tracking, offset))
+
+	bool palm_tracked = false;
+	// FIXME: The offset should be based off of the headset instead
+	if (cnx->get_info().palm_pose == false || configuration::read_user_configuration().force_palm) {
```
xytovl, line 458: `read_user_configuration` parses the json, we don't want to do that on every packet from the network!

The offset should be stored in the `pose_list` object at construction so that you can use it normally after.

Vixeliz: Kk, do you want the offset stored in pose list to be just a vec3 or also an optional so not rotating quats for ones that dont need it?

server/driver/configuration.cpp
```diff
@@ -138,6 +138,11 @@ configuration configuration::read_user_configuration()
 		else if (auto it = json.find("tcp_only"); it != json.end())
 			result.tcp_only = *it;
 
+		if (auto it = json.find("force-palm"); it != json.end())
+			result.force_palm = *it;
+		else if (auto it = json.find("force_palm"); it != json.end())
+			result.force_palm = *it;
```
xytovl, line 144: We don't need the version with underscores, `tcp_only` has it for compatibility, but from now on I would prefer to only have dashes in the json.

I am wondering if we should pass the rotation in the configuration and generate the palm only if configured

PR #255, opened 2025-02-06
Add HTC facial tracking support

Awzri:
This PR makes the `htc_face` branch up-to-date and fixes a couple bugs with it (most notably, facial tracking data not getting sent from the client and "Enable Face Tracking" in the lobby settings being grayed out for HTC headsets, while the icon still works).
This also explicitly adds the Focus Vision as a supported headset, else the facial tracking on the headset won't work.

I'm in no way a developer, and this is my first time ever making a pull request, so please let me know if there's something wrong.

This was tested on a Focus Vision and XR Elite.

(NOTE: There's an issue with the current release of oscavmgr crossing the eyes when looking left and right. I have a fix for this as well, which there should be a PR on that repo with that fix.)


client/scenes/stream_tracking.cpp
```diff
@@ -316,6 +316,7 @@ void scenes::stream::tracking()
 									application::get_htc_face_tracker_eye().get_weights(t0 + Δt, face_htc);
 								if (application::get_htc_face_tracking_lip_supported())
 									application::get_htc_face_tracker_lip().get_weights(t0 + Δt, face_htc);
+								packet.face_htc.emplace(face_htc);
```
xytovl, line 319: You should remove this line and instead replace line 315 with
```cpp
auto &face_htc = packet.face_htc.emplace();
```

galister:
hi, thanks for picking this up! not bad for a first PR.

please run a clang-format to clear the failed checks, otherwise looks good to me.

xytovl:
Can you split the focus vision into a separate commit and remove the merge commit (use a rebase instead)?

If you are not familiar with git, I can perform those operations myself instead, then force push to your branch

Awzri:
Hoping that's good? Ran into some issues, but everything was stashed and looks identical to the first PR.

PR #257, opened 2025-02-07
feat: change bitrate at runtime

galister:
works for vulkan + x264 so far


server/encoder/ffmpeg/video_encoder_ffmpeg.cpp
```diff
@@ -68,6 +68,14 @@ bool video_encoder_ffmpeg::once = set_log_level();
 
 std::optional<wivrn::video_encoder::data> video_encoder_ffmpeg::encode(bool idr, std::chrono::steady_clock::time_point target_timestamp, uint8_t slot)
 {
+	if (auto bitrate = pending_bitrate.exchange(BITRATE_UNCHANGED))
+	{
+		U_LOG_E("Reconfigure FFMPEG encoder");
```
xytovl, line 73: This should not be error level

PR #267, opened 2025-02-14
Add xrizer support

ImSapphire:
Since xrizer is still quite immature, OpenComposite will be selected before it. The AUR packages would need to add /opt/xrizer to OVR_COMPAT_SEARCH_PATH as well


xytovl:
@PassiveLemon, @damaestro is that fine for packaging?

I may extend it later with run-time overrides.

PassiveLemon:
I didn't even know this was a feature, I'll see if I can make it work

PassiveLemon:
LGTM, although run-time configuration would be greatly appreciated

PR #272, opened 2025-02-19
Update NVIDIA Video Codec SDK and enable AV1

Kierek:
I have updated the NVIDIA Video Codec SDK to the version 13.0 and migrated the settings based on
https://docs.nvidia.com/video-technologies/video-codec-sdk/12.1/nvenc-preset-migration-guide/index.html
but without enabling multipass, it added around 10ms latency during my brief testing.

I wasn't sure of the best approach to updating it, so I just copied files from
https://github.com/FFmpeg/nv-codec-headers
I hope that's ok. The only problem is that the newest SDK require fairly new drivers (570+), is that a problem?

It's my first time touching any encoding code and I'm not sure what to do with av1 params, there's no repeatSPSPPS but repeatSeqHdr seemed similar so I enabled it. There's also no videoFullRangeFlag.

H264 and HEVC appear a bit smoother than the current master. I'm not sure if AV1 provides significant improvements, even though at low bitrates it reports a slighly lower latency in graphs.

H264:
h264

HEVC:
hevc

AV1:
av1

### DIFF
--- a/server/encoder/video_encoder_nvenc.cpp
+++ b/server/encoder/video_encoder_nvenc.cpp
@@ -129,7 +129,7 @@ static auto encode_guid(video_codec codec)
 		case h265:
 			return NV_ENC_CODEC_HEVC_GUID;
 		case av1:
-			throw std::runtime_error("AV1 not implemented on nvenc");
+			return NV_ENC_CODEC_AV1_GUID;
 	}
 	throw std::out_of_range("Invalid codec " + std::to_string(codec));
 }
@@ -176,6 +176,7 @@ video_encoder_nvenc::video_encoder_nvenc(
 			break;
 
 		case video_codec::av1:
+			printf("%d AV1 presets\n", count);
 			break;
 	}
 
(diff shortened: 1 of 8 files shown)


xytovl:
I don't think we can requier 570+ drivers, would it work if we use 12.2 SDK? https://github.com/FFmpeg/nv-codec-headers/commit/c69278340ab1d5559c7d7bf0edf615dc33ddbba7 maybe just mention in the commit message the version that it was taken from.

I don't know what needs to be set for AV1 either, if it works it's already very good!
The videoFullRangeFlag is not really used, YCbCr conversion parameters are forced both for encode and decode. In the screenshots the colours are similar so it's fine.

Kierek:
Done, works fine with 12.2 SDK

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -129,7 +129,7 @@ static auto encode_guid(video_codec codec)
 		case h265:
 			return NV_ENC_CODEC_HEVC_GUID;
 		case av1:
-			throw std::runtime_error("AV1 not implemented on nvenc");
+			return NV_ENC_CODEC_AV1_GUID;
```
xytovl, line 132: Great! let's also update dashboard/qml/SettingsPage.qml and docs/configuration.md

Kierek: Sure. I'm not very familiar with dashboard since I mainly use envision, are these changes enough?

xytovl: looks good

ehfd:
Note that `two-pass-quarter` multipass doesn't harm latency that much and is included with Sunshine (Moonlight host server).

Kierek:
I've looked into Sunshine code and they are doing some things differently from us:

- P1 instead of P4
- NV_ENC_TUNING_INFO_ULTRA_LOW_LATENCY instead of NV_ENC_TUNING_INFO_LOW_LATENCY
- two-pass-quarter enabled

I will add options parameter support to nvenc and I hope we will find some better defaults.

ehfd:
P4 should be relatively fine instead of P1 in my experience. P1 could be to compromising on quality.

PR #277, opened 2025-02-23
Add Pico 4s/4 Ultra/A9210 interaction profile

TheRainTrain:
This fixes Pico controllers possibly using an incorrect profile, and identifies the Pico 4S / 4 Ultra / A9210 as model::pico_4s.  
Previously, P4S devices would not register button presses, this is the fix for that.  
  
The models used for left.glb and right.glb are the Pico 4 ones, as PicoXR does not currently distribute Pico 4S L/R 3D models officially.  
These are the closest official match, Pico 4S models may be added once PicoXR publishes them.


common/wivrn_packets.h
```diff
@@ -142,6 +142,13 @@ enum class interaction_profile : uint8_t
 	meta_touch_controller_quest_2,
 	samsung_odyssey_controller,
 	valve_index_controller,
+	// out-of-order to preserve enum ordinals
+	// otherwise this will force non-p4u users
+	// to update both their client and server
+	// for a feature they won't use
+	//
+	// TODO: order this correctly when the next major update comes around
+	bd_pico4s_controller,
```
xytovl, line 151: It is fine to break protocol compatibility, the automatic versionning won't detect changes in enums so you have to bump the `protocol_revision`.

client/hardware.cpp
```diff
@@ -250,6 +257,8 @@ std::string controller_name()
 			return "pico-neo3";
 		case model::pico_4:
 			return "pico-4";
+		case model::pico_4s:
+			return "pico-4s";
```
xytovl, line 261: If the asset is the same, you can return the same value for both. We'll split them when we have separate files.

client/scenes/stream_tracking.cpp
```diff
@@ -490,6 +490,7 @@ void scenes::stream::on_interaction_profile_changed(const XrEventDataInteraction
 			DO_PROFILE(khr, simple_controller)
 			DO_PROFILE(bd, pico_neo3_controller)
 			DO_PROFILE(bd, pico4_controller)
+			DO_PROFILE(bd, pico4s_controller)
```
xytovl, line 493: The macro expands to `"/interaction_profiles/" #vendor "/" #name`, in the interaction profile list it has the full "bytedance" name instead of "bd", the interaction profile won't be forwarded to the server correctly and it will always expose it as an Oculus Touch controller instead

TheRainTrain: `pico_neo3_controller` and `pico4_controller` also use "bd" instead of "bytedance", when they define "bytedance" in the interaction profile list. Is this a shared issue?

TheRainTrain: Additionally, these changes seem to work fine when tested

xytovl: This must be a shared issue, the logic is very new and I guess nobody tested it on Pico controllers.

The headset now sends the interaction profile that it selected, when the server resolves suggested bindings from the application, it first checks if there is a binding for the interaction profile reported by the headset, then for Oculus Touch controllers then for anything else.

The main differences will be trigger and squezze click that are unavailable on the Oculus Touch and applications may fail to report the controller name.

common/wivrn_packets.h
TheRainTrain: Donezo: https://github.com/WiVRn/WiVRn/pull/277/commits/beab3e70abab5e20214c24faa7ecc897d157b367

client/hardware.cpp
TheRainTrain: https://github.com/WiVRn/WiVRn/pull/277/commits/54ab2871bc17dfe531b2d39af77b30566888c2c7

client/scenes/stream_tracking.cpp
TheRainTrain: Strange.  
I went through and edited everything accordingly: https://github.com/WiVRn/WiVRn/pull/277/commits/beab3e70abab5e20214c24faa7ecc897d157b367

TheRainTrain:
I believe that should be everything addressed, including https://github.com/WiVRn/WiVRn/actions/runs/13482151558/job/37668868938 failing (quite a silly mistake on my end)

PR #281, opened 2025-02-25
Add Vive Ultimate Tracker support

Awzri:
This enables VUT support on the client. We send data about it to the server, but there currently isn't anything to handle this on the server yet.

Needs review from @xytovl for the client part, and afterwards @galister will work on the server part

### DIFF
--- a/client/scenes/stream_tracking.cpp
+++ b/client/scenes/stream_tracking.cpp
@@ -310,8 +310,7 @@ void scenes::stream::tracking()
 
 					if (control.enabled[size_t(tid::motion_trackers)])
 					{
-						xr::xr_tracker_fill_packet(instance, session, t0 + Δt, world_space, xr::vive_xr_trackers);
-						packet.motion_trackers = xr::vive_xr_trackers;
+						packet.motion_trackers = xr::xr_tracker_compose_packet(instance, session, t0 + Δt, world_space, xr::vive_xr_trackers);
 					}
 
 					// Hand tracking data are very large, send fewer samples than other items
@@ -595,8 +594,8 @@ void scenes::stream::on_interaction_profile_changed(const XrEventDataInteraction
 		interaction_profiles[i] = interaction_profile::none;
 	}
 
-	if (application::get_config().check_feature(feature::motion_tracking))
+	for (auto & tracker: xr::vive_xr_trackers)
 	{
-		xr::xr_tracker_prepare_packet(instance, session, xr::vive_xr_trackers);
+		tracker.set_active(instance, session);
 	}
 }
--- a/client/scenes/stream_tracking.cpp
+++ b/client/scenes/stream_tracking.cpp
@@ -310,7 +310,7 @@ void scenes::stream::tracking()
 
 					if (control.enabled[size_t(tid::motion_trackers)])
 					{
-						packet.motion_trackers = xr::xr_tracker_compose_packet(instance, session, t0 + Δt, world_space, xr::vive_xr_trackers);
+						packet.motion_trackers = xr::xr_tracker_compose_packet(instance, session, t0 + Δt, world_space, application::get_vive_xr_trackers());
 					}
 
 					// Hand tracking data are very large, send fewer samples than other items
@@ -594,8 +594,5 @@ void scenes::stream::on_interaction_profile_changed(const XrEventDataInteraction
 		interaction_profiles[i] = interaction_profile::none;
 	}
 
-	for (auto & tracker: xr::vive_xr_trackers)
-	{
-		tracker.set_active(instance, session);
-	}
+	xr::xr_tracker_update_active(instance, session);
 }
--- a/client/scenes/stream_tracking.cpp
+++ b/client/scenes/stream_tracking.cpp
@@ -242,6 +242,7 @@ void scenes::stream::tracking()
 	std::vector<serialization_packet> packets;
 
 	const bool hand_tracking = config.check_feature(feature::hand_tracking);
+	const bool motion_tracking = config.check_feature(feature::motion_tracking);
 	from_headset::face_type face_tracking = from_headset::face_type::none;
 	if (config.check_feature(feature::face_tracking))
 	{
@@ -308,7 +309,7 @@ void scenes::stream::tracking()
 							packet.device_poses.emplace_back(locate_space(device, space, world_space, t0 + Δt));
 					}
 
-					if (control.enabled[size_t(tid::motion_trackers)])
+					if (motion_tracking and control.enabled[size_t(tid::motion_trackers)])
 					{
 						packet.motion_trackers = xr::xr_tracker_compose_packet(instance, session, t0 + Δt, world_space, application::get_vive_xr_trackers());
 					}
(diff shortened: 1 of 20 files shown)


client/xr/htc_xr_tracker.cpp
```diff
@@ -0,0 +1,116 @@
...
+	// TODO wait for when vive makes a proper type and insert it when released
+	XrPathsForInteractionProfileEnumerateInfoHTC enum_info{
+	        .type = XR_TYPE_UNKNOWN,
+	        .next = nullptr,
+	        .interactionProfile = tracker_profile,
+	        .userPath = user_path};
+
+	auto xr_tracker_paths = xr::details::enumerate<XrPath>(xrEnumeratePathsForInteractionProfileHTC, inst, enum_info);
+
+	return xr_tracker_paths;
```
xytovl, line 56: You can omit the local variable:
```cpp
return xr::details::enumerate<XrPath>(xrEnumeratePathsForInteractionProfileHTC, inst, enum_info);
```

client/xr/htc_xr_tracker.cpp
```diff
@@ -0,0 +1,116 @@
...
+	for (auto & path: tracker_paths)
+	{
+		auto name_info = XrInputSourceLocalizedNameGetInfo{
+		        .type = XR_TYPE_INPUT_SOURCE_LOCALIZED_NAME_GET_INFO,
+		        .next = nullptr,
+		        .sourcePath = path,
+		        .whichComponents = XR_INPUT_SOURCE_LOCALIZED_NAME_USER_PATH_BIT};
+
+		auto role_name = xr::details::enumerate<char>(xrGetInputSourceLocalizedName, session, &name_info);
+		tracker_roles.emplace_back(role_name);
```
xytovl, line 73: You can remove the variable

client/scenes/stream_tracking.cpp
```diff
@@ -280,6 +281,26 @@ void scenes::stream::tracking()
...
+							XrSpace space = xr::xr_tracker_spaces[i];
+							XrSpaceLocation location{.type = XR_TYPE_SPACE_LOCATION};
+							xrLocateSpace(space, world_space, t0 + Δt, &location);
+							// If location and orientation is invalid, the tracker is disconnected. Don't send it in that case.
+							if (location.locationFlags == (XR_SPACE_LOCATION_ORIENTATION_VALID_BIT | XR_SPACE_LOCATION_POSITION_VALID_BIT))
+							{
+								auto new_tracker = wivrn::from_headset::tracking::motion_tracker{
+								        .id = i,
+								        .pose = location.pose,
+								        .role = xr::xr_tracker_get_roles_enum(instance, session)[i],
```
xytovl, line 297: This operation will be performed a huge amount of times: the outer loop runs every 2ms, then there is a loop on prediction times (usually 4 samples), and then a last one for trackers.

We must avoid at all cost memory allocations (may introduce occasional latency spikes) and useless calls.

I think one of the solutions is to wrap the trackers in an object, which could cache its role, then react to the appropriate XR event when they change.

Awzri: There's no specific event for if the trackers connect, disconnect, or have their roles change (at least publicly available and documented, Vive Streaming can check but I'm not going so far as to reverse engineer that).

If an event exists I'm not aware of in which we can check if a controller (dis)connected, if the localized name is changed, or if a location bit becomes (in)valid, let me know and I can try using that.

Right now though, I'm thinking we can change the get function to provide a vector of optional enums, and then put it into a variable at the start of the tracking function, before the loop.

```cpp
std::vector<std::optional<from_headset::tracking::tracker_role>> cached_xr_tracker_roles;
if (config.check_feature(feature::motion_tracking))
{
  cached_xr_tracker_roles = xr::xr_tracker_get_roles_enum(instance, session);
}
```

In the tracking thread, we only then have to add the role if the tracker is tracking and doesn't already have a cached role (the tracker connects during streaming), and remove the role if the tracker is disconnected.

Something like this:
``` cpp
// If location and orientation is invalid, the tracker is disconnected. Don't send it in that case.
if (location.locationFlags == (XR_SPACE_LOCATION_ORIENTATION_VALID_BIT | XR_SPACE_LOCATION_POSITION_VALID_BIT))
{
  // If the tracker just connected, find the role associated and store it.
  if (!cached_xr_tracker_roles[i])
    cached_xr_tracker_roles[i] = xr::xr_tracker_get_roles_enum(instance, session)[i].value();
  packet.motion_trackers.emplace_back(
        i,
        location.pose,
        cached_xr_tracker_roles[i].value());
}
// If the tracker disconnected still has a role cached, remove it, as it could change on reconnect.
else if (cached_xr_tracker_roles[i])
{
  cached_xr_tracker_roles[i] = std::nullopt;
}
```
This could also be handled on the server if desired, but that would currently send all 5 spaces and their location bits even if no trackers are connected.

We could also have no handling like this, but trackers become constant when streaming starts, and none can be added, removed, or changed during stream, which may be undesirable.

client/xr/htc_xr_tracker.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+	for (auto & path: tracker_paths)
+	{
+		auto name_info = XrInputSourceLocalizedNameGetInfo{
+		        .type = XR_TYPE_INPUT_SOURCE_LOCALIZED_NAME_GET_INFO,
+		        .next = nullptr,
+		        .sourcePath = path,
+		        .whichComponents = XR_INPUT_SOURCE_LOCALIZED_NAME_USER_PATH_BIT};
+
+		auto role_name = xr::details::enumerate<char>(xrGetInputSourceLocalizedName, session, &name_info);
```
xytovl, line 72: Is this really the only method to access the role of the tracker?

The specification says
> [xrGetInputSourceLocalizedName](https://registry.khronos.org/OpenXR/specs/1.0/man/html/xrGetInputSourceLocalizedName.html) returns a string for the bound source in the current system locale.

This is not supposed to work if the headset is not set to English (and maybe even US English).

client/scenes/stream_tracking.cpp
xytovl: The `scenes::stream::tracking()` is already too large before this pull request, if we need all this logic for trackers, we should encapsulate it in an object.

We also don't need to return or store a vector of roles, it seems to be easier to do it per tracker.

xytovl: If you have the hardware to test, you can check if the `XrEventDataInteractionProfileChanged` event is fired when trackers are added, removed or modified.

client/xr/htc_xr_tracker.cpp
Awzri: I guess.
https://github.com/ViveSoftware/VIVE-OpenXR-Unity/blob/25a5fd212420688952ead9deba735357656278ec/com.htc.upm.vive.openxr/Runtime/Features/Tracker/XR/Scripts/ViveXRTracker.cs#L3628

Switching language on my Focus Vision doesn't seem to affect the roles that appear either.
However, if you know someone with a Focus 3/Vision or XR Elite in a region other than the US though, it may be worth testing to see if it stays the same.

galister: we currently can't really map the role to anything on the server side. best we can do is expose them as xdevs

galister:
please push your latest working version and we can help clean it up

xytovl:
replaced by #369

PR #289, opened 2025-03-07
added zh-TW

Jaydenha09:


xytovl:
Code and tools now work, I would need a confirmation that the apk works correctly

PR #294, opened 2025-03-21
Add OpenVR runtime override as a setting, add --early-active-runtime to test this

20kdc:
I ran into some trouble debugging issues with Vivecraft because my attempts to patch OpenComposite were being automatically overridden by WiVRn.

It would be nice to attach a UI to this option, but I'm not sure on the details there and I'd like to keep the PR small and easily reviewed.

Being able to set the compat path is more useful than simply disabling compat override because of things like pressure-vessel bypass, which are finicky to get right.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -181,3 +181,12 @@ Default value: `avahi`
 How to publish the service over the network, `avahi` or null.
 
 If set to null, service will not be published and address has to be entered manually on the headset.
+
+## `openvr-compat-path`
+Default value: unset
+
+Provides the path to the directory of an OpenVR compatibility tool (such as OpenComposite).
+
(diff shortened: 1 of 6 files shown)


xytovl:
This looks fine, maybe add a description in the docs/configuration.md.

I will probably rework configuration/command line arguments soon, we need a more scalable mechanism.

20kdc:
I've attempted to document things; the automatic override behaviour wasn't previously documented there, it seems?

xytovl:
> I've attempted to document things; the automatic override behaviour wasn't previously documented there, it seems?

I don't think that was documented, thanks for adding it.

I'll merge it as soon as CI is happy, well except for translations, that's unrelated to the change.

I'll see later how I can improve the configuration/commandline parameters, it would be nice to be able to split openxr/openvr parts, control none/early/on-connection

Issue #295, opened 2025-03-23
[Bug]: Client crash upon leaving guardian when face tracking is active (Quest Pro)


Artefact2:
### Describe the bug

Streaming on the Quest Pro (v74) will crash upon leaving the guardian, when face tracking with oscavmgr is active. The bug does not happen when oscavmgr is not running.

### Steps to Reproduce

* Start WiVRn server, start the client (wired), start oscavmgr. In the oscavmgr interface, DRIVE, TRACK, GAZE and FACE are green.
* Physically move outside of the guardian area (make the "switch to stationary" menu pop up)
* Streaming will immediately stop, with a message on the server that reads: `ERROR [run] Exception in network thread: Socket shutdown.` Returning to the guardian does not resume streaming. In adb logcat, one can see `[WiVRn] [info] Exception in tracking thread, exiting: xrGetFaceExpressionWeights2FB(id, &info, &expression_weights): XR_ERROR_RUNTIME_FAILURE`.

### Expected behavior

Keep hearing streaming audio outside the guardian. Resume streaming when moving back inside the guardian.

### WiVRn Version

c23c0ed8e88eee38565b4170b999a66180485a92

### Server Config

n/a

### Server Logs

n/a

### What GPU did the bug occur on?

AMD

### Vulkan Info

n/a

### What HMD did the bug occur on?

Quest Pro

### ADB Logs

[adb-logcat.log](https://github.com/user-attachments/files/19409620/bar.log)

galister:
thanks for the report. are you able to test if this still happens with the PR linked above?

Artefact2:
That fixes it, it now works as expected. Thanks!

PR #297, opened 2025-03-23
handle soft failures in xrGetFaceExpressionWeights2FB

galister:
attempts to fix https://github.com/WiVRn/WiVRn/issues/295


Artefact2:
After testing, I can confirm this fixes my issue #295. Thanks!

Issue #147, opened 2024-10-12
Thumbs are twisted 90° on Quest 3


OrionMoonclaw:
When using hand tracking the thumbs are twisted both in the lobby and in-game and bend sideways
![JPEG_20241012_221902_1193161877113846659](https://github.com/user-attachments/assets/d6268571-da75-4058-acb9-107dda33df29)
![JPEG_20241012_221912_4615894127101636166](https://github.com/user-attachments/assets/872b3d45-7358-4594-809e-4564058142ca)
![JPEG_20241012_221938_4499927427502778051](https://github.com/user-attachments/assets/ad664100-30b3-4300-bd4c-0bf4534406cf)

xytovl:
Do you have a test that can be executed both native on the headset and through WiVRn?

RinLovesYou:
this video should show the discrepancy between the two a bit better 
https://github.com/user-attachments/assets/1d9e1bb2-6038-404c-8e6e-65fdf8dba568

xytovl:
Position/orientation data does not seem different between Meta runtime and WiVRn lobby, it looks like a difference in graphic model and rendering.

RinLovesYou:
> Position/orientation data does not seem different between Meta runtime and WiVRn lobby, it looks like a difference in graphic model and rendering.

Yet the thumb is rotated 90 degrees. this is very apparent in games that actually have textures like shown in the above screenshot, but even in the wivrn lobby you can tell it's wrong

OrionMoonclaw:
There is a possibility that both WiVRn and OC make the same mistake here, but that would be pretty unlikely imo

Sadly I don't know of any native OpenXR games that use hand tracking

RinLovesYou:
I didn't know, other quest apps that use hand tracking have no issues like this. Although i honestly wouldn't be surprised if this is so old that every game/game engine accounts for this and it never got fixed because of that. Though i can only find ancient issue posts about this that are marked as fixed

RinLovesYou:
> There is a possibility that both WiVRn and OC make the same mistake here, but that would be pretty unlikely imo

OC passes hand data through and does the same math as valve does for steam link.

> Sadly I don't know of any native OpenXR games that use hand tracking

On pc i know only of openxr-simple-playground

Stellanora64:
[...]

PR #300, opened 2025-03-24
Fix meta handtracking thumb rotation offset #147

Stellanora64:
This should fix the thumbs being rotated by 90 degrees with hand tracking on meta headsets in #147

Hopefully the device checking was done correctly, let me know if there is anything there I need to fix.

This patch works on my Quest 3, but I have not tested on other headsets.

### DIFF
--- a/client/xr/hand_tracker.cpp
+++ b/client/xr/hand_tracker.cpp
@@ -23,10 +23,7 @@
 #include "xr/session.h"
 #include <cassert>
 #include <glm/glm.hpp>
-#include <glm/gtc/matrix_transform.hpp>
-#include <glm/gtc/matrix_transform.inl>
 #include <glm/gtc/quaternion.hpp>
-#include <glm/gtx/quaternion.hpp>
 #include "hardware.h"
 
 static PFN_xrDestroyHandTrackerEXT xrDestroyHandTrackerEXT{};
--- a/client/xr/hand_tracker.cpp
+++ b/client/xr/hand_tracker.cpp
@@ -18,13 +18,13 @@
  */
 
 #include "hand_tracker.h"
+#include "hardware.h"
 #include "xr/check.h"
 #include "xr/instance.h"
 #include "xr/session.h"
 #include <cassert>
 #include <glm/glm.hpp>
 #include <glm/gtc/quaternion.hpp>
-#include "hardware.h"
 
 static PFN_xrDestroyHandTrackerEXT xrDestroyHandTrackerEXT{};
 
@@ -33,8 +33,8 @@ XrResult xr::destroy_hand_tracker(XrHandTrackerEXT id)
 	return xrDestroyHandTrackerEXT(id);
 }
 
-xr::hand_tracker::hand_tracker(instance & inst, session & session, const XrHandTrackerCreateInfoEXT & info): 
-	hand_id(info.hand)
+xr::hand_tracker::hand_tracker(instance & inst, session & session, const XrHandTrackerCreateInfoEXT & info) :
+        hand_id(info.hand)
 {
 	static auto xrCreateHandTrackerEXT = inst.get_proc<PFN_xrCreateHandTrackerEXT>("xrCreateHandTrackerEXT");
 	assert(xrCreateHandTrackerEXT);
@@ -76,7 +76,7 @@ std::optional<std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT>> xr::
 
 	if (!locations.isActive)
 		return std::nullopt;
-	
+
 	float offset_angle;
 
 	// check if we have a device that needs a thumb offset
@@ -92,7 +92,7 @@ std::optional<std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT>> xr::
 			else if (hand_id == XR_HAND_RIGHT_EXT)
 				offset_angle = glm::radians(90.0f);
 			break;
-		
+
 		default:
 			offset_angle = 0.0f;
 			break;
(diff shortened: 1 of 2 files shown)


client/xr/hand_tracker.h
```diff
@@ -43,5 +43,8 @@ class hand_tracker : public utils::handle<XrHandTrackerEXT, destroy_hand_tracker
 	std::optional<std::array<joint, XR_HAND_JOINT_COUNT_EXT>> locate(XrSpace space, XrTime time);
 
 	static bool check_flags(const std::array<joint, XR_HAND_JOINT_COUNT_EXT> & joints, XrSpaceLocationFlags position, XrSpaceVelocityFlags velocity);
+private:
+	XrHandEXT hand_id;
+	float offset_angle;
```
xytovl, line 48: No need to have the `offset_angle` as a member variable, it can be local to the `locate` method

client/xr/hand_tracker.cpp
```diff
@@ -32,11 +38,24 @@ XrResult xr::destroy_hand_tracker(XrHandTrackerEXT id)
 
 xr::hand_tracker::hand_tracker(instance & inst, session & session, const XrHandTrackerCreateInfoEXT & info)
 {
+	hand_id = info.hand;
```
xytovl, line 41: Prefer a `const XrHandExt hand_id` member, and initialize with
```cpp
xr::hand_tracker::hand_tracker(instance & inst, session & session, const XrHandTrackerCreateInfoEXT & info):
  hand_id(info.hand)
  {
  }
  ```

File: client/xr/hand_tracker.cpp [hunk 3]
```diff
@@ -76,7 +95,48 @@ std::optional<std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT>> xr::
 	std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT> joints;
 	for (int i = 0; i < XR_HAND_JOINT_COUNT_EXT; i++)
 	{
-		joints[i] = {joints_pos[i], joints_vel[i]};
+		switch (guess_model())
+		{
+			case model::meta_quest_3:
+			case model::meta_quest_pro:
+			case model::meta_quest_3s:
```
xytovl, line 102: We can add quest 1 and 2 as well

client/xr/hand_tracker.cpp
```diff
@@ -76,7 +95,48 @@ std::optional<std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT>> xr::
...
+				.z = offset_rotation.z,
+				.w = offset_rotation.w,
+			};
+
+			joints[i] = {joints_pos[i], joints_vel[i]};
+		}
+		else
+		{
+			joints[i] = {joints_pos[i], joints_vel[i]};
+		}
```
xytovl, line 138: Can be after the `if`, without an if/else, as they're doing the same thing.

client/xr/hand_tracker.cpp
```diff
@@ -76,7 +95,48 @@ std::optional<std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT>> xr::
...
+				if (hand_id == XR_HAND_LEFT_EXT)
+					offset_angle = glm::radians(-90.0f);
+				else if (hand_id == XR_HAND_RIGHT_EXT)
+					offset_angle = glm::radians(90.0f);
+				break;
+			
+			default:
+				offset_angle = 0.0f;
+				break;
+		}
```
xytovl, line 112: This doesn't need to be in the `for` loop, it doesn't depend on `i`.

client/xr/hand_tracker.h
Stellanora64: This should be fixed in [043d140](https://github.com/WiVRn/WiVRn/pull/300/commits/043d1409b70f514151367b58f2c4815f1adb468f)

client/xr/hand_tracker.cpp
Stellanora64: fixed in [1e4cbb3](https://github.com/WiVRn/WiVRn/pull/300/commits/1e4cbb323f5a9216f95d781720c705dce0bc797d)

client/xr/hand_tracker.cpp
Stellanora64: fixed in [e069e60](https://github.com/WiVRn/WiVRn/pull/300/commits/e069e60d43462ee705888a109c528af618ed64eb)

client/xr/hand_tracker.cpp
Stellanora64: Fixed this in the refactor [043d140](https://github.com/WiVRn/WiVRn/pull/300/commits/043d1409b70f514151367b58f2c4815f1adb468f)

client/xr/hand_tracker.cpp
Stellanora64: I tried to fix this in [043d140](https://github.com/WiVRn/WiVRn/pull/300/commits/043d1409b70f514151367b58f2c4815f1adb468f) using your suggestion

However, in `application.cpp` line 1155:
``` 	
       if (hand_tracking_supported)
	{
		left_hand = xr_session.create_hand_tracker(XR_HAND_LEFT_EXT);
		right_hand = xr_session.create_hand_tracker(XR_HAND_RIGHT_EXT);
	}
```
It errors that it `cannot be referenced -- it is a deleted function` so I tried removing the default initialising, but that also didn't work, same with the compiler's suggestion of:
```
38 |         hand_tracker() = default;
        |                          ^~~~~~~
        |                          delete
```

Stellanora64: I've reverted it back for now to not use a `const` anymore for `hand_id`, and it's at least building again. 

If you have any ideas on how to work around that do let me know, I did try `unique_ptr` on some things but that also didn't seem to help.

xytovl: I forgot we would be assigning to the object, it's fine to leave it like that

xytovl:
Just fix formatting, run `clang-format -i` on modified files and we're good for a merge.

You can either squash all the commits yourself or I'll do it when merging.

Stellanora64:
Formatting should be fixed now, probably be easier if you could squash the commits while merging

PR #315, opened 2025-04-12
Client-side post-processing

Baplar:
This is an early working version of the [Snapdragon Game Super Resolution V1 algorithm](https://github.com/SnapdragonStudios/snapdragon-gsr/tree/main/sgsr/v1), implemented within the reprojection shader.

I added toggles and sliders in the client settings to enable it and tweak its behavior (disabled by default, of course).
It seems to be working, in the sense that the result sent to the display is indeed of higher resolution, and sharpened with various results depending on the sliders.
From my initial testing, visual fidelity improvements are arguably present (slight sharpening) but difficult to notice. However, it might be due to me not having tried the right combinations of settings yet.

Although I am a programmer by trade, this is my very first time working with graphics programming, so I will gladly take your criticisms and advice!

### DIFF
--- a/.github/screenshot-script.txt
+++ b/.github/screenshot-script.txt
@@ -82,6 +82,17 @@ set right rotation 0.4 0 0 0.9
 sleep 1
 exec grim -g "0,0 2560x1440" ${LANG}/screenshot-04-settings.png
 
+# Post-processing
+set right direction -1.95 1.45 -1
+sleep 1
+set right trigger_value 1
+sleep 1
+set right trigger_value 0
+sleep 1
+set right rotation 0.4 0 0 0.9
+sleep 1
+exec grim -g "0,0 2560x1440" ${LANG}/screenshot-07-post-processing.png
+
 # About
 set right direction -2.4 0.45 -1
 sleep 1
--- a/client/application.cpp
+++ b/client/application.cpp
@@ -1044,6 +1044,7 @@ void application::initialize()
 	opt_extensions.push_back(XR_FB_COMPOSITION_LAYER_DEPTH_TEST_EXTENSION_NAME);
 	opt_extensions.push_back(XR_KHR_COMPOSITION_LAYER_COLOR_SCALE_BIAS_EXTENSION_NAME);
 	opt_extensions.push_back(XR_KHR_VISIBILITY_MASK_EXTENSION_NAME);
+	opt_extensions.push_back(XR_FB_COMPOSITION_LAYER_SETTINGS_EXTENSION_NAME);
 
 	for (const auto & i: interaction_profiles)
 	{
@@ -1118,6 +1119,12 @@ void application::initialize()
 		spdlog::info("    HTC lip tracking support: {}", (bool)htc_face_properties.supportLipFacialTracking);
 		htc_face_tracking_eye_supported = htc_face_properties.supportEyeFacialTracking;
 		htc_face_tracking_lip_supported = htc_face_properties.supportLipFacialTracking;
+	}
+
+	if (utils::contains(xr_extensions, XR_FB_COMPOSITION_LAYER_SETTINGS_EXTENSION_NAME))
+	{
+		spdlog::info("    OpenXR post-processing extension support: true");
+		openxr_post_processing_supported = true;
 	}
 
 	switch (xr_system_id.passthrough_supported())
--- a/client/configuration.h
+++ b/client/configuration.h
@@ -25,6 +25,7 @@
 #include <map>
 #include <mutex>
 #include <optional>
+#include <simdjson.h>
 #include <string>
 
 namespace xr
@@ -53,6 +54,17 @@ public:
 	bool passthrough_enabled = false;
 	bool mic_unprocessed_audio = false;
 
+	// Snapdragon Game Super Resolution
+	struct sgsr_settings
+	{
+		bool enabled = false;
+		float upscaling_factor = 1.5;
+		bool use_edge_direction = true;
+		float edge_threshold = 4.0;
+		float edge_sharpness = 2.0;
+	};
+	sgsr_settings sgsr{};
+
 	std::string virtual_keyboard_layout = "QWERTY";
 
 	bool check_feature(feature f) const;
@@ -61,6 +73,8 @@ public:
 private:
 	mutable std::mutex mutex;
 	std::map<feature, bool> features;
+
+	void parse_sgsr_options(simdjson::simdjson_result<simdjson::dom::object> root);
 
 public:
 	configuration(xr::system &);
--- a/client/configuration.h
+++ b/client/configuration.h
@@ -65,6 +65,14 @@ public:
 	};
 	sgsr_settings sgsr{};
 
+	// XR_FB_composition_layer_settings extension flags
+	struct openxr_post_processing_settings
+	{
+		XrCompositionLayerSettingsFlagsFB super_sampling = 0;
+		XrCompositionLayerSettingsFlagsFB sharpening = 0;
+	};
+	openxr_post_processing_settings openxr_post_processing{};
+
 	std::string virtual_keyboard_layout = "QWERTY";
 
 	bool check_feature(feature f) const;
@@ -75,6 +83,7 @@ private:
 	std::map<feature, bool> features;
 
 	void parse_sgsr_options(simdjson::simdjson_result<simdjson::dom::object> root);
+	void parse_openxr_post_processing_options(simdjson::simdjson_result<simdjson::dom::object> root);
 
 public:
 	configuration(xr::system &);
--- a/client/scenes/lobby_gui.cpp
+++ b/client/scenes/lobby_gui.cpp
@@ -145,8 +145,16 @@ void scenes::lobby::tooltip(std::string_view text)
 	const ImVec2 text_size = ImGui::CalcTextSize(text.data(), text.data() + text.size(), true);
 	const ImVec2 size = {text_size.x + style.WindowPadding.x * 2.0f, text_size.y + style.WindowPadding.y * 2.0f};
 	pos.x = std::min(pos.x, viewport->Pos.x + viewport->Size.x - size.x / 2);
+	ImVec2 pivot = {0.5, 1};
 
-	ImGui::SetNextWindowPos(pos, ImGuiCond_Always, {0.5, 1});
+	// Move tooltip below the item if it overflows on the top
+	if (pos.y - size.y <= viewport->Pos.y)
+	{
+		pos.y = ImGui::GetItemRectMax().y + constants::style::tooltip_distance;
+		pivot.y = 0;
+	}
+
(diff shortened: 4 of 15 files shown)


galister:
Since the SGSR setting can only be changed in the lobby (while not streaming), wouldn't it make sense to have separate shaders for SGSR and non-SGSR?

Baplar:
That's a good point. From my cursory understanding, the shader compilation is already able to trim the unused branches, so splitting the shader into its own file would probably not result in significant performance changes. But it sure would be nicer to browse through. I'll try doing that!

Baplar:
Now that I think about it, currently I put the SGSR pass after the vertex pass that performs foveation. Would it be better to run it before somehow? Would that even be possible on a technical level?
Sorry if it's a dumb question, I have never written GL until yesterday, I am mostly just copying code and making it up as I go ^^

galister:
Actually it's probably fine like this. Just do a clang format to clear the CI check.

OzzyHelix:
@Baplar is there any chance I could get an apk to try this out?

xytovl:
> @Baplar is there any chance I could get an apk to try this out?

Just check the "Actions" tab, https://github.com/WiVRn/WiVRn/actions/runs/14422093511?pr=315 is the latest build for this PR

OzzyHelix:
apk-Standard-Release is unsigned

> > @Baplar is there any chance I could get an apk to try this out?
> 
> Just check the "Actions" tab, https://github.com/WiVRn/WiVRn/actions/runs/14422093511?pr=315 is the latest build for this PR

Baplar:
> apk-Standard-Release is unsigned

I’ve heard from someone else that the apk-Standard-Debug works for them.

Baplar:
In the meantime, thanks to @xytovl ’s additional findings, I am cooking up a version which will use Meta’s upscaling/super-sampling/sharpening algorithm on Quest headsets when available, and will only fall back to this SGSR implementation if that’s not supported (e.g. Pico or Vive headsets)

OzzyHelix:
> In the meantime, thanks to @xytovl ’s additional findings, I am cooking up a version which will use Meta’s upscaling/super-sampling/sharpening algorithm on Quest headsets when available, and will only fall back to this SGSR implementation if that’s not supported (e.g. Pico or Vive headsets)

it looks interesting from what I have tested the image does look sharper. the issue I have noticed is there was some latency spikes while using it. I had like 146% scale and upscale set to 165% so idk

OzzyHelix:
it would be nice if we had options to sharpen the image more

Baplar:
Concerning the option of running both MQSR and SGSR: according to [this article from Meta](https://developers.meta.com/horizon/blog/vr-image-quality-meta-quest-super-resolution/), MQSR is literally an optimized version of SGSR running closer to the metal. So in fact, there is little to no need to use both at the same time.
I might still leave the option, just for the fact that manually doing SGSR allows people to select different values for their upscale factor, edge threshold and edge sharpness? Not sure how big the demand is for that.

client/scenes/lobby_gui.cpp
```diff
@@ -785,6 +806,159 @@ void scenes::lobby::gui_settings()
...
+				tooltip(_("Improve clarity of high contrast edges and counteract blur.\nUseful when the input resolution is low compared to the headset display"));
+			}
+		}
+	}
+	if (supports_sgsr(guess_model()))
+	{
+		// SGSR
+		{
+			bool enabled = config.sgsr.enabled;
+			if (ImGui::Checkbox(_S("Enable Snapdragon Game Super Resolution"), &enabled))
```
xytovl, line 874: It should work with `&config.sgsr.enabled` directly

client/configuration.h
```diff
@@ -53,6 +54,23 @@ class configuration
...
+	};
+	sgsr_settings sgsr{};
+
+	struct mqsr_settings
+	{
+		XrCompositionLayerSettingsFlagsFB super_sampling = 0;
+		XrCompositionLayerSettingsFlagsFB sharpening = 0;
+	};
+	mqsr_settings mqsr{};
+
```
xytovl, line 73: We should add a comment to explain sgsr = Snapdragon Game Super Resolution, Mqsr = Meta Quest Super resolution.

Mqsr may also have to be renamed to something else, as `XR_FB_composition_layer_settings` could be implemented by other runtimes. For instance "OpenXR post processing"

client/application.cpp
```diff
@@ -1044,6 +1044,8 @@ void application::initialize()
 	opt_extensions.push_back(XR_FB_COMPOSITION_LAYER_DEPTH_TEST_EXTENSION_NAME);
 	opt_extensions.push_back(XR_KHR_COMPOSITION_LAYER_COLOR_SCALE_BIAS_EXTENSION_NAME);
 	opt_extensions.push_back(XR_KHR_VISIBILITY_MASK_EXTENSION_NAME);
+	opt_extensions.push_back(XR_FB_COMPOSITION_LAYER_SETTINGS_EXTENSION_NAME);
+	opt_extensions.push_back(XR_META_AUTOMATIC_LAYER_FILTER_EXTENSION_NAME);
```
xytovl, line 1048: Second extension is unused

client/scenes/lobby_gui.cpp
```diff
@@ -785,6 +806,159 @@ void scenes::lobby::gui_settings()
...
+				ImGui::EndCombo();
+			}
+			vibrate_on_hover();
+			if (ImGui::IsItemHovered())
+			{
+				tooltip(_("Improve clarity of high contrast edges and counteract blur.\nUseful when the input resolution is low compared to the headset display"));
+			}
+		}
+	}
+	if (supports_sgsr(guess_model()))
```
xytovl, line 869: I think we can drop the condition, the only unsupported headset is Quest 1, which has a rather small install base and supports mqsr, users would get the warning already

client/scenes/lobby_gui.cpp
Baplar: Indeed, and same with the edge direction below. Fixed.

client/configuration.h
Baplar: Agreed. I renamed it `openxr_post_processing`, that seems like the most descriptive yet extensible name.

client/application.cpp
Baplar: I added it for experimenting with the "automatic layer selection" extension from Meta, but given that it’s only really useful for adjusting to GPU resources availability for standalone apps, it didn’t really make sense for WiVRn so I reverted it. And I forgot to remove the extension activation here, good catch.

client/scenes/lobby_gui.cpp
Baplar: Works for me!

client/scenes/lobby_gui.cpp
```diff
@@ -122,6 +122,21 @@ static void display_recentering_tip(imgui_context & ctx, const std::string & tip
...
+	switch (flag)
+	{
+		case XR_COMPOSITION_LAYER_SETTINGS_NORMAL_SUPER_SAMPLING_BIT_FB:
+		case XR_COMPOSITION_LAYER_SETTINGS_NORMAL_SHARPENING_BIT_FB:
+			return _("Normal##openxr_post_processing_normal");
+		case XR_COMPOSITION_LAYER_SETTINGS_QUALITY_SUPER_SAMPLING_BIT_FB:
+		case XR_COMPOSITION_LAYER_SETTINGS_QUALITY_SHARPENING_BIT_FB:
+			return _("Quality##openxr_post_processing_quality");
+		default:
+			return _("Disabled##openxr_post_processing_disabled");
```
xytovl, line 136: We should keep the `##openxr_post_processing_disabled` out of the localisable part.

Baplar: For all of them, or just Disabled? I was worried about future collisions where we might want to translate stuff like "Normal" differently based on context

xytovl: The `##` part is useful to disambiguate for ImGui. If it's just for translation, you can use `boost::locale::pgettext(context, text)` (with abbreviated macro `_cS()`).

.github/screenshot-script.txt
```diff
@@ -83,7 +83,7 @@ sleep 1
 exec grim -g "0,0 2560x1440" ${LANG}/screenshot-04-settings.png
 
 # Post-processing
-set right direction -1.93 1.4 -1
+set right direction -1.95 1.45 -1
```
xytovl, line 86: The way github is setup, I have to manually click "approve and run", if you don't want to wait you should be able to enable actions on your fork so it runs immediately

Baplar: Oh, I think I haven’t set up actions on my fork yet, I should see to that. That would allow you not to have to run them for my "intermediate" commits ^^'

client/scenes/lobby_gui.cpp
Baplar: Ohh I see, thanks!

xytovl: Actually you need to use the macro, `tools/update_messages.sh` doesn't check the full names.

Baplar: Drat, the _cS macro returns a char*, I wanted to avoid having to manually allocate a string. Oh well. I’ll do the fix.

xytovl: You can add a macro for the variant you need, and wire it in the scripts, that's fine as well.
Make sure that scripts detect the missing string.

Baplar: Huh, turns out C++ can auto-cast char* into std::string if I just return _cS(x, y), neat. No need to update the scripts then!

OzzyHelix:
Hey @Baplar if you want I can test your builds I just need the github actions to be updated

Baplar:
> Hey @Baplar if you want I can test your builds I just need the github actions to be updated

Sure!

The last couple commits were mostly about formatting stuff, for testing the feature you can grab the last workflow that was performed here: https://github.com/WiVRn/WiVRn/actions/runs/14430426756
Or alternatively grab the last one on my fork, but I’m not sure how the APK signing will have behaved: https://github.com/Baplar/WiVRn/actions/runs/14432502250

Only the client was modified by my code, I didn’t touch the server, so you only need to grab the APK.

client/scenes/lobby_gui.cpp
xytovl: The macro is defined as `#define _cS(c, x) boost::locale::pgettext(c, x).c_str()`

I'll still make a new one `#define _c(c, x) boost::locale::pgettext(c, x)` to return an `std::string` and avoid the string copy.
You can't change the function to return a `const char*` as it would be a reference to a temporary.

xytovl:
It now all looks good to me, either you want to rework that in relevant commits such as layer settings and sgsr, or I'll just squash it all into a single commit.

Baplar:
> It now all looks good to me, either you want to rework that in relevant commits such as layer settings and sgsr, or I'll just squash it all into a single commit.

I’ll try to split/merge it in 2 commits, and if I can’t get something satisfactory I’ll squash it myself

PR #333, opened 2025-05-12
Add Pride WiVRn icon to the SVG file :3

yayameowmeow:
Adds a new group folder into wivrn.svg for a Pride themed icon. hopefully does not break a bunch of stuff :3


xytovl:
Are there many shared elements between normal and pride version?

If it's just for the nightly theme, I would prefer 2 files: wivrn.svg and wivrn-pride.svg.

yayameowmeow:
No shared elements, so can be its own file, just assumed youd wanted the icons in one file to change the groups off or on like is done with the nightly background. :cat:

PR #335, opened 2025-05-13
Update spanish translations

RiQuY:
Changes
- [feat(locale): Update spanish translations](https://github.com/WiVRn/WiVRn/commit/484d475da5f4002f14b17b3c9f1e1e3974806c27)
- [feat(locale): Add autogenerated locale files for it,ja and pt](https://github.com/WiVRn/WiVRn/commit/7628c04d5c1cf0f8c53c3029397a0e855928587f)

To update the translations I ran the `./tools/update_messages.sh` command and then filled the empty "msgstr" keys.

### DIFF
--- a/locale/es/wivrn.po
+++ b/locale/es/wivrn.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: WiVRn 0.15\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-04-11 20:21+0200\n"
+"POT-Creation-Date: 2025-05-13 21:31+0200\n"
 "PO-Revision-Date: 2024-11-05 19:29+0100\n"
 "Last-Translator: Sergio Navarro <respuit@protonmail.com>\n"
 "Language: Spanish\n"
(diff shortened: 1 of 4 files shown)


locale/es/wivrn.po
```diff
@@ -37,23 +37,38 @@ msgstr "No se puede conectar a {} ({}): {}"
 msgid "Waiting for connection"
 msgstr "Esperando conexión"
 
-#: client/scenes/lobby_gui.cpp:174
+#: client/scenes/lobby_gui.cpp:131
+msgctxt "openxr_post_processing"
+msgid "Normal"
+msgstr "openxr_post_processing"
```
xytovl, line 43: msgctx is the "context" of the text, the text that should be translated is "Normal" (for openxr post-processing settings)

xytovl:
It would be better to not change non-es translation files at all. I can handle that if you prefer.

locale/es/wivrn.po
RiQuY: I fixed the translations with msgctxt and removed the locales for ja/pt from the PR in the last commit.

PR #336, opened 2025-05-13
WIP: Pico 4 Pro face tracking

ImSapphire:


client/xr/pico_face_tracker.cpp
```diff
@@ -0,0 +1,52 @@
...
+	xrStopEyeTrackingPICO = inst.get_proc<PFN_xrStopEyeTrackingPICO>("xrStopEyeTrackingPICO");
+	xrSetTrackingModePICO = inst.get_proc<PFN_xrSetTrackingModePICO>("xrSetTrackingModePICO");
+	xrGetFaceTrackingDataPICO = inst.get_proc<PFN_xrGetFaceTrackingDataPICO>("xrGetFaceTrackingDataPICO");
+
+	CHECK_XR(xrStartEyeTrackingPICO(s));
+	CHECK_XR(xrSetTrackingModePICO(s, XR_TRACKING_MODE_FACE_BIT_PICO));
+}
+xr::pico_face_tracker::~pico_face_tracker()
+{
+	CHECK_XR(xrStopEyeTrackingPICO(s, XR_TRACKING_MODE_FACE_BIT_PICO));
```
xytovl, line 39: CHECK_XR throws in case of error, in the destructor we should just log if it fails.

common/wivrn_packets.h
```diff
@@ -286,6 +288,13 @@ struct tracking
 		bool lip_active;
 	};
 	std::optional<htc_face> face_htc;
+
+	struct pico_face
+	{
+		std::array<float, XR_BLEND_SHAPE_COUNT_PICO> weights;
+		bool is_valid;
+	};
+	std::optional<pico_face> face_pico;
```
xytovl, line 297: We should change that to `std::variant<std::monostate, fb_face2, htc_face, pico_face>` instead of 3 optional fields. The client code already assumes that it is at most one of them.

common/xr/pico_eye_types.h
xytovl, line 1: What is the origin of this file? Just to be sure who owns the copyright and if it can be redistributed under GPL-3+

server/driver/wivrn_pico_face_tracker.cpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+static xrt_result_t wivrn_pico_face_tracker_get_face_tracking(struct xrt_device * xdev,
+                                                              enum xrt_input_name facial_expression_type,
+                                                              int64_t at_timestamp_ns,
+                                                              struct xrt_facial_expression_set * out_value);
+
+wivrn_pico_face_tracker::wivrn_pico_face_tracker(xrt_device * hmd,
+                                                 wivrn::wivrn_session & cnx) :
+        xrt_device{}, cnx(cnx)
```
xytovl, line 55: It is possible to initialize most members in the `xrt_device{}` constructor.

Check wivrn_hmd.cpp for the helper `method_pointer` template to bind methods directly, so that you can skip the static functions and casts.

server/driver/wivrn_pico_face_tracker.h
```diff
@@ -0,0 +1,55 @@
...
+#include "wivrn_packets.h"
+#include "xr/pico_eye_types.h"
+#include "xrt/xrt_defines.h"
+#include "xrt/xrt_device.h"
+
+#include "history.h"
+
+#include <array>
+#include <cmath>
+#include <openxr/openxr.h>
```
xytovl, line 34: Not all includes seem to be needed in the header file.

Only keep what is directly used in this file, move to the cpp what is only used in the implementation.

common/xr/pico_eye_types.h
ImSapphire: The typedefs are from https://pico.crx.moe/docs/pico-openxr/eye-tracking#types and https://pico.crx.moe/docs/pico-openxr/eye-gaze-interaction#types
I don't think there is any copyright information listed anywhere on the website. I'll e-mail the author

server/driver/wivrn_pico_face_tracker.cpp
ImSapphire: That constructor is mostly copied from wivrn_htc_face_tracker or wivrn_fb_face2_tracker, I forget which one. Should those be cleaned up as well?

common/wivrn_packets.h
ImSapphire: [`b3c4665`](https://github.com/WiVRn/WiVRn/pull/336/commits/b3c4665a70ac4a9f032f188b4f2e8b6bd767c816)

server/driver/wivrn_pico_face_tracker.cpp
ImSapphire: [`193e731`](https://github.com/WiVRn/WiVRn/pull/336/commits/193e7318fe0a45693c2b99023d07e472950ab22c)

common/xr/pico_eye_types.h
xytovl: Let's include the link as a comment, so that it's easier to track where it comes from

client/xr/pico_face_tracker.cpp
```diff
@@ -0,0 +1,193 @@
...
+		out_expressions.is_valid = false;
+		out_expressions.is_eye_following_blendshapes_valid = false;
+		return;
+	}
+
+	if (face_tracking.time == 0)
+	{
+		out_expressions.is_valid = false;
+		out_expressions.is_eye_following_blendshapes_valid = false;
+		return;
```
xytovl: I would set the valid flags to false at the beginning of the method, so that error returns can be just `return`, without a `goto`, including the non initalized case (`!xrGetFaceTrackingDataPICO`).

client/xr/pico_face_tracker.cpp
```diff
@@ -0,0 +1,193 @@
...
+xr::pico_face_tracker::~pico_face_tracker()
+{
+	if (started)
+		stop();
+}
+
+void xr::pico_face_tracker::start()
+{
+	if (started)
+		throw std::invalid_argument("face tracking already started");
```
xytovl, line 49: There is no need to consider this an error condition, just use an early return. (same for stop)

common/xr/to_string.h
```diff
@@ -0,0 +1,62 @@
...
+	case name:                  \
+		return #name;
+#define XR_ENUM_STR(enumType)                                                                 \
+	constexpr const char * to_string(enumType e)                                          \
+	{                                                                                     \
+		switch (e)                                                                    \
+		{                                                                             \
+			XR_LIST_ENUM_##enumType(XR_ENUM_CASE_STR) default : return "Unknown"; \
+		}                                                                             \
+	}
```
xytovl, line 41: Note to self: implement [`std::formatter<enumType>`](https://en.cppreference.com/w/cpp/utility/format/formatter.html)

xytovl (approved):

PR #338, opened 2025-05-14
Add CMake flag and toggle for pride logo

ToolMan2k:
This is a companion commit for @yayameowmeow's pride month logo. It gives options to toggle the easter egg, either at build or run time, using a CMake flag and a toggle button respectively.

https://github.com/WiVRn/WiVRn/pull/333

CMake flag is **WIVRN_CLIENT_PRIDE** and is set to ON by default.
New toggle setting is **pride_flag_enabled** and is set to ON by default.

When toggling **WIVRN_CLIENT_PRIDE** to OFF, it will skip the inclusion of the respective assets and code at build time.
When toggling **pride_flag_enabled** in the settings, it will take effect after the next launch of WiVRn. 

This could benefit WiVRn users who lives in places where LGBT culture is either frowned upon or straight out illegal. South Carolina or Saudi Arabia would be two such instances respectively.

I've also included translations for every available languages. However I cannot ensure the accuracy of languages other than french. I'm also debating on how should every settings be called. You are more than welcome to pitch ideas and replace such translations.


xytovl:
The easter egg is only in the about screen, this should not endanger anyone.

I wanted to share a small message of acceptance and support, I do not plan to integrate a toggle to disable it.

mitsuami-megane:
as i understand, the easter egg doesn't come into view on the splash screen. the splash screen will use the regular logo, so will the icon.

the about page is not something that is normally used on the daily, so making a toggle that's actually more visible than the easter egg does not make much sense.

xytovl:
I don't see the need to have a toggle for this.

The splash screen (and icon) are set at build time and cannot be affected by configuration or date.

If you think I am wrong and the toggle will really be needed, the only place where it would make sense is in the about screen itself, something like "enable seasonal mascot" at the bottom. But as I said earlier I don't plan on merging that feature.

PR #342, opened 2025-05-20
Add config options to enable debug window and steamvr_lh

ImSapphire:
The debug window and SteamVR tracking toggles will only be visible if the features are enabled during build.


server/main.cpp
```diff
@@ -298,9 +296,20 @@ void start_server()
 
 		setenv("AMD_DEBUG", "lowlatencyenc", false);
 
+		ipc_server_main_info server_info{
+		        .udgci = {
+		                .window_title = "WiVRn",
+#ifdef WIVRN_FEATURE_DEBUG_GUI
```
xytovl, line 302: It should be `#if`

flake.lock
xytovl, line 1: Is it possible to have a CI job to check this file validity?

ImSapphire: I'm not sure what you mean by validity, the file is only touched by the Nix CLI tool. Unless you mean checking if it's up to date? A CI job could probably auto-bump this, but I don't know if that would really be worth it

xytovl: What does this (and the other nix file) actually do?

If the file is commited, we should expect that it works for what it is supposed to do, otherwise it would just confuse people who will try it.

ImSapphire: The `flake.nix` defines a development shell with the tooling required to work on WiVRn (build dependencies, clangd and clang-format), that's about it. The `flake.lock` just specifies the locked versions of the dependencies (defined in `inputs`) so that it's reproducible and all that

xytovl: Can we spin this commit into a separate PR, and include something like https://github.com/marketplace/actions/install-nix#usage-with-flakes so that we can have nix CI as well?

ImSapphire: > Can we spin this commit into a separate PR, and include something like https://github.com/marketplace/actions/install-nix#usage-with-flakes so that we can have nix CI as well?

#345

Issue #343, opened 2025-05-21
[Bug]: v0.25 Tag Doesn't Patch Cleanly (--fuzz=0)


damaestro:
### Describe the bug

When updating the Fedora package, the patches do not cleanly apply. We require `--fuzz=0`.

```
+ echo 'Patch #6 (0006-Use-mipmaps-for-distortion-shader.patch):'
+ /usr/bin/patch --no-backup-if-mismatch -f -p1 --fuzz=0
patching file src/external/CMakeLists.txt
Hunk #1 succeeded at 215 (offset 4 lines).
patching file src/external/nvpro_pyramid/CMakeLists.txt
patching file src/external/nvpro_pyramid/LICENSE
patching file src/external/nvpro_pyramid/nvpro_pyramid.glsl
patching file src/external/nvpro_pyramid/nvpro_pyramid_dispatch.hpp
patching file src/external/nvpro_pyramid/srgba8_mipmap_fast_pipeline.comp
patching file src/external/nvpro_pyramid/srgba8_mipmap_general_pipeline.comp
patching file src/external/nvpro_pyramid/srgba8_mipmap_preamble.glsl
patching file src/xrt/auxiliary/vk/vk_helpers.c
patching file src/xrt/compositor/CMakeLists.txt
patching file src/xrt/compositor/main/comp_renderer.c
patching file src/xrt/compositor/main/comp_settings.c
patching file src/xrt/compositor/main/comp_settings.h
patching file src/xrt/compositor/render/render_compute.c
patching file src/xrt/compositor/render/render_gfx.c
patching file src/xrt/compositor/render/render_interface.h
patching file src/xrt/compositor/render/render_resources.c
patching file src/xrt/compositor/render/render_shaders.c
patching file src/xrt/compositor/shaders/distortion.comp
patching file src/xrt/compositor/util/comp_mipmap.cpp
patching file src/xrt/compositor/util/comp_mipmap.h
patching file src/xrt/compositor/util/comp_render.h
patching file src/xrt/compositor/util/comp_render_cs.c
patching file src/xrt/compositor/util/comp_render_helpers.h
Hunk #1 succeeded at 149 (offset 4 lines).
Hunk #2 FAILED at 172.
1 out of 2 hunks FAILED -- saving rejects to file src/xrt/compositor/util/comp_render_helpers.h.rej
patching file src/xrt/compositor/util/comp_scratch.c
patching file src/xrt/compositor/util/comp_scratch.h
[...]

PR #344, opened 2025-05-21
Disable Fuzzy Patching

damaestro:
Disable fuzz when patching in relation to #343


xytovl:
This is a bad idea, it makes bisection of the Monado version harder and provides no benefit except passing some arbitrary rule.

If we want to enforce clean patches, this should be a CI job instead, or an optional flag.

damaestro:
@xytovl okay, makes sense. I just wanted this for reference to #343

PR #345, opened 2025-05-21
Add Nix devshell, package, and CI step

ImSapphire:
Spun off of #342
I'm not sure if having the package here is the best idea since it would mean extra upkeep with adding dependencies and keeping this Monado version in sync as well. If desired, I could remove the package override, keeping the devshell making the CI step only eval the flake.


xytovl:
To sum up the discussion on discord, the ideal situation is to have files that rarely require updates (not even on Monado updates) and CI job to validate file correctness and that it can build the software.
If that is not possible, maintaining the file should not require any nix environment, so that anyone can make a pull request to update Monado.

xytovl:
Can you test a dummy change to the revision to see how the CI failure looks like? If it's clear enough we can merge it

flake.nix
```diff
@@ -0,0 +1,58 @@
...
+          postUnpack = null;
+          prePatch = ''
+            ourMonadoRev="${finalAttrs.monado.src.rev}"
+            theirMonadoRev=$(cat monado-rev)
+            if [ ! "$theirMonadoRev" == "$ourMonadoRev" ]; then
+              echo "Our Monado source revision doesn't match CMakeLists.txt." >&2
+              echo "  theirs: $theirMonadoRev" >&2
+              echo "    ours: $ourMonadoRev" >&2
+              return 1
+            fi
```
xytovl, line 34: Can this still happen?

ImSapphire: No, I'll remove that bit

ImSapphire:
strange, I think it cached the result because the hash didn't change. that's quite annoying. It should have printed something like:
```
error: hash mismatch in fixed-output derivation '/nix/store/01541mcfgrl8fssab1bk5s5y9f48b5lh-source.drv':
         specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
            got:    sha256-OV9/W6E3hVemZhBU9l7+yruP89m8ucZASEGEUzmgLRo=
```

xytovl:
> strange, I think it cached the result because the hash didn't change. that's quite annoying. It should have printed something like:
> 
> ```
> error: hash mismatch in fixed-output derivation '/nix/store/01541mcfgrl8fssab1bk5s5y9f48b5lh-source.drv':
>          specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
>             got:    sha256-OV9/W6E3hVemZhBU9l7+yruP89m8ucZASEGEUzmgLRo=
> ```

Does this means that simply changing the git revision would not actually update the monado sources when using nix tools?

ImSapphire:
Yes, it would only refetch the monado source if the hash is different from before. When updating you would have to set hash to a blank string, then wait for the error and put the right hash in, which is quite inconvenient

ImSapphire:
@xytovl It now re-fetches when any attribute in the fetcher changes, so CI properly fails now!
```
copying path '/nix/store/42dlyn8qbxdp55gxd35xymik0nhy0hdw-gtk+3-3.24.49' from 'https://cache.nixos.org/'...
copying path '/nix/store/904qf4rqvs1821q9d4lafjdzh5j9j4hh-util-linux-2.41-dev' from 'https://cache.nixos.org/'...
error: hash mismatch in fixed-output derivation '/nix/store/rsix4a0cnbx05arijz2zmk3qsy80pdl0-source-salted-jd5sgii5pix3.drv':
         specified: sha256-Bus9GTNC4+nOSwN8pUsMaFsiXjlpHYioQfBLxbQEF+0=
            got:    sha256-OV9/W6E3hVemZhBU9l7+yruP89m8ucZASEGEUzmgLRo=
error: 1 dependencies of derivation '/nix/store/m2si44k4cviqg9jc86dhs3787vq81j7c-source-salted-jd5sgii5pix3-patched.drv' failed to build
error: 1 dependencies of derivation '/nix/store/yr0l3zb247qngip0v1mqp1zr8ajnrsgh-wivrn-next.drv' failed to build
Error: Process completed with exit code 1.
```

xytovl:
The failure is good enough for people to fix it easily.

Will `invalidateFetcherByDrvHash` always redownload sources? For CI it's fine but I let you decide what works best for local development.

ImSapphire:
@xytovl it only redownloads when an argument to the fetcher changes, so if the rev or url hasn't changed it will use the cached one

PR #351, opened 2025-05-27
Make patch commits reproducible

kitlith:
The timestamp reproducibility will be useful for building wivrn for multiple architectures in separate build trees.

Committer reproducibility could be useful for ensuring compatibility between wivrn builds across multiple *machines* with different default commit authors, but is probably not desirable if the script is ever used outside of the cmake build.

### DIFF
--- a/patches/apply.sh
+++ b/patches/apply.sh
@@ -1,9 +1,9 @@
 #!/bin/sh
 
 if [ -d ".git" ] ; then
-	git config user.email > /dev/null || git config user.email "git-am@invalid"
-	git config user.name > /dev/null || git config user.name "no name"
-	git config --local commit.gpgsign false
+	git config --worktree user.email "git-am@invalid"
+	git config --worktree user.name "no name"
+	git config --worktree commit.gpgsign false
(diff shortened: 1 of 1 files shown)


kitlith:
A test you can run before and after this commit to view what's different and ensure it's been fixed:

```sh
mkdir build-a build-b
cmake -B build-a
cmake -B build-b
# diff the raw commit metadata of the first patch between the two build trees
diff <(git -C build-a/_deps/monado-src show --pretty=raw HEAD~10) <(git -C build-b/_deps/monado-src show --pretty=raw HEAD~10)
```

before this patch, diff output something like this:
```diff
1c1
< commit 194e495e7ed28c4d51c133b720975a8ff34a5e0b
---
> commit e4a178569afe2f211f11e691c79da2db58c10b05
5c5
< committer Kitlith <kitlith@kitl.pw> 1748316800 -0700
---
> committer Kitlith <kitlith@kitl.pw> 1748316814 -0700
```

after this patch, the commits are identical and have a committer line of:
```
committer no name <git-am@invalid> 1729603399 +0200
```

PR #360, opened 2025-06-07
Fix foveation eye convergence, account for async vertical fov

galister:


galister:
scale 0.35 seems to work well with convergence distance of 0.5, giving us a sweet spot size and placement that works comfortably both when looking up close and far away.

PR #363, opened 2025-06-09
wivrnctl toggle-performance-graph

galister:
Adds a `wivrnctl toggle-performance-graph` sub-command. This allows toggling the plots for people who don't use standalone controllers, or people who prefer to use a shortcut from their computer (thumbstick actuation lifespans are scary).

Behavior summary:
- the imgui context and swapchain are created when needed
- client hotkey combo is only used if "Show Performance Metrics" was enabled before connecting
- the client settings stay untouched (on next connect, plots will show depending on "Show Performance Metrics")

### DIFF
--- a/client/scenes/stream.cpp
+++ b/client/scenes/stream.cpp
@@ -342,33 +342,40 @@ std::shared_ptr<scenes::stream> scenes::stream::create(std::unique_ptr<wivrn_ses
 	return self;
 }
 
+void scenes::stream::setup_imgui()
+{
+	swapchain_imgui = xr::swapchain(
+	        session,
+	        device,
+	        swapchain_format,
+	        1500,
+	        1000);
+
+	imgui_context::viewport vp{
+	        .space = xr::spaces::view,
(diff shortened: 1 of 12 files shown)


xytovl:
I think it would be better to have a boolean property and be able to know the state.
This would make it easier to eventually have it in the dashboard.

galister:
That's fair, but I think while the hmd is connected, the user will notice if it's on :smile: 

I was trying to avoid having to do the back n forth between the client, but if you say it's worth it, might as well.

If it were up to me, I'd just add a Toggle Metrics View button that doesn't change label.

xytovl:
My long term idea would be to replace the in-headset plots with a proper UI.

The two sticks gestures would show the UI, where you could have some settings (post-processing, ability to pause some trackers, audio).

We can do the quick and dirty dbus toggle and drop it when we implement the in-stream UI.

xytovl:
closing as in-stream GUI was merged

PR #369, opened 2025-06-15
Vive Ultimate Tracker support

ImSapphire:
Supersedes #281.


client/application.h
```diff
@@ -140,6 +141,8 @@ class application : public singleton<application>
 
 	bool fb_body_tracking_supported = false;
 	xr::fb_body_tracker fb_body_tracker;
+	bool htc_body_tracking_supported = false;
+	xr::htc_body_tracker htc_body_tracker;
 	bool pico_body_tracking_supported = false;
 	xr::pico_body_tracker pico_body_tracker;
```
xytovl: It looks like we should unify the body trackers more.

The system used by passthrough may be appropriate: a single `std::variant` with possible implementations, where `std::monostate` implies no support.

If interfaces are identical, inheritance might be a suitable option as well.

client/xr/htc_body_tracker.h
```diff
@@ -0,0 +1,66 @@
...
+class vive_xr_tracker
+{
+	bool is_active = false;
+	xr::space space;
+
+public:
+	vive_xr_tracker(xr::space && s);
+	void set_active(bool active);
+	bool get_active() const;
+	xr::space * get_space();
```
xytovl: I would return an `XrSpace` here

client/xr/htc_body_tracker.cpp
```diff
@@ -0,0 +1,155 @@
...
+	for (auto & path: paths)
+	{
+		XrInputSourceLocalizedNameGetInfo get_info{
+		        .type = XR_TYPE_INPUT_SOURCE_LOCALIZED_NAME_GET_INFO,
+		        .next = nullptr,
+		        .sourcePath = path,
+		        .whichComponents = XR_INPUT_SOURCE_LOCALIZED_NAME_USER_PATH_BIT,
+		};
+
+		roles.emplace_back(xr::details::enumerate<char>(xrGetInputSourceLocalizedName, s, &get_info));
```
xytovl: This creates a vector of `path.size()` elements, initialized to the default value (empty string)
```cpp
std::vector<std::string> roles(paths.size());
```
You should instead do
```cpp
std::vector<std::string> roles:
roles.reserve(paths.size());
```
Or do a `roles[i] = ...` instead of `emplace_back`

client/application.h
ImSapphire: I'm not really sure how to implement that, I tried reading the code for passthrough and it confuses me

xytovl: There's 2 things that I'd like to avoid here:

the growing number of members in application.h, either variant or inheritance could reduce it a single member.

the members don't match the expected states, we want to only have one body tracker type during a session, but the code accepts all 3 concurrently. I don't think a headset would implement all of them.

galister: I would make all of the body trackers implement the same interface and then we could lower the number of branches in the tracking thread as well.

client/application.cpp
```diff
@@ -1223,14 +1232,26 @@ void application::initialize()
...
+
+	if (body_tracker.index() == 0 && utils::contains_all(xr_extensions, std::array{XR_HTC_PATH_ENUMERATION_EXTENSION_NAME, XR_HTC_VIVE_XR_TRACKER_INTERACTION_EXTENSION_NAME}))
 	{
-		fb_body_tracker = xr_session.create_fb_body_tracker();
+		spdlog::info("    HTC body tracking support: true");
+		body_tracker = xr_session.create_htc_body_tracker();
 	}
 
-	if (pico_body_tracking_supported)
+	if (body_tracker.index() == 0 && utils::contains(xr_extensions, XR_BD_BODY_TRACKING_EXTENSION_NAME))
```
xytovl: Prefer `std::holds_alternative<std::monostate>` so that it doesn't depend on variant ordering

ImSapphire: that's just checking if the body_tracker is monostate, i think monostate will always be the first in the types anyway. I could still change it to `std::holds_alternative<std::monostate>(body_tracker)` if preferred

client/scenes/lobby_gui.cpp
```diff
@@ -757,16 +758,29 @@ void scenes::lobby::gui_settings()
...
+					                                             return _("Requires 'Hand and body tracking' to be enabled in the Quest movement tracking settings,\notherwise body data will be guessed from controller and headset positions");
+				                                             },
+				                                             [&](xr::htc_body_tracker &) {
+					                                             return _("Forwards positions of Vive Ultimate Trackers");
+				                                             },
+				                                             [&](xr::pico_body_tracker &) {
+					                                             return _("Uses the Pico Motion Trackers to track body joint positions");
+				                                             },
+				                                     },
+				                                     application::get_body_tracker());
```
xytovl: Does the lambda need to capture anything? the `&` can be removed.

Or just call `tooltip` inside the lambda, and don't return a value.

client/scenes/stream_tracking.cpp
```diff
@@ -360,7 +350,7 @@ void scenes::stream::tracking()
 							        locate_hands(application::get_right_hand(), world_space, t0 + Δt));
 					}
 
-					if (body_tracking != body_none and t0 >= last_body_sample + period and
+					if (!std::holds_alternative<std::monostate>(body_tracker) and t0 >= last_body_sample + period and
```
xytovl: Maybe move the test inside the `std::visit`, and we don't need to test for `std::monostate` as it already has its overload.

xytovl:
This looks fine to me, as soon as we get confirmation that it works fine for all 3 tracker types it can be merged.

ImSapphire:
@xytovl It's been tested on FB and HTC, I don't know if anyone can test the Pico one at the moment though. I think it's good enough to merge, I'll just squash the commits real quick

ImSapphire:
squashed

PR #374, opened 2025-06-20
Update pt_BR locale

LuckShiba:
Added missing translations and updated some old translations to make them sound more natural.


ImSapphire:
Have you verified the tooltip translations for 'Enable body tracking' and 'Enable lower body tracking' fit inside the viewport for the gui? If the line is too long, the tooltip will get cut off and won't be readable.

LuckShiba:
> Have you verified the tooltip translations for 'Enable body tracking' and 'Enable lower body tracking' fit inside the viewport for the gui? If the line is too long, the tooltip will get cut off and won't be readable.

Now that the workflows ran, I just tested and yes, it's too long. Will push a fix.

LuckShiba:
I couldn't get to build the app locally, will test after the workflows are approved

LuckShiba:
This was the last one broken.

xytovl:
Just add a comment if it's all fine, CI build are available.

Issue #379, opened 2025-06-23
[Bug]: Pico 4 Controllers are off


ghost:
### Describe the bug

I have recorded a video of controllers behavior outside and inside wivrn and it's clearly visible how much off they are inside wivrn.
Video: https://ody.sh/JEjXxO5RUm

### WiVRn Version

v0.25

### What GPU did the bug occur on?

NVIDIA

### What HMD did the bug occur on?

Pico 4

xytovl:
In order to further tweak the controller model position, you should try the same steps performed in #177, with details in https://github.com/WiVRn/WiVRn/issues/94#issuecomment-2465105210

It seems that grip_left and grip_right would need different offsets. The debug GUI does not have settings for different left and right values, so you may have to iterate by changing the cpp directly.

PR #380, opened 2025-06-24
Update hardware.cpp

ghost:
A quick fix for a lot more 1 to 1 controller offset


xytovl:
@Mhowser can you check if the offsets are more accurate compared to #94 #177? 
#379 shows a video about the incorrect offsets.

ghost:
From my testing these offsets are nearly one-to-one making them significantly better compared to those in the videos.
Hopefully the results will be consistent across different users as I've heard the Pico 4 sometimes has issues with offsets not applying the same way for everyone.

ghost:
Here I made a video so you can actually compare it https://ody.sh/U35HkX4hBT

PR #381, opened 2025-06-24
Split foveation environment variables

Baplar:


server/driver/wivrn_foveation.cpp
```diff
@@ -268,15 +279,21 @@ void wivrn_foveation::compute_params(
 		if (foveated_width < src[i].extent.w)
 		{
 			auto angle_x = convergence_angle(views[i].pose.position.x, e.x);
-			auto center = angles_to_center(view.x + angle_x + angle_offset.x * (i == 0 ? 1 : -1), fov.angle_left, fov.angle_right);
+			auto center = angles_to_center(view.x + angle_x, fov.angle_left, fov.angle_right);
```
Baplar, line 285: Given that we now have a more explicit way to specify that offset (the convergence distance) I removed the x component of the offset.

server/driver/wivrn_foveation.cpp
```diff
@@ -268,15 +279,21 @@ void wivrn_foveation::compute_params(
...
 		if (foveated_height < src[i].extent.h)
 		{
-			auto view_center = (views[i].fov.angleDown + views[i].fov.angleUp) / 2;
-			auto center = angles_to_center(-view.y - e.y + angle_offset.y + view_center, fov.angle_up, fov.angle_down);
+			auto angle_y = e.y;
+			if (is_zero_quat(gaze))
+			{
+				// Offset gaze towards the center of the fov if no eye tracking.
+				// Seems to work for most headsets, need to verify on some (HTC, Lynx)
+				angle_y -= (views[i].fov.angleDown + views[i].fov.angleUp);
```
Baplar, line 297: @galister checked that this seems to have the intended effect on most headsets, but we may need to add a 2nd part to that `if` depending on future findings.

server/driver/wivrn_foveation.cpp
```diff
@@ -142,10 +142,33 @@ static std::tuple<float, float> solve_foveation(float λ, float c)
...
+
 static xrt_vec2 yaw_pitch(xrt_quat q)
 {
-	if (q.x == 0 and q.y == 0 and q.z == 0 and q.w == 0)
-		return xrt_vec2{};
+	if (is_zero_quat(q))
+	{
+		// Add manual offset, useful if the default angle does not suit the user
+		return {0, get_angle_offset()};
+	}
```
galister: yaw_pitch is used with both view and gaze. i think let's pass a default value as arg instead

Baplar: Is it possible for the view argument to be a zero-value quaternion?
Regardless, I can just move the call to get_angle_offset into the conditional at line 290

galister: i would just keep it out of such a generically named func as yaw_pitch.

server/driver/wivrn_foveation.cpp
```diff
@@ -171,9 +194,25 @@ static float angles_to_center(float e, float l, float r)
 	return (e - l) / (r - l) * 2 - 1;
 }
 
+static float get_convergence_distance()
```
Baplar: More general question, what’s the best C++ idiomatic way to store the result of that method, in order not to re-fetch the var and re-convert it to float every call? My C++ is a bit rusty

server/driver/wivrn_foveation.cpp
```diff
@@ -220,30 +226,39 @@ static void fill_param_2d(
...
+
+static float get_convergence_distance()
+{
+	if (const char * cvar = std::getenv("WIVRN_FOVEATION_DISTANCE"))
+	{
+		std::string_view var(cvar);
+		float res;
+		auto e = std::from_chars(var.begin(), var.end(), res);
+		if (e.ec == std::errc())
+			return std::min(res, 0.05f);
```
xytovl: Did you mean std::max?

Baplar: Oops, indeed! That was a last-minute addition, I was probably too tired to still be writing code ^^'

Baplar:
```suggestion
			return std::max(res, 0.05f);
```

server/driver/wivrn_foveation.cpp
xytovl: For this to be physically correct, the middle would be `(up + down)/2`

Baplar: Correct in theory, but in practice removing the halving seems to fix the off-center issue we had been having on non eye tracked headsets.

I suspect some double-counting must be occurring in angles_to_center, given that we also use the fov angles there to normalize the value. But I haven't done the full maths to check if that's correct.

xytovl: If we take some [old research](https://apps.dtic.mil/sti/tr/pdf/AD0758339.pdf) p393, normal sight line is marked as 15° below horizontal.

This could be a viable default value, and we would disregard any headset configuration.

Baplar: That’s a fair point. However, that may already have been a design consideration for the headset vendors, considering many of them have more FOV down than up, so we would actually need to use lower values than 15° to achieve that effect. 
For example on Quest 3, the "center" of the FOV is already 6° down using the (up+down)/2 formula, which would require us to add an additional 9° offset (that explains why our manual testings ended up with 8° being a quite comfortable value for a fixed offset).

But the FOV values are different for every headset.
In particular, on Pico, the FOV is symmetrical, but using an offset of 0° (value computed by using the proposed formula) does result in the fovea being centered in a good spot according to galister’s testing.

So I think there is really merit to this formula. Whether it is due to a mathematical coincidence, or there is a possible formal explanation for this, I am not sure yet. I’ll try to crunch the numbers this evening when I have time.
But in the meantime, until we find a counter-example of a headset for which this formula results in a bad positioning of the fovea, I think we can keep it that way with a hefty warning comment. But this is not my call to make officially, of course.

galister: Big part of the reason that the Pico 4 is good as is is that the lenses are not great and the user wouldn't want to look too far down to avoid the distortion.

PR #385, opened 2025-06-27
Fix hands snapping to head when out of view on Quest + minor code cleanup

### LINKED ISSUE #310: [Bug]: Hand tracking desync when either hand goes out of the quest 3's detection range.
### Describe the bug

When using hand tracking when either your hands are obscured by an external object or too low to be in range of the quest 3 the hands go into the users facial area.

### Steps to Reproduce

1. After wivrn is connected with hand tracking enabled open VRChat
2. Move your hands out of the camera's range.

### Expected behavior

I expected the hands to stay in place until tracking is restored.

### WiVRn Version

2f195857e7eb2be43db1d2763cca4fd799f70943

### Server Config

[config.json](https://github.com/user-attachments/files/19661350/config.json)

### Server Logs

[wivrn-server.txt](https://github.com/user-attachments/files/19661417/wivrn-server.txt)

AMD

### Vulkan Info

[vulkaninfo.txt](https://github.com/user-attachments/files/19661371/vulkaninfo.txt)

Quest 3

### ADB Logs

_No response_

ImSapphire:
Works around #310 by sending blank controller poses when hand tracking data is invalid and the interaction profile is khr/simple_controller.
Fixes #310

### DIFF
--- a/client/scenes/stream_tracking.cpp
+++ b/client/scenes/stream_tracking.cpp
@@ -238,6 +238,7 @@ void scenes::stream::tracking()
 	std::vector<from_headset::tracking> tracking;
 	std::vector<from_headset::tracking> tracking_pool; // pre-allocated objects
 	std::vector<from_headset::hand_tracking> hands;
+	std::array<bool, 2> should_skip_simple_controllers{};
 	std::vector<from_headset::body_tracking> body;
 	std::vector<XrView> views;
 
@@ -245,8 +246,22 @@ void scenes::stream::tracking()
 	std::vector<serialization_packet> packets;
 
 	const bool hand_tracking = config.check_feature(feature::hand_tracking);
+	// Quest hand tracking creates a fake khr/simple_controller when hand tracking
+	// is enabled, let's not send it when hand tracking data is invalid
+	bool block_khr_simple_controller = false;
+	switch (guess_model())
+	{
+		case model::meta_quest_3:
+		case model::meta_quest_pro:
+		case model::meta_quest_3s:
+		case model::oculus_quest_2:
+			block_khr_simple_controller = hand_tracking;
+			break;
+		default:
+			break;
+	}
+
 	from_headset::face_type face_tracking = from_headset::face_type::none;
-	const bool body_tracking = config.check_feature(feature::body_tracking);
 	if (config.check_feature(feature::face_tracking))
 	{
 		if (application::get_fb_face_tracking2_supported())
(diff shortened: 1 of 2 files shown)


client/xr/hand_tracker.cpp
```diff
@@ -77,6 +77,10 @@ std::optional<std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT>> xr::
 	if (!locations.isActive)
 		return std::nullopt;
 
+	// bail if none of the joints are valid/tracked
+	if (!std::ranges::any_of(joints_pos, [](const auto & loc) { return loc.locationFlags != 0; }))
+		return std::nullopt;
+
```
xytovl, line 83: Is the headset returning something with `isActive = true` and invalid joints?

Per https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XrHandJointLocationsEXT
> If the returned isActive is true, the runtime must return all joint locations with both XR_SPACE_LOCATION_POSITION_VALID_BIT and XR_SPACE_LOCATION_ORIENTATION_VALID_BIT set. Although, in this case, some joint space locations may be untracked (i.e. XR_SPACE_LOCATION_POSITION_TRACKED_BIT or XR_SPACE_LOCATION_ORIENTATION_TRACKED_BIT is unset).

ImSapphire: I believe so, I wasn’t hitting the skip path until I added that check.

client/scenes/stream_tracking.cpp
```diff
@@ -327,31 +343,70 @@ void scenes::stream::tracking()
...
 							hands.emplace_back(
 							        t0,
 							        t0 + Δt,
 							        from_headset::hand_tracking::left,
-							        locate_hands(application::get_left_hand(), world_space, t0 + Δt));
+							        joints);
+						}
+						else
+						{
+							should_skip_simple_controllers[0] = false;
```
xytovl, line 363: This case is when the application on PC is not consuming hand tracking data, I suppose we will get the same issue.

Isn't there something in the grip/aim pose itself to see that it should be discarded?
I suppose it's not setting the "tracked" flags, in that case we can do the check with controller pose only.

ImSapphire: The quest is always setting `orientation_valid | position_valid | linear_velocity_valid | angular_velocity_valid | orientation_tracked | position_tracked` flags on the grip pose for the fake controller device, so I don't think there's any other way to do this.

PR #387, opened 2025-06-28
[refactor] Store face tracker in variant

ImSapphire:


xytovl:
We'll have to make a point release for the hand tracking fix and crash in foveation. I'll have a look at this PR once those are done.

common/wivrn_packets.h
```diff
@@ -198,7 +198,6 @@ enum face_type : uint8_t
 	none,
 	fb2,
 	htc,
-	pico,
```
xytovl, line 201: Changes in enum values are not picked up by automatic protocol version calculation.
You can bump the `protocol_revision` value to force version incompatibility.

PR #389, opened 2025-06-29
Forward XR_EXT_user_presence

### LINKED ISSUE #258: [Feature Request]: XR_EXT_user_presence
### Feature description

This allows clients to detect when the user has removed their HMD from their head and respond appropriately such as saving power or throttling or displaying some idle animation

### Reason for the feature

This adds capability for great power saving, resource management of clients, and for clients with avatar systems to display idle animations instead of crushing players into pretzels.

### Proposed implementation

The extension will need an implementation in Monado before it can be mapped from headset runtime through WiVRn.

### Additional context

https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_EXT_user_presence

ImSapphire:
Currently, the current user presence isn't sent to server on connect, so if the headset was taken off before stream start it will be out of sync on server.
Waiting on https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2507
Closes #258


client/scenes/stream.cpp
```diff
@@ -1277,6 +1279,30 @@ void scenes::stream::on_xr_event(const xr::event & event)
...
+				case XR_SESSION_STATE_VISIBLE:
+					network_session->send_control(from_headset::user_presence_changed{
+					        .present = false,
+					});
+					break;
+				case XR_SESSION_STATE_FOCUSED:
+					network_session->send_control(from_headset::user_presence_changed{
+					        .present = true,
+					});
+					break;
```
xytovl: We should not send user presence packets in this case.
The server could handle both events in the same way, but there's no reason to merge them at the protocol level.

server/driver/wivrn_hmd.h
```diff
@@ -52,6 +52,7 @@ class wivrn_hmd : public xrt_device
 
 	view_list views;
 	from_headset::battery battery{};
+	bool presence{true};
```
xytovl: We can use an atomic and not use the mutex.

server/driver/wivrn_hmd.cpp
```diff
@@ -222,4 +235,11 @@ void wivrn_hmd::update_visibility_mask(const from_headset::visibility_mask_chang
 	auto m = visibility_mask.lock();
 	m->at(mask.view_index) = mask.data;
 }
+
+void wivrn_hmd::update_presence(bool presence)
+{
+	std::lock_guard lock(mutex);
+	U_LOG_D("Updating hmd presence to %s", presence ? "true" : "false");
```
xytovl: Events should not happen that often, I would increase level to info and rephrase to "user presence" instead of "hmd presence"

patches/monado/0012-st-oxr-Implement-XR_EXT_user_presence.patch
xytovl, line 1: Linking the associated merge request https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2507

client/scenes/stream.cpp
ImSapphire: > The server could handle both events in the same way, but there's no reason to merge them at the protocol level.

Should I forward the session state changed event as well, and on server fire the presence changed event to clients?

xytovl: It can be an independent change, or part of this one, as you prefer.

We could even decide on server to still issue an user presence event if changing the session state breaks too much of the xrt_system

ImSapphire: If we fire presence events based on state change it causes a new issue: when you put the quest into sleep, take it off, then wake it, the presence event with `isUserPresent=false` is sent before the state transition to `FOCUSED`, which causes presence to become out of sync on server. I'm not sure how to best solve this, maybe keep track of 'actual' presence changes in a separate variable?

xytovl: If we can stay as close as possible to the remote runtime that would be better: synchronize states and presence independently one from the other.

Let's just remove session state events from this pull request and only send presence information.

server/driver/wivrn_hmd.h
xytovl: Is there an issue with just `std::atomic<bool>`?

ImSapphire: oops, I just copied what was below, I'll change that

server/driver/wivrn_hmd.cpp
```diff
@@ -147,6 +149,16 @@ void wivrn_hmd::update_battery(const from_headset::battery & new_battery)
 	battery = new_battery;
 }
 
+xrt_result_t wivrn_hmd::get_presence(bool * out_presence)
+{
+	if (!supported.presence)
+		return XRT_ERROR_NOT_IMPLEMENTED;
```
xytovl: Isn't this supposed to be managed by Monado?
I would drop the check entirely, or change it to an assertion.

ImSapphire: done

PR #396, opened 2025-07-05
WIP: Forward XR_EXT_hand_interaction

### LINKED ISSUE #238: [Bug]: Hand actions dont work
### Describe the bug

There seems to be no hand profiles? or something like that, sorry i dont know much about openxr

Here is a video of how it should work, running natively on the quest 2: https://www.youtube.com/watch?v=-6u2ThRX24Q
Here is a video of the bug, running on my pc trough WiVRn: https://youtube.com/watch?v=n2IJW54qJJQ

### Steps to Reproduce

Here is an example app i used to test this
- Download the openxr hand tracking demo from https://github.com/godotengine/godot-demo-projects/tree/master/xr/openxr_hand_tracking_demo
- Open it in godot and run it with WiVRn
- Try to grab the cubes with hand... doesnt work

(For the native quest app do the same steps but)
- Install the godot vendor plugin from the asset store or (https://github.com/GodotVR/godot_openxr_vendors)
- Enable Hand support in the export settings
- Install it on the Quest 2, and it works perfectly (using ext hand_interaction_ext)

### Expected behavior

Pinch and other gestures dont work, the profile seems to stay as touch_controller even when using hands

### WiVRn Version

0.22

### Server Config

_No response_

### Server Logs

_No response_

AMD

### Vulkan Info

_No response_

Quest 2

### ADB Logs

_No response_

ImSapphire:
Poke, pinch, and grasp pose don't seem to be making it to PC....
Closes #238


ImSapphire:
~~This also regresses #310~~

client/scenes/stream_tracking.cpp
```diff
@@ -306,19 +289,61 @@ void scenes::stream::tracking()
...
+	
+	auto enabled = [this](const to_headset::tracking_control & control, device_id id) -> bool {
+		switch (id)
+		{
+			case device_id::HEAD:
+			case device_id::EYE_GAZE:
+				return true;
+			case device_id::LEFT_PINCH_POSE:
+			case device_id::LEFT_POKE:
+				return interaction_profiles[0].load() == interaction_profile::ext_hand_interaction_ext;
```
xytovl: This should also be a `control.enabled[...]`, so we can skip polling and sending if the server is not consuming the data.

ImSapphire: done

server/driver/wivrn_controller.cpp
```diff
@@ -237,7 +238,13 @@ static input_data map_input(device_id id)
 		default:
 			break;
 	}
-	throw std::range_error("bad input id " + std::to_string((int)id));
+	// If the headset supports hand_interaction_ext, upon switch
+    // to/from hand tracking we may get the inputs packet before
+    // the interaction profile change, so if we get a bad input
+    // just return an invalid index so we can ignore it
+	U_LOG_D("wivrn_controller: bad input id %s", std::string(magic_enum::enum_name(id)).c_str());
```
xytovl: This should not happen, the active interaction profile is in the `trackings` packet, and is read before the individual `tracking`.

ImSapphire: inputs are in a different packet sent separately from the tracking loop, I was getting crashes because of a bad input when switching to my controllers from hand tracking

server/driver/wivrn_session.h
```diff
@@ -106,7 +107,13 @@ class wivrn_session : public xrt_system_devices
 
 	wivrn_hmd hmd;
 	wivrn_controller left_hand;
+	int32_t left_hand_index;
+	wivrn_hand_interaction left_hand_interaction;
+	int32_t left_hand_interaction_index;
```
xytovl: I would rename `left_hand` to `left_controller`
Eventually we'll split out the hand tracking from the controller class

server/driver/wivrn_session.cpp
```diff
@@ -418,8 +424,18 @@ void wivrn_session::operator()(from_headset::trackings && tracking)
 		        std::string(magic_enum::enum_name(roles.right_profile)).c_str(),
 		        std::string(magic_enum::enum_name(right)).c_str());
 		std::lock_guard lock(roles_mutex);
+		if (left == XRT_DEVICE_EXT_HAND_INTERACTION)
+			roles.left = left_hand_interaction_index;
+		else
+			roles.left = left_hand_index;
```
xytovl: We should only assign to `roles.left` if the previous role is one of the WiVRn managed ones.
In case of lighthouse devices, we could have the left hand on WiVRn, while the right one is lighthouse.

In that case, only the `roles.left` shall be updated, even if both interaction profiles changed.
The `XRT_DEVICE_INVALID` profile can be returned by the headset if there is no active interaction profile, so this is not a sufficient check either.

server/driver/wivrn_controller.cpp
```diff
@@ -696,7 +703,7 @@ void wivrn_controller::set_inputs(const from_headset::inputs & inputs, const clo
 	{
 		int64_t last_change_time = input.last_change_time ? clock_offset.from_headset(input.last_change_time) : 0;
 		auto [index, type, device] = map_input(input.id);
-		if (device != device_type)
+		if (index == WIVRN_CONTROLLER_INPUT_INVALID || device != device_type)
```
xytovl: Technically not needed, device type is unknown in this case

server/driver/wivrn_session.cpp
```diff
@@ -484,8 +515,21 @@ void wivrn_session::operator()(from_headset::inputs && inputs)
...
+			left_controller.set_inputs(inputs, offset);
+	}
+
+	if (roles.right_profile != XRT_DEVICE_INVALID)
+	{
+		if (roles.right == right_hand_interaction_index)
+			right_hand_interaction.set_inputs(inputs, offset);
+		else
+			right_controller.set_inputs(inputs, offset);
+	}
```
xytovl: Just check `roles.left`, with an if/else if. The `roles.left_profile` should not be relevant here.

server/driver/wivrn_controller.cpp
```diff
@@ -811,6 +818,17 @@ void wivrn_controller::set_derived_pose(const from_headset::derived_pose & deriv
...
 {
+	const bool left = device_type == XRT_DEVICE_TYPE_LEFT_HAND_CONTROLLER;
+
+	xrt_system_roles roles;
+	cnx->get_roles(&roles);
+	if (cnx->get_xdev(left ? roles.left : roles.right) != this)
+	{
+		// we don't need to disable aim/grip/palm, hand_interaction will
+		// end up disabling those when unused
+		return;
```
xytovl: The first update_tracking after a change will disable them, because the last stored request will be too old for the new data.

Also, the get_roles/get_xdev dance is more complicated than necessary.
there's already some dispatch logic in wivrn_session already, in the end it may be better to have all of them in wivrn_session.

The alternative would be for the controller/hand_interaction object to know if it's active and move all the discard logic there, the current split is awkward.

server/driver/wivrn_session.cpp
```diff
@@ -418,8 +424,26 @@ void wivrn_session::operator()(from_headset::trackings && tracking)
...
 		        std::string(magic_enum::enum_name(right)).c_str());
 		std::lock_guard lock(roles_mutex);
+
+		// don't change role when hand from other driver is used
+		if (roles.left == left_hand_interaction_index || roles.left == left_controller_index)
+		{
+			if (left == XRT_DEVICE_EXT_HAND_INTERACTION)
+				roles.left = left_hand_interaction_index;
+			else
+				roles.left = left_controller_index;
```
xytovl: We may have to refresh the xxx_list objects in the controller/hand_interaction so that the `last_request` won't trigger a deactivation

ImSapphire:
Tested tracking control with this version, it seems to be handled correctly, it doesn't switch to disabled then immediately back enabled on interaction profile switch.

xytovl (approved):
looks good, thanks for the big feature!

ImSapphire:
@xytovl squashed

ImSapphire:
@xytovl Is this good to merge?

PR #397, opened 2025-07-05
vaapi: 10-bit hevc/av1 encoding

galister:
10-bit encoding is now supported for H265 and AV1 on VAAPI only. Tested H265+AV1 on AMD.

Added a new boolean `use_10bit` to encoder config. This must be the same for all encoders. The pass-through encoder inherits the encoder+codec+10bit settings from the first encoder if its 10-bit setting is different from the color encoders.

I ended up not separating the alpha channel for now. I'm not sure if the performance benefits would make it worth the extra maintenance burden. For example, I can't think of use cases where pass-through would be on, but the decoding latency of a 6Mib/s stream would matter.

---

I'm not looking into the dashboard changes just yet, mostly because we'll also want to think about how to make this feel intuitive. I think the 10-bit option should be a global switch that affects all encoders

I would probably ship a 10-bit preset at the very least, but there are other challenges we'll need to consider for the UI, such as:
- making sure people don't mix 10-bit with non-10-bit encoders
- making sure people don't select codect/encoders that do not support 10-bit

### DIFF
--- a/server/driver/configuration.cpp
+++ b/server/driver/configuration.cpp
@@ -135,6 +135,7 @@ configuration::encoder parse_encoder(const nlohmann::json & item)
 	SET_IF(codec);
 	if (e.codec == wivrn::video_codec(-1))
 		throw std::runtime_error("invalid codec value " + item["codec"].get<std::string>());
+	SET_IF(use_10bit);
 	SET_IF(options);
 	SET_IF(device);
 	return e;
--- a/server/driver/configuration.cpp
+++ b/server/driver/configuration.cpp
@@ -135,7 +135,6 @@ configuration::encoder parse_encoder(const nlohmann::json & item)
 	SET_IF(codec);
 	if (e.codec == wivrn::video_codec(-1))
 		throw std::runtime_error("invalid codec value " + item["codec"].get<std::string>());
-	SET_IF(use_10bit);
 	SET_IF(options);
 	SET_IF(device);
 	return e;
@@ -188,6 +187,9 @@ configuration::configuration()
 
 		if (auto it = json.find("use-steamvr-lh"); it != json.end())
 			use_steamvr_lh = *it;
+
+		if (auto it = json.find("encode_10bit"); it != json.end())
+			encode_10bit = *it;
 
 		if (auto it = json.find("tcp-only"); it != json.end())
 			tcp_only = *it;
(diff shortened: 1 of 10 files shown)


server/driver/configuration.h
```diff
@@ -49,6 +49,7 @@ struct configuration
 		std::optional<double> offset_y;
 		std::optional<int> group;
 		std::optional<wivrn::video_codec> codec;
+		std::optional<bool> use_10bit;
```
xytovl: I have a small preference for `bit-depth` as an int instead of a boolean.

server/driver/configuration.cpp
```diff
@@ -146,6 +146,9 @@ configuration configuration::read_user_configuration()
 		if (auto it = json.find("use-steamvr-lh"); it != json.end())
 			result.use_steamvr_lh = *it;
 
+		if (auto it = json.find("encode_10bit"); it != json.end())
```
xytovl: This will have to be documented in docs/configuration.md

In JSON I'd like to be consistent and use dashes only: `encode-10bit` or `bit-depth`.

server/encoder/encoder_settings.cpp
```diff
@@ -141,7 +143,7 @@ static constexpr auto ffmpeg_version()
 	return result;
 }
 
-static std::optional<wivrn::video_codec> filter_codecs_vaapi(wivrn_vk_bundle & bundle, const std::vector<wivrn::video_codec> & codecs)
+static std::optional<wivrn::video_codec> filter_codecs_vaapi(wivrn_vk_bundle & bundle, const std::vector<wivrn::video_codec> & codecs, int bit_depth)
 {
 	video_encoder_ffmpeg::mute_logs mute;
 	encoder_settings s{
```
xytovl, line 149: bit_depth is unset in the structure, auto probing will fail. We will hit the throw in the constructor because it's neither 8 or 10.

galister: it's set to init to 8 in configuration.h? am i missing something?

xytovl: This is the code to detect supported codecs, used when encoders are not specified in the configuration file.

It will try to create vaapi decoders, if an exception is thrown it marks the codec as unsupported.

xytovl: Just add after initialisation
```c++
s.bit_depth = bit_depth;
```
Or you can shuffle around fields so that it can be written in the initialiser list. Because of the base class it is not possible to use designated initialisers here and there's fields in between we want to keep at the default value.

galister: sorry I wasn't around, will look at this soon. thanks so much!

PR #398, opened 2025-07-05
Forward session state to server

### LINKED ISSUE #395: Forward session state from client to server
https://github.com/WiVRn/WiVRn/pull/389#discussion_r2173908120

ImSapphire:
Closes #395

### DIFF
--- a/server/driver/wivrn_hmd.cpp
+++ b/server/driver/wivrn_hmd.cpp
@@ -76,7 +76,7 @@ wivrn_hmd::wivrn_hmd(wivrn::wivrn_session * cnx,
                 .supported = {
                         .orientation_tracking = true,
                         .position_tracking = true,
-                        .presence = info.presence,
+                        .presence = true,
                         .battery_status = true,
                 },
                 .update_inputs = [](xrt_device *) { return XRT_SUCCESS; },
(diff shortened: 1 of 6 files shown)


server/driver/wivrn_hmd.cpp
```diff
@@ -232,9 +232,24 @@ void wivrn_hmd::update_visibility_mask(const from_headset::visibility_mask_chang
 	m->at(mask.view_index) = mask.data;
 }
 
-void wivrn_hmd::update_presence(bool presence)
+bool wivrn_hmd::update_presence(bool presence, bool real)
 {
-	U_LOG_I("Updating user presence to %s", presence ? "true" : "false");
-	this->presence = presence;
+	if (real || presence == this->real_presence || !presence)
```
xytovl: The conditions are very hard to read, there's 4 booleans so plenty of possible states.

This should either be broken down into separate checks or have some comments about what it's doing.

`real`  means the source of the information is XR_EXT_user_presence on the headset
`presence` is the target value
`this->presence` is the computed value
`this->real_presence` is the last value received with source `XR_EXT_user_presence`

ImSapphire: Is the new version better?

xytovl: I think it's better

PR #400, opened 2025-07-06
Resolve symlinks before writing to configs

ImSapphire:


xytovl:
In which conditions can we hit the "server not found" error?
I have not seen any report of failures because of that

ImSapphire:
> In which conditions can we hit the "server not found" error? I have not seen any report of failures because of that

I just remembered people getting obscure problems when dashboard was used without server installed, but that was because the dep on wivrn-server was accidentally removed. I can drop that part if you'd prefer, it's not important

xytovl:
> > In which conditions can we hit the "server not found" error? I have not seen any report of failures because of that
> 
> I just remembered people getting obscure problems when dashboard was used without server installed, but that was because the dep on wivrn-server was accidentally removed. I can drop that part if you'd prefer, it's not important

I would drop that check

ImSapphire:
> > > In which conditions can we hit the "server not found" error? I have not seen any report of failures because of that
> > 
> > 
> > I just remembered people getting obscure problems when dashboard was used without server installed, but that was because the dep on wivrn-server was accidentally removed. I can drop that part if you'd prefer, it's not important
> 
> I would drop that check

Dropped and rebased

Issue #307, opened 2025-04-05
[Bug]: Default Encoder Settings Bury Midrange nVidia Cards


damaestro:
### Describe the bug

Starting v0.24 without a configuration saturates a mid-range card to the point where nothing is working well.

```
jon@davr:~$ nvidia-smi 
Sat Apr  5 16:03:14 2025       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.133.07             Driver Version: 570.133.07     CUDA Version: 12.8     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 3080 Ti     Off |   00000000:08:00.0  On |                  N/A |
|  0%   45C    P8             36W /  350W |     594MiB /  12288MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
```

When NMS is running, the card is at 100% across CPU/MEM/ENC.

HL:A doesn't struggle as much but the card is still pretty maxed.

### Steps to Reproduce

1. Start WiVRn
2. Configure NMS (no man's sky) with required settings in steam
3. Start game in VR mode

### Expected behavior

Defaults work well

### WiVRn Version

v0.24

### Server Config

Default detected (i.e. no configuration)

### Server Logs

```
WiVRn v0.24 starting
Service published: davr.damaestro.us
Client connected
Server started, PID 25727
 INFO [ipc_server_main]  'v24.0.0-373-gb2a502655' starting up...
DEBUG [comp_main_create_system_compositor] Doing init 0x564c5333e1c0
 WARN [compositor_check_and_prepare_xdev] Had to fill in meshuv on xdev 'WiVRn HMD', this should be done in the driver.
 INFO [create_device] Created device and COMPUTE queue with QUEUE_GLOBAL_PRIORITY_REALTIME.
 INFO [vk_print_device_info] Device info:
	name: NVIDIA GeForce RTX 3080 Ti
	vendor: 0x10de
	product: 0x2208
	deviceType: VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU
	apiVersion: 1.4.303
[...]

PR #414, opened 2025-07-27
Share CUDA context between all nvenc encoder instances

Kierek:
Hey, this PR fixes the huge VRAM usage when using split encoding with NVENC.

The old implementation was creating a whole new CUDA context for every single encoder instance, which was eating up an extra ~500 MiB of VRAM per slice. I've refactored this so that all NVENC encoders now share a single, global context. This makes the overhead for adding more encoders minimal and stops the unreasonable memory consumption.

The results are pretty dramatic. In my tests on an RTX 4080 Super (AV1 @ 130% on Quest 3 with a 150 Mbit/s bitrate), the VRAM usage dropped significantly:

- **Before (master):**
  - 1 encoder: **1590 MiB**
  - 3 encoders: **2652 MiB**
- **After (this PR):**
  - 1 encoder: **847 MiB**
  - 3 encoders: **916 MiB**

Besides the obvious memory savings, everything feels much smoother now and latency seems to be in a very good spot.

---
**A quick note:**
The fix works and the results are great, but I'd really appreciate some feedback on the implementation itself. My main language is Kotlin and to be honest, I find C++ a bit of a struggle. I'm not that familiar with common C++ practices, as a lot of it feels a bit unpleasant and unintuitive to me.

Any advice from a C++ expert on how this could be improved or written more idiomatically would be great. Thanks.

### DIFF
--- a/server/encoder/video_encoder_nvenc.h
+++ b/server/encoder/video_encoder_nvenc.h
@@ -20,6 +20,7 @@
 #pragma once
 
 #include "video_encoder.h"
+#include "video_encoder_nvenc_shared_state.h"
 #include <array>
 #include <ffnvcodec/dynlink_cuda.h>
 #include <ffnvcodec/dynlink_loader.h>
@@ -31,22 +32,13 @@ namespace wivrn
 
 class video_encoder_nvenc : public video_encoder
 {
-public:
-	struct deleter
-	{
-		void operator()(CudaFunctions * fn);
-		void operator()(NvencFunctions * fn);
-	};
-
 private:
 	wivrn_vk_bundle & vk;
 	// relevant part of the input image to encode
 	vk::Rect2D rect;
 
-	std::unique_ptr<CudaFunctions, deleter> cuda_fn;
-	std::unique_ptr<NvencFunctions, deleter> nvenc_fn;
-	NV_ENCODE_API_FUNCTION_LIST fn;
-	CUcontext cuda;
+	std::shared_ptr<video_encoder_nvenc_shared_state> shared_state;
+
 	void * session_handle = nullptr;
 	NV_ENC_OUTPUT_PTR bitstreamBuffer;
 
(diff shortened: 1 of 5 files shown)


server/encoder/video_encoder_nvenc.h
```diff
@@ -62,7 +54,7 @@ class video_encoder_nvenc : public video_encoder
 	int bitrate;
 
 public:
-	video_encoder_nvenc(wivrn_vk_bundle & vk, encoder_settings & settings, float fps, uint8_t stream_idx);
+	video_encoder_nvenc(std::shared_ptr<video_encoder_nvenc_shared_state> state, wivrn_vk_bundle & vk, encoder_settings & settings, float fps, uint8_t stream_idx);
```
xytovl: It is possible to hide the shared state object inside the constructor.

You can use the same pattern used in `video_encoder::sender`:
* Add a `static std::shared_ptr<video_encoder_nvenc_shared_state> get()` method in `video_encoder_nvenc_shared_state`. You can basically copy-paste the the method and just change the types.
* In the encoder constructor, assign the pointer from the get method.

server/encoder/video_encoder_nvenc_shared_state.cpp
```diff
@@ -0,0 +1,67 @@
...
+	fn.version = NV_ENCODE_API_FUNCTION_LIST_VER;
+	NVENC_CHECK_NOENCODER(nvenc_fn->NvEncodeAPICreateInstance(&fn));
+}
+
+video_encoder_nvenc_shared_state::~video_encoder_nvenc_shared_state()
+{
+	if (cuda)
+	{
+		cuda_fn->cuCtxDestroy(cuda);
+		cuda = nullptr;
```
xytovl: No need to set it to null, the member will cease to exist when the destructor finishes

server/encoder/video_encoder_nvenc_shared_state.h
xytovl, line 1: You should add the copyright header, including previous authors in case the code is copy/pasted from a different file.

xytovl:
That's great news if you fixed the multi-encoder issues on nvenc!

We will have to check if it is a better fix for https://github.com/WiVRn/WiVRn/issues/307 and enable split encoding by default on nvidia.

Kierek:
Thanks for the feedback! I've addressed all your comments and updated the branch.

PR #431, opened 2025-08-08
added TR translation

fuzunspm:


xytovl:
Have you tested the dashboard and the client app to validate it picks up the language, layout is correct etc?

I have triggered CI, https://github.com/WiVRn/WiVRn/actions/runs/16832625105?pr=431 has APK and flatpak for dashboard.

Issue #440, opened 2025-08-11
[Bug]: hotkey for new WiVRn with xrizer conflicts with UEVR


Patola:
### Describe the bug

The new hotkey (both thumbsticks at the same time) used by WiVRn 25.8 makes UEVR games unplayable, because it's the very same thumbstick this universal VR mod uses to call its menu. Without access to this, the games are unplayable.

### Steps to Reproduce

1 - try playing any of the games on this playlist (UEVR games that run on Linux under WiVRn -- the instructions are there for any individual game): https://www.youtube.com/watch?v=X_5FBRsRP-g&list=PLoRF2wMS15vV65fXuudiHHpYZlNMZpkQB

### WiVRn Version

25.8

### What GPU did the bug occur on?

NVIDIA

### What HMD did the bug occur on?

Quest

Patola:
Please add at least a way to deactivate the WiVRn shortcut, if not to change the combination.

xytovl:
9725ae7fc055ebd08a9bebfa98c61f4cb67dd23e adds a setting to disable the in-stream window.
This can be used until a mechanism for alternative binding can be found.

Patola:
Thanks, will test this solution.

Patola:
Ok, how to use the build for this setting? When I set my WiVRn branch to master and call for a clean build, it complains about the lack of a stubs.gen.cpp.

Patola:
This solution works. Thanks!

PR #460, opened 2025-08-17
Update pr_BR strings for small GUI activation option change

Patola:
Bug #440 has a new option added for deactivating the wivrn GUI, so a couple of new strings were added.
This patch adds pt_BR strings for these new strings.


Patola:
This fixes issue #446

PR #473, opened 2025-09-07
Update bug_report.yml to include plain render type

damaestro:
Added render type as plain for multiple fields in bug report template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -49,7 +49,6 @@ body:
         Provide the content of `config.json` for the server.
         Located in `${XDG_CONFIG_HOME-${HOME}/.config}/wivrn/config.json`
         or `~/.var/app/io.github.wivrn.wivrn/config/wivrn/config.json` for the flatpak version
-      render: plain
       placeholder: |
         {
           "encoders": [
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -49,6 +49,7 @@ body:
         Provide the content of `config.json` for the server.
         Located in `${XDG_CONFIG_HOME-${HOME}/.config}/wivrn/config.json`
         or `~/.var/app/io.github.wivrn.wivrn/config/wivrn/config.json` for the flatpak version
+      render: plain
```
xytovl: Does it still offer the option to attach a file?
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#attributes-for-textarea
> When this key is provided, the text area will not expand for file attachments or Markdown editing

damaestro: Yeah, it seems we'll need to remove it for things where you'd like to keep the upload/drag-n-drop.

I'm also not certain the render option is correct. I don't know the right lookup key in the [supported languages](https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml).

damaestro: The only one that bugs me is the vulkan summary. So let's start with setting just that one?

xytovl: Is it possible to force it to be an attachment? Or just remove it from the template, it is rarely useful.

damaestro: Let's see if changing the output render makes it better and if not we can just remove?

PR #485, opened 2025-09-15
Fix wrong refresh rate shown on fresh config

Knifa:
The config screen currently uses the highest reported refresh rate if no preferred refresh rate is set, which it will be with a fresh config.
https://github.com/WiVRn/WiVRn/blob/master/client/scenes/lobby_gui.cpp#L802

But it gets set to 0 by default here (so default for platform...?).
https://github.com/WiVRn/WiVRn/blob/master/client/scenes/lobby.cpp#L936 

On my Quest 3, this means the refresh rate defaults to 72Hz even though the menu shows 120Hz.

### DIFF
--- a/client/scenes/lobby_gui.cpp
+++ b/client/scenes/lobby_gui.cpp
@@ -799,7 +799,7 @@ void scenes::lobby::gui_settings()
 		const auto & refresh_rates = session.get_refresh_rates();
 		if (not refresh_rates.empty())
 		{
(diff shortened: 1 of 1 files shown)


xytovl:
The logic should be to default to the maximum supported rate, in https://github.com/WiVRn/WiVRn/blob/master/client/scenes/stream.cpp#L242 when the value is unset the default frame rate is the last value (highest).

The refresh rate in the lobby may be different.

PR #514, opened 2025-10-01
fix index controller mapping

galister:
allows future mapping of controllers that do not have symmetric bindings

### DIFF
--- a/server/driver/wivrn_controller.cpp
+++ b/server/driver/wivrn_controller.cpp
@@ -260,7 +260,29 @@ xrt_binding_output_pair simple_output_binding[] = {
         {XRT_OUTPUT_NAME_SIMPLE_VIBRATION, XRT_OUTPUT_NAME_TOUCH_HAPTIC},
 };
 
-xrt_binding_input_pair index_input_binding[] = {
+xrt_binding_input_pair index_input_binding_left[] = {
+        {XRT_INPUT_INDEX_SYSTEM_CLICK, XRT_INPUT_TOUCH_MENU_CLICK},
+        {XRT_INPUT_INDEX_SYSTEM_TOUCH, XRT_INPUT_INDEX_SYSTEM_TOUCH},
+        {XRT_INPUT_INDEX_A_CLICK, XRT_INPUT_TOUCH_X_CLICK},
+        {XRT_INPUT_INDEX_A_TOUCH, XRT_INPUT_TOUCH_X_TOUCH},
+        {XRT_INPUT_INDEX_B_CLICK, XRT_INPUT_TOUCH_Y_CLICK},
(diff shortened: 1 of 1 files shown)


xytovl:
~~Can't we just map both a/b and x/y? This is done for other controllers.~~
edit: the left part of the binding must be unique, we would have to use index a/b for all controllers, and not use per-hand button names. This should work:

* Replace `SET_INPUT(TOUCH, X_CLICK);` with `SET_INPUT(INDEX, A_CLICK);` and the other similar declaration.
* As `WIVRN_CONTROLLER_X_CLICK` and `WIVRN_CONTROLLER_A_CLICK` are already aliases, nothing needs to be done there
* Replace all `XRT_INPUT_TOUCH_A_CLICK` and simlar to the INDEX variant.

galister:
Not exactly sure I understand.

There's also how index has `system` on both hands while touch has `menu` on left and `system` on right.

xytovl:
Actually my idea wouldn't work, the TOUCH interaction profile must exactly match the input names, so we must retain different buttons on left and right controller.

What bothers me in the pull request is the special case for index, where we need `index_input_binding` and `index_input_binding_count` variables.

server/driver/wivrn_controller.cpp
```diff
@@ -430,57 +452,70 @@ xrt_binding_output_pair pico4_output_binding[] = {
...
-};
+template <int HandId>
+constexpr auto
+make_binding_profiles()
+{
+	constexpr auto & index_input_binding =
+	        HandId == 0 ? index_input_binding_left : index_input_binding_right;
+	constexpr size_t index_input_binding_count =
+	        HandId == 0 ? std::size(index_input_binding_left)
+	                    : std::size(index_input_binding_right);
```
xytovl: This variable is not used

galister:
I would imagine that it's not just the index but there are probably other controllers where 1-2 buttons don't work. We just don't know yet.

The new Valve headset as well for example will have a D-pad on the left and 4 buttons on the right.

My goal with this is to make it easier to do these kinds of mappings in the future. I'm not a C++ wizard however so feel free to give me pointers.

As for not introducing variables, I think we can do ternary instead, but I wanted to avoid that for readability.

xytovl:
> I would imagine that it's not just the index but there are probably other controllers where 1-2 buttons don't work. We just don't know yet.
> 
> The new Valve headset as well for example will have a D-pad on the left and 4 buttons on the right.
> 
> My goal with this is to make it easier to do these kinds of mappings in the future. I'm not a C++ wizard however so feel free to give me pointers.
> 
> As for not introducing variables, I think we can do ternary instead, but I wanted to avoid that for readability.

As `index_input_binding_count` is not actually used, there's only one new variable, which is fine. Just drop the unused variable and it's good for merge

PR #517, opened 2025-10-05
Add the config path for flatpack to the documentation

Krzeszny:


docs/configuration.md
```diff
@@ -4,6 +4,7 @@ Configuration is done on server side.
 Files are read from
 - `/usr/share/wivrn/config.json` (where `/usr` is selected at configure time with `CMAKE_INSTALL_PREFIX`)
 - `/etc/wivrn/config.json`
+- `$HOME/.var/app/io.github.wivrn.wivrn/config/wivrn/config.json` if you're using flatpack
 - `$XDG_CONFIG_HOME/wivrn/config.json` or if `$XDG_CONFIG_HOME` is not set, `$HOME/.config/wivrn/config.json`.
 
 Files later in the list replace top-level values from previous ones.
```
xytovl: For flatpak, it's that file only, not an additional element in the list.

Krzeszny: What element?

xytovl: Normally, it read configuration in /usr, /etc and home, merging the results.

When running in a flatpak, the only one that will be read is the file in ~/.var/...

The way you added it, one could understand that it reads content from `/usr/share/wivrn/config.json`, `/etc/wivrn/config.json`, `$HOME/.var/app/io.github.wivrn.wivrn/config/wivrn/config.json` (only for flatpak), `$XDG_CONFIG_HOME/wivrn/config.json`.

Krzeszny: Ah, okay. How about now?

xytovl: That looks ok

PR #521, opened 2025-10-06
More documentation improvements

Krzeszny:
I forgot to make the previous PR a draft. 
Added more troubleshooting steps and made tiny changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,9 +128,7 @@ If the server list is empty in the headset app:
 - You can now continue the pairing process as documented in the running section.
 
 ## How do I see server logs when using the dashboard?
-
-```
-journalctl -f --no-hostname -u io.github.wivrn.wivrn.desktop
+Click Troubleshoot > Open server logs or navigate to `${XDG_STATE_HOME}/wivrn/wivrn-dashboard` (with fallback to `${HOME}/.local/state` for `${XDG_STATE_HOME}`, and for flatpak `${HOME}/.var/app/io.github.wivrn.wivrn/.local/state/wivrn/wivrn-dashboard`.
 ```
 
 # Contributing
--- a/README.md
+++ b/README.md
@@ -131,6 +131,18 @@ If the server list is empty in the headset app:
 
 Click Troubleshoot > Open server logs, or navigate to `${XDG_STATE_HOME}/wivrn/wivrn-dashboard` (with fallback to `${HOME}/.local/state` for `${XDG_STATE_HOME}`, or for flatpak `${HOME}/.var/app/io.github.wivrn.wivrn/.local/state/wivrn/wivrn-dashboard`.
 
+## I hear crackling audio or corrupted video signal with an Nvidia GPU
+
+- Switch to NVENC encoding in Settings, as Vulkan and VAAPI might (will) not work on Nvidia GPUs.
+- Try using h.265 instead of h.264 encoding.
+
+## I can see high motion latency. There's a black border following the image when I turn my head.
+
+- Try different encoders in the dashboard
+- If you're using manual foveation, make sure it's not set to 0% (the default is 50%)
+- Decrease the resolution in the WiVRn app
+- Use a different GPU. Nvidia GPUs have higher encoding latency on WiVRn (even three times worse when using resolutions much higher than default).
+
 # Contributing
 
 ## Translations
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -129,7 +129,21 @@ If the server list is empty in the headset app:
...
+
+## I hear crackling audio or see corrupted video signal with an Nvidia GPU
+
+- Switch to NVENC or x264 encoding in Settings, as Vulkan and VAAPI might (will) not work on Nvidia GPUs
+- Try using x264 H.264 instead of NVENC H.264.
+
+## I can see high motion latency. There's a black border following the image when I turn my head
+
+- Try different encoders in the dashboard
+- If you're using manual foveation, make sure it's **not** set to 0% (the default is 50%)
```
xytovl: This was recently changed, clicking "manual" now sets the slider to 50%, the user then has to select a value.

My general recommendation would be to set everything to default values, with the "reset" button.
We may suggest software encoding to rule out hardware acceleration issues.

README.md
```diff
@@ -129,7 +129,21 @@ If the server list is empty in the headset app:
...
 
-Click Troubleshoot > Open server logs, or navigate to `${XDG_STATE_HOME}/wivrn/wivrn-dashboard` (with fallback to `${HOME}/.local/state` for `${XDG_STATE_HOME}`, or for flatpak `${HOME}/.var/app/io.github.wivrn.wivrn/.local/state/wivrn/wivrn-dashboard`.
+- Click Troubleshoot > Open server logs, or
+- Navigate to `${XDG_STATE_HOME}/wivrn/wivrn-dashboard` (with fallback to `${HOME}/.local/state` for `${XDG_STATE_HOME}`, or
+- For flatpak, navigate to `${HOME}/.var/app/io.github.wivrn.wivrn/.local/state/wivrn/wivrn-dashboard`.
+
+## I hear crackling audio or see corrupted video signal with an Nvidia GPU
+
+- Switch to NVENC or x264 encoding in Settings, as Vulkan and VAAPI might (will) not work on Nvidia GPUs
+- Try using x264 H.264 instead of NVENC H.264.
```
xytovl: I don't think this is generic enough, per user reports this happens for varied reasons:
* bitrate set way too high (this is the reason for the 200Mb/s cap)
* bad network
* headset plugged in by USB, but using normal network because not started with the specific USB button
* overloaded decoder, due to too high encoded stream

We may replace the paragraph with a note about NVIDIA support, stating that due to lack of developer with such hardware it may not be properly optimized and software encoding or vulkan video encode may offer better performance.

README.md
```diff
@@ -129,7 +129,21 @@ If the server list is empty in the headset app:
...
+
+## I can see high motion latency. There's a black border following the image when I turn my head
+
+- Try different encoders in the dashboard
+- If you're using manual foveation, make sure it's **not** set to 0% (the default is 50%)
+- Decrease the resolution in the WiVRn app
+- Use a different GPU. Nvidia GPUs have higher encoding latency on WiVRn (even three times worse when using resolutions much higher than default).
```
xytovl: NVIDIA stylizes their name as uppercase.

README.md
Krzeszny: Done

README.md
Krzeszny: fixed

README.md
Krzeszny: I made it more generic. I'll add something about lowering the bitrate

Krzeszny: How about now?

xytovl: The crackling audio section is now weird, with a single item, about software encoding.
I would drop that section as we don't have some relevant advice. We should merge it with the following section which is for generic performance/network issues. The "Try using different encoders" should mention specifically software encoding as a method to check if hardware encoding is not working well.

README.md
```diff
@@ -129,7 +129,24 @@ If the server list is empty in the headset app:
...
+## I have high motion latency. There's a black border following the image when I turn my head
+
+- When connecting through USB, make sure the headset isn't connected through WiFi (switch off WiFi)
+- Reset the settings using the button at the bottom of the settings tab
+- Try using different encoders
+- Decrease the resolution in the WiVRn app
+- Decrease the bitrate
+- Connect through USB or use a better WiFi router.
+
+Note: WiVRn isn't properly optimized for NVIDIA GPUs due to the lack of developers with NVIDIA hardware. Motion latency can be two or three times worse when rendering resolutions higher than default. NVIDIA GPU users must choose between low latency and high resolution.
```
xytovl: > NVIDIA GPU users must choose between low latency and high resolution.

This is not true for everyone, some are happy with current performance.

Krzeszny: Improved

xytovl:
I think it's generally good now, I can merge if you don't plan on adding anything

Krzeszny:
> I think it's generally good now, I can merge if you don't plan on adding anything

Okay

PR #523, opened 2025-10-07
update monado

galister:
Primarily in order to pull in https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2588

Please verify that the `lock` calls make sense.

Tested for ~5m.


xytovl:
The lock is correct, however, to update Monado I usually split the bump on each interface break to make bisections possible:

* create a separate Monado clone
* in that clone, `git bisect start`, `git bisect good $(cat "~${WIVRN_DIR}/monado-rev")` `git bisect bad origin/main`
* create a bisect.sh script
```bash
#!/bin/bash

git rev-parse HEAD > ${WIVRN_DIR}/monado-rev
exec ninja -C ${WIVRN_BUILD_DIR}
```
* run `git bisect run ./bisect.sh`

This will find the first Monado revision that breaks build.

You can then create a commit with that revision, push it for CI to check the nix stuff, fix the hash in flake.nix, push --force and then continue with the next interface break.

galister:
The commit in monado-rev is the first commit that broke the build, so I think we're good there. Also updated the nix flake.

Issue #275, opened 2025-02-23
[Feature Request]: HEVC and AV1 10bit encoding


Colin1224:
### Feature description

Add 10bit depth encoding for HEVC and AV1.

### Reason for the feature

Enabling 10bit encoding will result in better video compression and a slight saving in bitrate.

When using Virtual Desktop on Windows, I always make sure to use the 10bit encoding as the visual difference is very clear, 10bit encoding results in significantly less banding. I was looking through the code and as far as I could tell WiVRn only uses 8bit. The encoding quality in dark scenes seems to reflect this.

8bit depth only has a color range of 256 values, so on the darker end the gap between values is large, causing banding. 10bit depth has 1024 values to work with, so the gap between darker values isn't nearly as large. These benefits happen even when encoding from an 8bit source

In addition, 10bit encoding actually results in a slightly [lower bitrate](https://yukisubs.wordpress.com/wp-content/uploads/2016/10/why_does_10bit_save_bandwidth_-_ateme.pdf).

### Proposed implementation

WiVRn already uses FFMPEG which supports the necessary pixel formats with NVENC codecs and NVIDIA Video Codec SDK v12.2 which already [supports](https://docs.nvidia.com/video-technologies/video-codec-sdk/12.2/nvenc-application-note/index.html#nvenc-capabilities) 10bit encoding for HEVC and AV1.

`ffmpeg -h encoder=hevc_nvenc`

```
Encoder hevc_nvenc [NVIDIA NVENC hevc encoder]:
    General capabilities: dr1 delay hardware 
    Threading capabilities: none
    Supported hardware devices: cuda cuda 
    Supported pixel formats: yuv420p nv12 p010le yuv444p p016le yuv444p16le bgr0 bgra rgb0 rgba x2rgb10le x2bgr10le gbrp gbrp16le cuda
```

`ffmpeg -h encoder=av1_nvenc`

```
Encoder av1_nvenc [NVIDIA NVENC av1 encoder]:
    General capabilities: dr1 delay hardware 
    Threading capabilities: none
    Supported hardware devices: cuda cuda 
    Supported pixel formats: yuv420p nv12 p010le yuv444p p016le yuv444p16le bgr0 bgra rgb0 rgba x2rgb10le x2bgr10le gbrp gbrp16le cuda
```

Both support `p010le` which I believe is the pixel format that's needed.

### Additional context

[...]

PR #532, opened 2025-10-14
10-bit encoding on NVENC

aomiki:
An attempt at implementing #275 for nvenc.
So this is what I tried and, most notably, HEVC has distorted colors that are all blue.
Followed the docs at [nvenc site](https://docs.nvidia.com/video-technologies/video-codec-sdk/12.2/nvenc-video-encoder-api-prog-guide/index.html#high-bit-depth-encoding).

<details><summary>what yuview shows, looks the same in headset</summary>
<p>

image

</p>
</details> 

I think I missed some configuration, but ran out of ideas at this point, would appreciate help with which direction to take.

I thought maybe changing buffer format from NV_ENC_BUFFER_FORMAT_NV12 to something else would help (there are formats that specifically say that they are 10 bit), but got either the same result or invalid argument error. 
Also maybe buffer_size should be made bigger? But not sure how to calculate new size.

If AV1 is selected, nvenc fails to get caps with "EncodeAPI Internal Error" and status 8 (invalid arguments). Which seems to be unrelated to 10 bit encoding.

### DIFF
--- a/server/encoder/video_encoder_nvenc.cpp
+++ b/server/encoder/video_encoder_nvenc.cpp
@@ -88,7 +88,7 @@ static auto encode_guid(video_codec codec)
 	throw std::out_of_range("Invalid codec " + std::to_string(codec));
 }
 
-static void check_encode_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encodeGUID)
+static void check_encode_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encode_guid)
 {
 	uint32_t count;
 	NVENC_CHECK(shared_state->fn.nvEncGetEncodeGUIDCount(session_handle, &count));
@@ -96,21 +96,21 @@ static void check_encode_guid_supported(std::shared_ptr<video_encoder_nvenc_shar
 	std::vector<GUID> encodeGUIDs(count);
 	NVENC_CHECK(shared_state->fn.nvEncGetEncodeGUIDs(session_handle, encodeGUIDs.data(), count, &count));
 
-	if (!std::ranges::contains(encodeGUIDs, encodeGUID))
+	if (!std::ranges::contains(encodeGUIDs, encode_guid))
 	{
 		throw std::runtime_error("nvenc: GPU doesn't support selected codec.");
 	}
 }
 
-static void check_preset_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encodeGUID, GUID presetGUID)
+static void check_preset_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encode_guid, GUID preset_guid)
 {
 	uint32_t count;
-	NVENC_CHECK(shared_state->fn.nvEncGetEncodePresetCount(session_handle, encodeGUID, &count));
+	NVENC_CHECK(shared_state->fn.nvEncGetEncodePresetCount(session_handle, encode_guid, &count));
 
 	std::vector<GUID> presetGUIDs(count);
-	NVENC_CHECK(shared_state->fn.nvEncGetEncodePresetGUIDs(session_handle, encodeGUID, presetGUIDs.data(), count, &count));
+	NVENC_CHECK(shared_state->fn.nvEncGetEncodePresetGUIDs(session_handle, encode_guid, presetGUIDs.data(), count, &count));
 
-	if (!std::ranges::contains(presetGUIDs, presetGUID))
+	if (!std::ranges::contains(presetGUIDs, preset_guid))
 	{
 		throw std::runtime_error("nvenc: Internal error. GPU doesn't support selected encoder preset.");
 	}


xytovl:
The `NV_ENC_BUFFER_FORMAT_NV12` needs to be changed indeed, `NV_ENC_REGISTER_RESOURCE` says `pitch` is in bytes, si it has to be doubled for 10bit.
In the `copyImageToBuffer`, the offset also needs to be doubled. The other parameters are in texels and should be fine.

While we're changing this, the 2 `copyImageToBuffer` can be combined into one by replacing the `vk::BufferImageCopy` with

```cpp
std::array{
  vk::BufferImageCopy{/*parameters*/},
  vk::BufferImageCopy{/*parameters*/},
}
```

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -147,24 +152,55 @@ video_encoder_nvenc::video_encoder_nvenc(
...
+
 			break;
 		case video_codec::h265:
+			params.profileGUID = NV_ENC_HEVC_PROFILE_MAIN10_GUID;
 			params.encodeCodecConfig.hevcConfig.repeatSPSPPS = 1;
 			params.encodeCodecConfig.hevcConfig.maxNumRefFramesInDPB = 0;
 			params.encodeCodecConfig.hevcConfig.idrPeriod = NVENC_INFINITE_GOPLENGTH;
 			params.encodeCodecConfig.hevcConfig.hevcVUIParameters.videoFullRangeFlag = 1;
+			params.encodeCodecConfig.hevcConfig.inputBitDepth = NV_ENC_BIT_DEPTH_8;
+			params.encodeCodecConfig.hevcConfig.outputBitDepth = bitDepth;
```
xytovl: Both should be set to `bitDepth`

aomiki: I thought that input refers to the game window that we are streaming, so for 10 bit the game has to be in hdr mode? Is it something else?

xytovl: The game renders in whatever bit depth it wants, when 10 bit encoding is enabled the RGB -> YCbCr conversion performed by WiVRn writes to a 10 bit texture, which is then used as input to the encoder.

aomiki: > writes to a 10 bit texture, which is then used as input to the encoder.

Ahh the context I've been missing, thank you!

xytovl: https://github.com/WiVRn/WiVRn/blob/master/server/driver/wivrn_comp_target.cpp#L417-L420 is where the image format is selected.

https://github.com/WiVRn/WiVRn/blob/master/server/driver/wivrn_comp_target.cpp#L187-L192 is where the individual plane formats are defined, for Y and CbCr.

aomiki:
This one is actually not crashing and has normal colors for the first time!
(it crashed with only pitch & offset increase, without buffer_size increase and profile change)

Not sure how to check if it's actually 10 bit though.

xytovl:
You can dump the raw stream and confirm with YUView, but if it's showing the right image with the 10 bit intermediate data, it's very likely to be correct.

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -279,43 +315,38 @@ std::pair<bool, vk::Semaphore> video_encoder_nvenc::present_image(vk::Image y_cb
...
+	                        .imageOffset = {
+	                                .x = rect.offset.x,
+	                                .y = rect.offset.y,
+	                        },
+	                        .imageExtent = {
+	                                .width = rect.extent.width,
+	                                .height = rect.extent.height,
+	                                .depth = 1,
+	                        }},
+	                vk::BufferImageCopy{.bufferOffset = rect.extent.width * rect.extent.height * 2, .bufferRowLength = uint32_t(rect.extent.width / 2), .imageSubresource = {
```
xytovl: The 2 is for 10 bit only.

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -122,6 +122,11 @@ video_encoder_nvenc::video_encoder_nvenc(
 	presets.resize(count);
 	NVENC_CHECK(shared_state->fn.nvEncGetEncodePresetGUIDs(session_handle, encodeGUID, presets.data(), count, &count));
 
+	std::vector<GUID> profiles;
+	NVENC_CHECK(shared_state->fn.nvEncGetEncodeProfileGUIDCount(session_handle, encodeGUID, &count));
+	profiles.resize(count);
+	NVENC_CHECK(shared_state->fn.nvEncGetEncodeProfileGUIDs(session_handle, encodeGUID, profiles.data(), count, &count));
+
```
xytovl: This seems unused

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -147,24 +152,55 @@ video_encoder_nvenc::video_encoder_nvenc(
...
 			params.encodeCodecConfig.h264Config.maxNumRefFrames = 0;
 			params.encodeCodecConfig.h264Config.idrPeriod = NVENC_INFINITE_GOPLENGTH;
 			params.encodeCodecConfig.h264Config.h264VUIParameters.videoFullRangeFlag = 1;
+
+			if (settings.bit_depth != 8)
+				throw std::runtime_error("selected codec only supports 8-bit encoding");
+
 			break;
 		case video_codec::h265:
+			params.profileGUID = NV_ENC_HEVC_PROFILE_MAIN10_GUID;
```
xytovl: Should be for 10 bit only

aomiki:
> You can dump the raw stream and confirm with YUView

Yes that's what I did, but couldn't find where to look for it in YUView.

<details><summary>The image saved from yuview of the new stream</summary>
<p>

supposedly-10bit-valve-logo

</p>
</details>

server/encoder/video_encoder_nvenc.cpp
aomiki: Same for presets (right above this block of code). I think both presets and profiles are meant for checking against them the availability of the desired profile/preset.

server/encoder/video_encoder_nvenc.cpp
aomiki: Was about to ask, how do I get the selected bit depth in this function?

xytovl: You can add a new member, `bytes_per_pixel` for instance, you can use it in other places where the ternary operator is used.

aomiki:
I think I figured out why AV1 isn't working, [encoding AV1 is supported starting from 4000 series GPUs](https://developer.nvidia.com/video-encode-decode-support-matrix). And I have 3000 series Ampere GPU.  

Added function to check if codec is supported. (GUIDs don't have equality operator defined, so had to compare field by field)
<details><summary>And get the message when AV1 is selected.</summary>
<p>

image

</p>
</details>

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -80,6 +80,30 @@ static auto encode_guid(video_codec codec)
...
+	{
+		if (guid.Data1 == encodeGUID.Data1 &&
+		    guid.Data2 == encodeGUID.Data2 &&
+		    guid.Data3 == encodeGUID.Data3)
+		{
+			if (std::equal(std::begin(guid.Data4), std::end(guid.Data4), std::begin(encodeGUID.Data4)))
+			{
+				return;
+			}
+		}
```
xytovl: Can be extracted to
```cpp
bool operator==(const GUID& l, const GUID&r)
{
  return l.Data1 == r.Data1 and
    l.Data2 == r.Data2 and
    l.Data3 == r.Data3 and
    std::ranges::equal(l.Data4, r.Data4);
}
```

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -80,6 +80,30 @@ static auto encode_guid(video_codec codec)
...
 
+static void check_encode_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encodeGUID)
+{
+	uint32_t count;
+	std::vector<GUID> encodeGUIDs;
+	NVENC_CHECK(shared_state->fn.nvEncGetEncodeGUIDCount(session_handle, &count));
+	encodeGUIDs.resize(count);
+	NVENC_CHECK(shared_state->fn.nvEncGetEncodeGUIDs(session_handle, encodeGUIDs.data(), count, &count));
+
+	for (GUID guid: encodeGUIDs)
```
xytovl: Once operator== is added, can be replaced with `std::range::contains(encodeGUIDs, encodeGUID)`

aomiki:
While I'm at it, is it okay if I update nvenc sdk to 13.0 and also add H.264 10 bit encoding?

It was [released with sdk 13](https://developer.nvidia.com/blog/nvidia-video-codec-sdk-13-0-powered-by-nvidia-blackwell/) for the latest 5000 series GPUs. It would be just adding a few lines of code [in the same pattern as with other codecs](https://docs.nvidia.com/video-technologies/video-codec-sdk/13.0/nvenc-video-encoder-api-prog-guide/index.html#high-bit-depth-encoding).

xytovl:
Updating the sdk version also increases the minimum driver version, per https://github.com/FFmpeg/nv-codec-headers this would bump the requirement to 570.0. I don't recommend it as some distros may be on older versions.

10bit H.264 has to be supported by decoders as well, I have not tested if it is.

aomiki:
> Updating the sdk version also increases the minimum driver version, per https://github.com/FFmpeg/nv-codec-headers this would bump the requirement to 570.0. I don't recommend it as some distros may be on older versions.

I see, didn't know about the requirement. Judging by the fact that only recently it was update to 580 on arch, I guess it can take some time for others to catch up.

xytovl:
We should first complete the current pull request, then make a separate one for nvenc update.
As we don't have any telemetry I can't say if there's people with ancient NVIDIA drivers, it looks like Ubuntu backports them and driver 580 is available on the 22.04 LTS.
We don't have a defined list of supported distribution, but latest Ubuntu LTS, Debian testing are usually what is needed.

Regarding the current pull request, if it's working correctly, you can mark it as ready for review and I'll do a last round of review/comments. If you need more time, I may merge https://github.com/WiVRn/WiVRn/tree/feat/reliable-reconnect in the meantime that changes the encoder interface a little, so you would have to rebase.

aomiki:
> We should first complete the current pull request, then make a separate one for nvenc update.

That would be cleaner, makes sense. Should we still wait or just check if ubuntu and debian has the 570 driver and open it after this one?

> Ubuntu LTS, Debian testing are usually what is needed.

Yeah I tried checking which driver version Ubuntu has, but it seems more complicated to do that than on arch. 
Thought that it may be easier for me to just run it in VM and execute [all the commands for checking version.](https://documentation.ubuntu.com/server/how-to/graphics/install-nvidia-drivers/)

> If you need more time, I may merge https://github.com/WiVRn/WiVRn/tree/feat/reliable-reconnect in the meantime that changes the encoder interface a little, so you would have to rebase.

Probably need more time, everything up to now was done in a hurry, so I want to sit with it more. Why rebase in this case? Why not just merge with it? (Just curious, but no problem doing either one)

xytovl:
> > We should first complete the current pull request, then make a separate one for nvenc update.
> 
> That would be cleaner, makes sense. Should we still wait or just check if ubuntu and debian has the 570 driver and open it after this one?
> 
> > Ubuntu LTS, Debian testing are usually what is needed.
> 
> Yeah I tried checking which driver version Ubuntu has, but it seems more complicated to do that than on arch. Thought that it may be easier for me to just run it in VM and execute [all the commands for checking version.](https://documentation.ubuntu.com/server/how-to/graphics/install-nvidia-drivers/)
> 

If both have the driver, there's no need to wait, especially if there's a benefit in updating the SDK.

> > If you need more time, I may merge https://github.com/WiVRn/WiVRn/tree/feat/reliable-reconnect in the meantime that changes the encoder interface a little, so you would have to rebase.
> 
> Probably need more time, everything up to now was done in a hurry, so I want to sit with it more. Why rebase in this case? Why not just merge with it? (Just curious, but no problem doing either one)

I prefer rebase as history is easier to read, it's not a hard requirement.
As for the time, if you have tests in mind or further improvements to do, take your time. If you think it's good enough, it can be merged and in case of regressions be fixed later. Nobody will test your development branch or give feedback on it, it's already quite hard to get feedback from master before a release is made.

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -80,28 +88,46 @@ static auto encode_guid(video_codec codec)
...
-			if (std::equal(std::begin(guid.Data4), std::end(guid.Data4), std::begin(encodeGUID.Data4)))
-			{
-				return;
-			}
-		}
+		throw std::runtime_error("nvenc: GPU doesn't support selected codec.");
+	}
+}
+
+static void assert_preset_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encodeGUID, GUID presetGUID)
```
xytovl: I would prefer `check_preset_guid_supported`, as we expect assertion to be the `assert` macro, which depends on compile type.

aomiki:
Okay I feel like it's okay to merge.

Was curious as to how expensive it is to query guids in those `is_*_guid_supported` functions. Profiled a game session with nvidia-nsight & nvtx, turns out average execution time is just a few microseconds.
_(encode function is Avg and Median 3.5 ms and Max 17 ms)_

Also you mentioned how to improve error handling, but I struggle to understand how to use `std::system_error` and `NVENC_CHECK` macro together, but regardless I can do it in another PR.

PR #543, opened 2025-10-20
Updated 1 message

Patola:
Updated a new pt_BR message for WiVRn. I only noticed one single translation that changed slightly, is that it? Closes #446


xytovl:
Yes, that's all that changed: there was a copy/paste between the avahi-only and avahi+wivrn firweall message

by the way, the issue will get closed automatically even without the "Closes" mention, CI checks if all translations are done and closes/reopens automatically

Issue #303, opened 2025-03-27
[Bug]: Vulkan video encoding breaks down after ~12hrs of continuous stream


Beyley:
### Describe the bug

Started testing vulkan video again, now that Mesa 25.0 is out on arch, and after approximately 12 hours of streaming (suspiciously started happening at the 12 hour mark today), the stream slowly starts turning white with compression artifacts, before disconnecting from the headset altogether, and refusing to let me re-connect.

### Steps to Reproduce

1. Start a WiVRn stream
2. Leave it running on some static scene
3. Wait many hours, eventually the static scene will start filling up with white compression artifacts before the headset disconnects altogether with "video stream interrupted"

### Expected behavior

Stream continues without issue indefinitely, like it does on VAAPI and x264

### WiVRn Version

f03cb86c19000fb28dd4c000d76642534d5eb53d

### Server Config

{
  "scale": [
    0.4,
    0.4
  ],
  "bitrate": 100000000,
  "tcp_only": true,
  "encoders.disabled": [
    {
      "height": 1.0,
      "offset_x": 0.0,
      "offset_y": 0.0,
      "width": 1.0
    }
  ]
}

### Server Logs

Entire scrollback is dominated by `WARN [log_frame_time_diff] Frame late by 11.11ms!`, likely because I'm running Resonite.

### What GPU did the bug occur on?

AMD

### Vulkan Info

[vulkaninfo.txt](https://github.com/user-attachments/files/19493268/vulkaninfo.txt)

### What HMD did the bug occur on?

Quest

### ADB Logs

Headset (Quest 2) has since run out of battery, so I can't fill this in right now, as logs are lost. Next time I hit this problem, I'll try to capture logs

xytovl:
https://github.com/WiVRn/WiVRn/commit/6b91848d29a478261f1e2508d068aa3cadd4e514 fixes a command buffer not being reset in vulkan video encode.

It might fix this issue, but I can't really test it...

xytovl:
closing as there's no reasonable way to test.

PR #565, opened 2025-10-30
vulkan h265 /w 10-bit + fix encoder hang

galister:
Vulkan H265 encoding works properly now.

Tested Vulkan encoders (RADV):
- h264　encode ✅　bitrate change ✅
- h265 8-bit　encode ✅　bitrate change ✅
- h265 10-bit　encode ✅　bitrate change ✅

With 8-bit h265, ffmpeg gives a warning about deprecated format. This **also happens with VAAPI**, and the picture in the headset looks fine.
```
[swscaler @ 0x7f2041193e80] deprecated pixel format used, make sure you did set range correctly
```

### DIFF
--- a/dashboard/settings.cpp
+++ b/dashboard/settings.cpp
@@ -468,12 +468,15 @@ QList<Settings::video_codec> Settings::allowedCodecs() const
 			        video_codec::H265,
 			        video_codec::Av1,
 			};
-		case encoder_name::X264:
 		case encoder_name::Vulkan:
 			return {
 			        video_codec::CodecAuto,
 			        video_codec::H264,
 			        video_codec::H265,
+			};
+		case encoder_name::X264:
+			return {
+			        video_codec::H264,
 			};
 	}
 	return {video_codec::CodecAuto};


server/encoder/video_encoder_vulkan.cpp
```diff
@@ -745,8 +745,12 @@ std::pair<bool, vk::Semaphore> wivrn::video_encoder_vulkan::present_image(vk::Im
 		                .flags = vk::VideoCodingControlFlagBitsKHR::eEncodeRateControl,
 		        });
 
+	// avoid self-reference when the DPB ring wraps (cannot reference the slot we're writing).
+	auto curr_slot_idx = slot->info.slotIndex;
+	auto use_ref = (ref_slot && ref_slot->info.slotIndex == curr_slot_idx);
+
```
galister: this part avoids a case where the encoder will get stuck not producing any frames until an idr is requested.

galister: possible fix for https://github.com/WiVRn/WiVRn/issues/303 but in my case i was just shown the "launch an app" panel, no artifacts.

galister:
~~hold on with this, just noticed a pretty big flaw~~

galister:
removed wip as it is working now. let's continue with the review.

server/encoder/video_encoder_vulkan_h265.cpp
```diff
@@ -17,15 +17,90 @@
...
+}
+
+static uint32_t find_msb(uint32_t v)
+{
+	for (int i = 31; i >= 0; i--)
+		if (v & (1u << i))
+			return i;
+
+	return UINT32_MAX;
+}
```
xytovl, line 60: Those look like ` countl_one` and `countr_one` from https://en.cppreference.com/w/cpp/utility/bit.html

server/encoder/video_encoder_vulkan.cpp
```diff
@@ -667,15 +675,40 @@ std::pair<bool, vk::Semaphore> wivrn::video_encoder_vulkan::present_image(vk::Im
 	{
 		slot_item.idr = false;
 	}
+
+	// avoid self-reference when the DPB ring wraps (cannot reference the slot we're writing).
+	bool use_ref = (ref_slot && ref_slot->info.slotIndex != static_cast<int32_t>(slot_index));
+
```
xytovl: It looks like we could remove the bool, use `ref_slot`  (implicitly `!= nullptr`) instead by moving the `slotIndex` check in the `if (i.frame_index == last_ack and i.info.slotIndex != -1)` condition.
The condition becomes: reference slot is valid, not selected for overwrite and acknowledged by client.

server/encoder/video_encoder_vulkan_h265.cpp
galister: throwing exception now

server/encoder/video_encoder_vulkan.cpp
galister: adjusted as per request

server/encoder/video_encoder_vulkan_h265.cpp
```diff
@@ -418,6 +418,13 @@ void * wivrn::video_encoder_vulkan_h265::encode_info_next(uint32_t frame_num, si
 
 		reference_lists_info.num_ref_idx_l0_active_minus1 = 0;
 		reference_lists_info.RefPicList0[0] = static_cast<uint8_t>(*ref_slot);
+
+		if (st_rps.used_by_curr_pic_s0_flag == 0)
+		{
+			// ref_slot is too old → encode an IDR instead
+			ref_slot.reset();
+			poc_history.clear();
+		}
```
galister: I did wonder about this. I had no chance to test, but nothing in the spec indicated that it wouldn't work.

- `ref_slot` here is the local for `encode_info_next`.
  - base class will still do `setReferenceSlots`, but this is fine.
- no `ref_slot` → `encode_info_next` configures for IDR instead of P-frames.
- spec hints that reference slots are ignored in case of IDR/CRA, so they're ok to be set in this case.

server/encoder/video_encoder_vulkan_h265.cpp
galister: for `find_lsb` we could technically use `std::countr_zero(v)` instead, but if `v` is zero then it'll give the wrong result, so we can't really have a one-liner → would need to keep the helper function either way

for `find_msb` it'd look something like `31 - std::countl_zero(v)` but again `v == 0` needs to be handled separately.

keep in mind i'm not familiar with the c++ std library. it's the first time i'm hearing about these functions. but this is the idea i'm getting from reading the reference pages. please correct me if i'm wrong.

i think this `for` implementation is easier to comprehend for both cases.

xytovl:
Needs doc and [dashboard](https://github.com/WiVRn/WiVRn/blob/master/dashboard/settings.cpp#L472) updates to allow vulkan/h265

server/encoder/video_encoder_vulkan_h265.cpp
xytovl: If the edge cases (returning UINT32_MAX) are supposed to be used, then
```cpp
uint32_t max_transform_hierarchy = (find_msb((uint32_t)encode_h265_caps.ctbSizes) + 4) - (find_lsb((uint32_t)encode_h265_caps.transformBlockSizes) + 2);
```
is going to overflow, is this intended?

server/encoder/video_encoder_vulkan_h265.cpp
xytovl, line 427: This happens after `video_cmd_buf.beginVideoCodingKHR` has been called, which already uses ref_slot. I don't think this will work.

dashboard/settings.cpp
```diff
@@ -470,7 +470,11 @@ QList<Settings::video_codec> Settings::allowedCodecs() const
 			};
 		case encoder_name::X264:
 		case encoder_name::Vulkan:
-			return {video_codec::H264};
+			return {
+			        video_codec::CodecAuto,
+			        video_codec::H264,
+			        video_codec::H265,
+			};
```
xytovl, line 476: x264 falls through, vulkan should be be its own clause

PR #572, opened 2025-11-02
nvenc refactoring

aomiki:
Tidying up of the encoder files here and there

### DIFF
--- a/server/encoder/nvenc/nvenc_helper.h
+++ b/server/encoder/nvenc/nvenc_helper.h
@@ -23,6 +23,7 @@
 #include "wivrn_packets.h"
 #include <ffnvcodec/nvEncodeAPI.h>
 #include <memory>
+#include <system_error>
 
 #define NVENC_CHECK_NOENCODER(x)                                          \
 	do                                                                \
@@ -33,17 +34,6 @@
 			U_LOG_E("%s:%d: %d", __FILE__, __LINE__, status); \
 			throw std::runtime_error("nvenc error");          \
 		}                                                         \
-	} while (0)
-
-#define NVENC_CHECK(x)                                                                                                                  \
-	do                                                                                                                              \
-	{                                                                                                                               \
-		NVENCSTATUS status = x;                                                                                                 \
-		if (status != NV_ENC_SUCCESS)                                                                                           \
-		{                                                                                                                       \
-			U_LOG_E("%s:%d: %d, %s", __FILE__, __LINE__, status, shared_state->fn.nvEncGetLastErrorString(session_handle)); \
-			throw std::runtime_error("nvenc error");                                                                        \
-		}                                                                                                                       \
 	} while (0)
 
 #define CU_CHECK(x)                                                                               \
@@ -63,6 +53,32 @@ bool operator==(const GUID & l, const GUID & r);
 
 namespace wivrn
 {
+
+class nvenc_error_category : public std::error_category
+{
+protected:
+	std::shared_ptr<video_encoder_nvenc_shared_state> shared_state;
+	void * session_handle;
+
+public:
+	nvenc_error_category(std::shared_ptr<video_encoder_nvenc_shared_state> sharedState) : shared_state(sharedState), session_handle(nullptr)
+	{
+	}
+
+	nvenc_error_category(std::shared_ptr<video_encoder_nvenc_shared_state> sharedState, void * sessionHandle) : shared_state(sharedState), session_handle(sessionHandle)
+	{
+	}
+
+	const char *
+	name() const noexcept override
+	{
+		return "nvenc";
+	}
+
+	std::string
+	message(int nvenc_err) const override;
+};
+
 GUID encode_guid(wivrn::video_codec codec);
 
 int get_caps(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encode_guid, NV_ENC_CAPS caps);
--- a/server/encoder/nvenc/nvenc_helper.cpp
+++ b/server/encoder/nvenc/nvenc_helper.cpp
@@ -28,8 +28,35 @@ bool operator==(const GUID & l, const GUID & r)
 	       std::ranges::equal(l.Data4, r.Data4);
 }
 
+#define NVENC_CHECK(x)                                                                                              \
+	do                                                                                                          \
+	{                                                                                                           \
+		NVENCSTATUS status = x;                                                                             \
+		if (status != NV_ENC_SUCCESS)                                                                       \
+		{                                                                                                   \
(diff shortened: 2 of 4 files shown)


server/encoder/nvenc/nvenc_helper.h
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include "util/u_logging.h"
+#include "video_encoder_nvenc_shared_state.h"
+#include "wivrn_packets.h"
+#include <algorithm>
```
xytovl: This one should not be needed in the header

server/encoder/nvenc/nvenc_helper.h
```diff
@@ -0,0 +1,73 @@
...
+bool operator==(const GUID & l, const GUID & r);
+
+namespace wivrn
+{
+GUID encode_guid(wivrn::video_codec codec);
+
+int get_caps(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encode_guid, NV_ENC_CAPS caps);
+void check_encode_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encode_guid);
+void check_preset_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encode_guid, GUID preset_guid);
+void check_profile_guid_supported(std::shared_ptr<video_encoder_nvenc_shared_state> shared_state, void * session_handle, GUID encodeGUID, GUID profileGUID, std::string err_msg = "GPU doesn't support selected encoding profile.");
```
xytovl, line 87: Would it make sense to move those as member functions of the `video_encoder_nvenc_shared_state`?

server/encoder/nvenc/nvenc_helper.h
```diff
@@ -0,0 +1,73 @@
...
+	} while (0)
+
+#define NVENC_CHECK(x)                                                                                                                  \
+	do                                                                                                                              \
+	{                                                                                                                               \
+		NVENCSTATUS status = x;                                                                                                 \
+		if (status != NV_ENC_SUCCESS)                                                                                           \
+		{                                                                                                                       \
+			U_LOG_E("%s:%d: %d, %s", __FILE__, __LINE__, status, shared_state->fn.nvEncGetLastErrorString(session_handle)); \
+			throw std::runtime_error("nvenc error");                                                                        \
```
xytovl: The dashboard is now able to show exception contents, it may be useful to include the error string here.

Maybe something similar to [client/xr/check.h](https://github.com/WiVRn/WiVRn/blob/master/client/xr/check.h) with an nvenc error category, and the message would be whatever nvenc returns.

server/encoder/nvenc/nvenc_helper.h
aomiki: I thought about that, but so many functions potentially can depend on it, that are unrelated to what it is doing right now - only managing the state.

But it would make one less argument, so either way we gain one we lose one.

xytovl: What it is really is a proxy to cuda/nvenc methods, let's rename it.

aomiki:
Please let me know if it's the correct usage of `error_category`. For some reason it was pretty confusing for me to start using it, but it works now. 

The only weird thing is that encoder's constructor emits errors only to logs, no GUI errors. But `get_max_size` emits errors to GUI just fine.
<details><summary>Example from get_max_size</summary>
<p>

image

</p>
</details> 

I'll then do the same for cuda and bring all the macros to the same format.

xytovl:
> Please let me know if it's the correct usage of `error_category`. For some reason it was pretty confusing for me to start using it, but it works now.
> 
> The only weird thing is that encoder's constructor emits errors only to logs, no GUI errors. But `get_max_size` emits errors to GUI just fine.
> Example from get_max_size
> 
> I'll then do the same for cuda and bring all the macros to the same format.

I forgot that converting the error code to message needs context and makes it more complex than I thought.
It's up to you to either use system_error or runtime_errors, the important part is to have some information in the exception so that it can flow to the UI.

aomiki:
> converting the error code to message needs context
<details><summary>My thoughts on it</summary>
<p>

I couldn't find any proper information on how long `error_category` and `system_error` live, whether nvenc session can be destroyed before the `message()` is called.
But from my testing it works fine and results in less duplication, since only `message()` creates the error text.
But if this is going to cause problems in the future, we can change it to pass message in `error_category` constructor right away, instead of context for later message retrieval.

</p>
</details> 

I prefer `system_error`, so if everything else generally looks good to you, I'll implement it fully.

I'm still curious if it is intended that encoder's constructor doesn't emit errors to UI.
I tried walking up the call stack and I guess only errors from `get_encoder_settings` are catched and sent to UI, which explains why errors from `get_max_size` are visible.

xytovl:
The constructor errors not reaching the UI is a bug, not a feature.
I suppose it's reaching https://github.com/WiVRn/WiVRn/blob/master/server/main.cpp#L273 and it's not publishing the error on dbus as https://github.com/WiVRn/WiVRn/blob/master/server/driver/wivrn_comp_target.cpp#L329 does.

The socket thing is because we have the UI process that is independent, errors are passed by signals on dbus. The server process forks to handle clients but does dbus in the parent process, and has a socket to communicate with the child.

xytovl:
For the error_category/system_error stuff, I think we should capture the error message at exception creation, `nvEncGetLastErrorString` may return incorrect data if other calls were made or the session closed.

The `system_error` constructor can take a message in addition to the code.
`nvenc_error_category::message(int nvenc_err)` could just return the enum name, `magic_enum` can help with conversion so you don't have to manually enumerate them.

server/encoder/nvenc/nvenc_helper.h
aomiki: After renaming, would it still make sense to move functions? I do feel like it would reduce clutter quite a bit.

xytovl: We have an object that essentially stores cuda and nvenc function pointers.
The get_caps and check_xxx functions are high level wrappers of some of those functions, they can be member functions.

One could go one step further and have a "session" object that holds the session, but I don't think we want to start a whole RAII wrapper for nvenc. My long term goal is to move to vulkan video encode only, so don't worry if the nvenc code is not the best on earth, the cleaner error messages are a great improvement already.

xytovl:
closing due to no activity and conflicts, you can reopen if you want to complete it

PR #574, opened 2025-11-02
Allow changing refresh rate during stream

ImSapphire:


xytovl:
Let's delay this a little bit, I'd like to freeze features until the next release.

PR #575, opened 2025-11-03
more descriptive vulkan encoder option

galister:
updated phrasing to line up with the other options:
- nvenc (NVIDIA GPUs)
- vaapi (AMD and Intel GPUs)

merge whenever convenient, I wouldn't want to batphone our translators just for this.


xytovl:
We'll need some feedback that it works properly on NVIDIA and Intel before rewording it like this

PR #585, opened 2025-11-07
Initial Galaxy XR port

### LINKED ISSUE #584: [Feature Request]: Galaxy XR support
### Feature description

Port WiVRn on Galaxy XR

### Reason for the feature

_No response_

### Proposed implementation

_No response_

### Additional context

_No response_

perillamint:
This PR is an initial attempt to port WiVRn to Galaxy XR.

Current limitation:

* Controller coordinate transform seems not correct (but I can't fix it because I don't have controller yet)
* When you put off headset, logcat screams with `Error getting tracked pose for HMD`.
* Face tracking is not working (yet)

This PR closes #584

### DIFF
--- a/client/hardware.cpp
+++ b/client/hardware.cpp
@@ -231,8 +231,9 @@ const char * permission_name(feature f)
 				case model::htc_vive_xr_elite:
 				case model::lynx_r1:
 				case model::unknown:
-					return nullptr;
+					break;
 			}
+			return nullptr;
 		case feature::eye_gaze:
 			switch (guess_model())
 			{


AndroidManifest.xml
```diff
@@ -47,6 +47,10 @@
     <uses-permission android:name="com.oculus.permission.FACE_TRACKING" />
     <uses-permission android:name="com.picovr.permission.FACE_TRACKING" />
     <uses-permission android:name="com.oculus.permission.BODY_TRACKING" />
+    <uses-permission android:name="android.permission.EYE_TRACKING_COARSE" />
+    <uses-permission android:name="android.permission.EYE_TRACKING_FINE" />
+    <uses-permission android:name="android.permission.FACE_TRACKING" />
+    <uses-permission android:name="andorid.permission.HAND_TRACKING" />
```
xytovl: typo andorid

AndroidManifest.xml
```diff
@@ -91,6 +104,11 @@
         <!-- https://developers.meta.com/horizon/documentation/native/android/mobile-passthrough-loading-screens -->
         <meta-data android:name="com.oculus.ossplash.background" android:value="passthrough-contextual"/>
 
+        <!-- https://developer.android.com/develop/xr/get-started#xr_activity_start_mode_full_space_unmanaged -->
+        <property
+            android:name="android.window.PROPERTY_XR_ACTIVITY_START_MODE"
+            android:value="XR_ACTIVITY_START_MODE_FULL_SPCAE_UNMANAGED" />
```
xytovl: typo SPCAE

client/hardware.cpp
```diff
@@ -166,6 +171,7 @@ XrViewConfigurationView override_view(XrViewConfigurationView view, model m)
 		case model::htc_vive_xr_elite:
 			return scale_view(view, 1920);
 		case model::lynx_r1:
+		case model::galaxy_xr:
```
xytovl: If we follow the other device's conventions, this would be `scale_view(view, 3552)`.
I believe on 140% scale (the default) it would fail to create encoders and any computer would struggle to render.

AndroidManifest.xml
perillamint: Lessons learned, do not write code when extremely sleepy. :laughing: 

Fixed. Thanks for pointing out that stupid mistake!

AndroidManifest.xml
perillamint: Thanks for pointing out, too!

client/hardware.cpp
perillamint: Applied, but, yes, I also think it should be adjusted base on the real-world test. 3552 * 3840 * 2 pixels are not that easy for modest computers (and I also doubt extreme high-end PCs will be able to do that too)

client/hardware.h
```diff
@@ -41,6 +41,7 @@ enum class model
 	htc_vive_xr_elite,
 	htc_vive_focus_vision,
 	lynx_r1,
+	samsung_galaxy_xr,
```
xytovl, line 44: For now we can keep it like this, but the solution doesn't scale very well.

I think I'll go for something with 2 levels like
```cpp
enum class meta
{
  quest_1,
  quest_2,
  quest_pro,
  quest_3s,
  quest_3,
};

enum class android_xr
{
  samsung_galaxy_xr,
  generic,
};

using model = std::variant<meta, android_xr>;
```
with obviously htc and pico, I suppose Lynx and unknown would get their own categories.

Permissions would only depend on the toplevel, the second one would be used for quirks or resolution override.

perillamint: I agree on it too. However, I think it should be handled in separate refactoring PR, not in here.

client/hardware.cpp
perillamint: Configured it to return `scale_view(view, 3552)`, and it seems the default value (140%) is fine, at least on Radeon RX 9070XT.

I think it is good to go with this value.

xytovl:
If you want to keep separate commits, please rework them into logical elements like Android XR stuff, Galaxy XR ones, initial run permission requests.
If you don't think it's important, I'll squash into a single commit instead.

I also triggered CI, which will report if there are any issues or formatting to be done

PR #588, opened 2025-11-07
Attempt on auto selecting the WiVRn audio sink.

xoniq:


xytovl:
I don't think WiVRn should alter the configuration, wireplumber is able to remember the selected sink and will restore it when it becomes available again.

What should be done instead is ship some configuration to increase the default priority so that the first time it appears it becomes default.
I don't know if pipewire/wireplumber offer some standard mechanism to provide hints for priority.

xytovl:
See https://danielpower.ca/blog/wireplumber-device-priority for a similar issue (but with alsa devices).

If we want to change priorities, we should install a file in `/usr/share/wireplumber/wireplumber.conf.d/`, with appropriate contents. For flatpak versions, this would have to be adapted to be done at run-time into user configuration instead.
I wouldn't care for the pulseaudio version, it's there for legacy reasons and barely works.

Issue #484, opened 2025-09-14
[Feature Request]: Forward keyboard/mouse inputs from client to server


Beyley:
### Feature description

When a keyboard or a mouse is connected to the client, it should forward those inputs over to the server, and act as if it was a normal wireless keyboard.

### Reason for the feature

I'm often not at my PC physically, in another room entirely or across the house, and being able to connect a wireless keyboard to my quest instead of my PC to do work in VR remotely would be very useful.

### Proposed implementation

Use the android keyboard/mouse APIs to read the data, and use the same input APIs that wlx-overlay-s uses to inject the keybaord/mouse inputs to the host. It's not a terrible amount of data to send over the wire, so I don't expect bandwidth to be an issue.

### Additional context

Meta has many keyboard-tracking related OpenXR extensions, but I don't think any are of particular use in this specific case.

galister:
proof of concept here:
https://github.com/WiVRn/WiVRn/tree/poc/runtime-lh-discovery

feel free to test. mouse is pretty bad because it will use the mouse acceleration and other settings from android, not the PC.

PR #597, opened 2025-11-09
keyboard & mouse forwarding

galister:
forward keyboard & mouse input from android client to the server via uinput.

notes: 
- there are specific keys and key combinations that the Android OS consumes and does not expose to apps, these can't be forwarded (e.g. super + return)
- mouse sensitivity & acceleration follows android settings
- there's no support for high-resolution wheel on android, so scrolling may not feel as smooth

there is a new checkbox on the dashboard that only appears if uinput is available and accessible.
if the feature is enabled, the server notifies the client to enable forwarding.

closes https://github.com/WiVRn/WiVRn/issues/484

### DIFF
--- a/dashboard/settings.cpp
+++ b/dashboard/settings.cpp
@@ -25,6 +25,7 @@
 #include <QList>
 #include <QObject>
 #include <cassert>
+#include <fcntl.h>
 #include <nlohmann/json.hpp>
 #include <qqmlintegration.h>
 
@@ -120,6 +121,7 @@ void Settings::emitAllChanged()
 	openvrChanged();
 	debugGuiChanged();
 	steamVrLhChanged();
+	hidForwardingChanged();
 }
 
 void Settings::load(const wivrn_server * server)
@@ -384,6 +386,22 @@ void Settings::set_application(const QString & value)
 		applicationChanged();
 }
 
+bool Settings::hidForwarding() const
+{
+	auto it = m_jsonSettings.find("hid-forwarding");
+	if (it != m_jsonSettings.end() and it->is_boolean())
+		return *it;
+	return false;
+}
+
+void Settings::set_hidForwarding(const bool & value)
+{
+	auto old = hidForwarding();
+	m_jsonSettings["hid-forwarding"] = value;
+	if (old != value)
+		hidForwardingChanged();
+}
+
 bool Settings::debugGui() const
 {
 	// Advanced options (debug window, steamvr_lh)
@@ -504,6 +522,7 @@ void Settings::restore_defaults()
 	m_jsonSettings.erase("scale");
 	m_jsonSettings.erase("bitrate");
 	m_jsonSettings.erase("application");
+	m_jsonSettings.erase("hid-forwarding");
 	m_jsonSettings.erase("debug-gui");
 	m_jsonSettings.erase("use-steamvr-lh");
 	m_jsonSettings.erase("tcp-only");
@@ -531,6 +550,27 @@ bool Settings::steamvr_lh() const
 #else
 	return false;
 #endif
+}
+bool Settings::hid_forwarding() const
+{
+	// only called from the UI thread → no locking needed
+	static std::optional<bool> can_open_uinput;
+	if (can_open_uinput.has_value())
+		return can_open_uinput.value();
+
+	can_open_uinput = false;
+	constexpr std::array paths = {"/dev/uinput", "/dev/input/uinput"};
+	for (const char * p: paths)
+	{
+		int fd = ::open(p, O_WRONLY | O_NONBLOCK);
+		if (fd >= 0)
+		{
+			::close(fd);
+			can_open_uinput = true;
+			break;
+		}
+	}
+	return can_open_uinput.value();
 }
 
 #include "moc_settings.cpp"
--- a/server/driver/wivrn_session.cpp
+++ b/server/driver/wivrn_session.cpp
@@ -21,6 +21,7 @@
 
 #include "accept_connection.h"
 #include "application.h"
+#include "configuration.h"
 #include "driver/app_pacer.h"
 #include "main/comp_compositor.h"
 #include "main/comp_main_interface.h"
@@ -54,7 +55,6 @@
 #include <vulkan/vulkan.h>
 
 #if WIVRN_FEATURE_STEAMVR_LIGHTHOUSE
-#include "configuration.h"
 #include "steamvr_lh_interface.h"
 #endif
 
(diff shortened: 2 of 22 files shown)


server/driver/wivrn_uinput.cpp
```diff
@@ -0,0 +1,232 @@
...
+		::close(mouse_fd);
+		mouse_fd = -1;
+	}
+	if (kbd_fd >= 0)
+	{
+		(void)ioctl(kbd_fd, UI_DEV_DESTROY);
+		::close(kbd_fd);
+		kbd_fd = -1;
+	}
+}
```
xytovl: This can be the destructor directly, or even the destructor removed if we use an raii wrapper for file descriptors.

`fd_base` in common/wivrn_sockets.h could be reused, by splitting it further to remove the counters in the base.
Or just have an `uinput_fd` class that does the ioctl+close on destruction so it's written only once.

server/driver/wivrn_session.h
```diff
@@ -125,6 +126,7 @@ class wivrn_session : public xrt_system_devices
 	std::unique_ptr<wivrn_fb_face2_tracker> fb_face2_tracker;
 	std::unique_ptr<wivrn_htc_face_tracker> htc_face_tracker;
 	std::vector<std::unique_ptr<wivrn_generic_tracker>> generic_trackers;
+	std::unique_ptr<wivrn_uinput> uinput_handler;
```
xytovl: I don't remember why we went with unique_ptr for the other objects, it should all be optionals instead.

galister: do you want me to do a refactor on the other ones, while at it?

galister: (in a separate commit)

server/driver/wivrn_uinput.cpp
```diff
@@ -0,0 +1,232 @@
...
+		return nullptr;
+	if (!instance->init_keyboard())
+		return nullptr;
+	if (!instance->init_mouse())
+		return nullptr;
+	return instance;
+}
+
+wivrn_uinput::wivrn_uinput() = default;
+
```
xytovl: Do the initialization in the constructor, don't be afraid to throw in case of error, catch in wivrn_session

xytovl: It is possible to forward errors to dbus/gui through https://github.com/WiVRn/WiVRn/blob/master/server/wivrn_ipc.h, see https://github.com/WiVRn/WiVRn/blob/master/server/driver/wivrn_comp_target.cpp#L329 for an example usage

server/driver/wivrn_session.h
xytovl: As you prefer, I have not checked if there is a valid reason to use pointers. Very likely some legacy reason for xdev compatibility.

dashboard/settings.cpp
```diff
@@ -529,5 +548,26 @@ bool Settings::steamvr_lh() const
...
+bool Settings::hid_forwarding() const
+{
+	// only called from the UI thread → no locking needed
+	static std::optional<bool> can_open_uinput;
+	if (can_open_uinput.has_value())
+		return can_open_uinput.value();
+
+	can_open_uinput = false;
+	constexpr std::array<const char *, 2> paths = {"/dev/uinput", "/dev/input/uinput"};
+	for (const char * p: paths)
```
xytovl: `std::array` can deduce the types, there's no need to write `<const char *, 2>`.
You can even pass the initializer list in the for.

dashboard/settings.cpp
```diff
@@ -529,5 +548,26 @@ bool Settings::steamvr_lh() const
...
+	{
+		int fd = ::open(p, O_WRONLY | O_NONBLOCK);
+		if (fd >= 0)
+		{
+			::close(fd);
+			can_open_uinput = true;
+			break;
+		}
+	}
```
xytovl, line 572: Can you check when running the server through systemd user service if it can open uinput?

If it can't, then we have to add the relevant config, and maybe move the "can" method to dbus to have an actual value.

galister: I don't have a setup with systemd. didn't occur to me that it would be a limiting factor

server/driver/wivrn_uinput.cpp
galister: I ended up just using fd_base as is after not being able to split it cleanly. Right now it'll have the counters that don't do anything in this case, but it can be refactored in a follow-up commit by someone who knows what they're doing. (I'd rather not have duplicate code for the sole reason that I can't do it properly)

galister:
client CI failing due to http error on harfbuzz

server/driver/wivrn_session.h
galister: i'll do that in a separate pr

server/driver/wivrn_session.cpp
```diff
@@ -582,6 +600,42 @@ void wivrn_session::operator()(from_headset::inputs && inputs)
 		right_controller.set_inputs(inputs, offset);
 }
 
+void wivrn_session::operator()(from_headset::hid_key_down && e)
+{
+	if (uinput_handler)
+		uinput_handler.value().send_key(e.key, true);
```
xytovl: I like `operator->` better

server/driver/wivrn_session.cpp
```diff
@@ -293,6 +293,24 @@ wivrn::wivrn_session::wivrn_session(std::unique_ptr<wivrn_connection> connection
 		system_name += " on WiVRn";
 		strlcpy(xrt_system.base.properties.name, system_name.c_str(), std::size(xrt_system.base.properties.name));
 	}
+
+	if (configuration().hid_forwarding)
+	{
+		try
+		{
+			uinput_handler = wivrn_uinput();
```
xytovl: We can use `uinput_handler.emplace()` to avoid doing a move.

server/driver/wivrn_uinput.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include <sys/ioctl.h>
+#include <sys/time.h>
+#include <unistd.h>
+
+#include <linux/input-event-codes.h>
+#include <linux/uinput.h>
+
+namespace
+{
+int open_uinput_node()
```
xytovl: If we return an `fd_base` here, we can be sure it gets closed in any error case.

We can then use exceptions for error management, letting the actual error reach the caller

server/driver/wivrn_uinput.cpp
```diff
@@ -0,0 +1,185 @@
...
+	constexpr std::array paths = {"/dev/uinput", "/dev/input/uinput"};
+	for (const char * p: paths)
+	{
+		int fd = ::open(p, O_WRONLY | O_NONBLOCK);
+		if (fd >= 0)
+			return fd;
+		else if (fd != ENOENT) // exists but inaccessbile
+			throw std::system_error(fd, std::generic_category(), "error while opening uinput");
+	}
+	return ENOENT;
```
xytovl: Should throw, we'd end up using ENOENT as a file descriptor

server/driver/wivrn_uinput.cpp
```diff
@@ -0,0 +1,185 @@
...
+int open_uinput_or_throw()
+{
+	constexpr std::array paths = {"/dev/uinput", "/dev/input/uinput"};
+	for (const char * p: paths)
+	{
+		int fd = ::open(p, O_WRONLY | O_NONBLOCK);
+		if (fd >= 0)
+			return fd;
+		else if (fd != ENOENT) // exists but inaccessbile
+			throw std::system_error(fd, std::generic_category(), "error while opening uinput");
```
xytovl: Per `open` docs:
> On error, -1 is returned and errno is set to indicate the error.

We should use `errno` instead of `fd`.

server/driver/wivrn_uinput.cpp
```diff
@@ -0,0 +1,185 @@
...
+	auto n = ioctl(fd, op);
+	if (n < 0)
+		throw std::system_error(n, std::generic_category(), "ioctl error");
+}
+void ioctl_or_throw(int fd, unsigned long op, unsigned long int arg)
+{
+	auto n = ioctl(fd, op, arg);
+	if (n < 0)
+		throw std::system_error(n, std::generic_category(), "ioctl error");
+}
```
xytovl, line 79: use `errno` in the error.

galister:
tested with changes up to now

PR #602, opened 2025-11-12
Update monado

ImSapphire:


xytovl:
Ideally the monado updates should be split so that each monado-rev bump corresponds to a breaking change.

In order to do so, what I usually do is I run a bisect on a monado repository at the target revision:
```bash
git bisect start
git bisect bad HEAD
git bisect good $(cat "$WIVRN_CLONE/monado-rev")
git bisect run ./bisect.sh
```
Where bisect.sh is:
```
#!/bin/bash

git rev-parse HEAD > $WIVRN_CLONE/monado-rev
exec ninja -C ~$WIVRN_CLONE/build-desktop/
```
This provides the first monado revision that doesn't build.

The goal of this manipulation is to make it easier to later bisect on WiVRn, if a regression happens with a monado update, it is then possible to bisect the monado-rev itself without needing to fix the patches.

ImSapphire:
I couldn't figure out how to rebase the foveation patch on top of `ef0f552c3d773a2a4a04236c5d5ca78e3f90aad4`

xytovl:
> I couldn't figure out how to rebase the foveation patch on top of `ef0f552c3d773a2a4a04236c5d5ca78e3f90aad4`

I have pushed the rebased patches for that commit

server/accept_connection.cpp
```diff
@@ -25,6 +25,11 @@
 
 #include <sys/poll.h>
 
+static void handle_event_from_main_loop(to_monado::stop)
+{
+	// Ignore stop request when no headset is connected
+}
+
```
xytovl: This function can be called from `wivrn_session::reconnect`, in this case it should terminate the monado session

server/ipc_server_cb.cpp
```diff
@@ -39,11 +39,22 @@ void ipc_server_cb::mainloop_leaving(ipc_server * server, xrt_instance * xrt_ins
...
+	IPC_INFO(server, "Client %u disconnected", client_id);
+}
+
 ipc_server_cb::ipc_server_cb() :
         ipc_server_callbacks{
                 .init_failed = method_pointer2<&ipc_server_cb::init_failed>,
                 .mainloop_entering = method_pointer2<&ipc_server_cb::mainloop_entering>,
                 .mainloop_leaving = method_pointer2<&ipc_server_cb::mainloop_leaving>,
+                .client_connected = method_pointer2<&ipc_server_cb::client_connected>,
+                .client_disconnected = method_pointer2<&ipc_server_cb::client_disconnected>,
```
xytovl, line 57: TBD if this could make the `quit_if_no_client` inside `wivrn_session::reconnect` less ugly.

The intent is to terminate the monado session when the headset is diconnected first and the OpenXR client disconnects later. The accept connection loop needs to exit.

PR #603, opened 2025-11-12
fix: correct typo in Japanese text (変わり → 代わり)

akazdayo:
## Summary
Fixed typo in Japanese text: 変わり → 代わり (alternative/substitute)

The correct word for "alternative" is "代わり", not "変わり".


galister:
neat, thanks

PR #616, opened 2025-11-17
Extend session loss timer

ImSapphire:
Give applications a reasonable amount of time to quit before losing the instance.

Applications may want to perform some tasks before leaving, e.g. VRChat does a fade out, leaves the world, and then deinits everything and closes, which may take too much time for the 200ms timeout. I feel like 10 seconds is reasonable, though 5 seconds is probably fine as well.


galister:
there's a `U_TIME_1S_IN_NS` which might be more suitable in this case

Issue #626, opened 2025-11-19
[Feature Request]: Implement Network Traffic Shaping (Pacing) and IDR Recovery Mechanism for High-Bitrate Stability


ZokuTe:
### Feature description

Video unavailable over UDP, periodic stuttering over TCP

### Reason for the feature

I am experiencing severe instability with streaming at 500 Mbps, despite verifying my network environment is capable of handling this load.

I have benchmarked my local network UDP throughput at 800 Mbps with 0% packet loss (using steady-stream testing tools like iperf).

However, when setting the WiVRn streaming bitrate to 500 Mbps, the stream fails immediately.

- UDP Mode: The stream suffers from severe visual artifacts and never recovers. [video](https://1drv.ms/v/c/a091b6ce7a81dc29/IQThwmtakyd_TZ2lM8-AuvLAAQWqmep-qkyNPNfyIULM2ec?width=1024&height=1024)
- TCP Mode: The stream exhibits rhythmic, periodic stuttering. [video, significant micro-bursts and the resulting performance degradation](https://1drv.ms/v/c/a091b6ce7a81dc29/IQQVyY2To7FpQLQrhJkxBQoIAQvhDdCWzFqgZ-_fZ0tJk5g?width=1024&height=1024)

### Proposed implementation

1. Implement Traffic Shaping (Pacing):
The application must not send large frames in a single burst. Please implement a pacer that spreads the transmission of a frame (UDP/TCP packets) evenly over the frame interval (e.g., 7-8ms).

- Goal: Convert the >1Gbps peaks into a smoother curve that stays within the proven network capacity.

2. Implement Robust IDR Recovery:
Ensure the client sends a high-priority Request_IDR packet immediately upon detecting packet loss or decoding errors to fix the "infinite artifact" issue in UDP or TCP mode.

### Additional context

1. The "Micro-burst" Problem (Why 500Mbps Stream fails on an 800Mbps Network):
Although my network can sustain a steady 800 Mbps, video encoding is inherently bursty.

- Without application-level Traffic Shaping, a 500 Mbps video stream generates I-frames (Keyframes) that are significantly larger than P-frames.
- When these large frames are flushed to the socket instantly, the instantaneous bitrate can spike well beyond 1 Gbps (far exceeding the 800 Mbps verifiable limit) for a few milliseconds.
- This causes "Micro-burst" packet loss at the router/switch buffer, even though average bandwidth usage appears safe.

2. UDP Failure Mode (Lack of Recovery):
[...]

Issue #612, opened 2025-11-14
[Bug]: bad nvenc configuration


ZokuTe:
### Describe the bug

![Image](https://github.com/user-attachments/assets/df3712ba-4e63-47d5-a452-a8986f19342e)

The figure above is the unmodified output from the master branch code.

![Image](https://github.com/user-attachments/assets/5709510b-fd00-4fd2-bbfb-b45fadf2051d)

The above figure shows the modified runtime result. All are 500 Mbps AV1 10‑bit, and it is evident that the modification provides better latency performance (Reduced from 130–140 ms to 90–105 ms.). However, after large-scale motion, the video still suffers from corruption and damage.

### Steps to Reproduce

clone the code, compile and run it, observe the behavior. Then modify the code, compile and run it again, and observe the behavior.

Patch:

```cpp
NV_ENC_RC_PARAMS video_encoder_nvenc::get_rc_params(uint64_t bitrate, float framerate)
{
	return {
	        .rateControlMode = NV_ENC_PARAMS_RC_CBR,
	        .averageBitRate = static_cast<uint32_t>(bitrate),
	        .maxBitRate = static_cast<uint32_t>(bitrate),
	        .vbvBufferSize = static_cast<uint32_t>(bitrate),
	        .vbvInitialDelay = static_cast<uint32_t>(bitrate / 2),
	        .multiPass = NV_ENC_TWO_PASS_QUARTER_RESOLUTION};
}
```

### WiVRn Version

master branch

### Server Config

```json
{
  "application": [
    "wlx-overlay-s"
  ],
  "bit-depth": 10,
  "bitrate": 500000000,
  "debug-gui": false,
  "encoders": [
    {
      "codec": "av1",
      "encoder": "nvenc",
      "width": 1.0
    }
  ],
  "hid-forwarding": false,
  "openvr-compat-path": "/opt/opencomposite",
  "scale": 0.800000011920929,
  "use-steamvr-lh": true
}
```

### What GPU did the bug occur on?

NVIDIA

### Vulkan Info

```text
==========
VULKANINFO
==========

Vulkan Instance Version: 1.4.328


Instance Extensions: count = 25
-------------------------------
VK_EXT_acquire_drm_display             : extension revision 1
VK_EXT_acquire_xlib_display            : extension revision 1
VK_EXT_debug_report                    : extension revision 10
VK_EXT_debug_utils                     : extension revision 2
VK_EXT_direct_mode_display             : extension revision 1
[...]

PR #279, opened 2025-02-24
Better defaults for NVENC

Kierek:
All settings were tested on the RTX 4080 Super, Quest 3 with 1.4 resolution scale and dual nvenc encoders.

It seems that the multipass encoder is a must. Without it, I experienced bitrate spikes when turning around in games. With it on, the bitrate is much more stable. Going from P4, LL, Single pass to P1, ULL, 1/4 pass results in similar quality with better encoder and decoder performances. There's a little drop in quality at low bitrates, but seems a little better at higher bitrates. The drop in latency numbers seems significant, so I guess any eventual quality loss can be mitigated with a slightly higher bitrate. Since these are the defaults Sunshine uses and almost the same is used by ALVR, I guess it's safe to change.

Before taking the screenshots with the graphs, I quickly rotated in the game to check for bitrate fluctuations.

Screenshots from tests:
https://drive.google.com/drive/folders/1XiOSHf1Dof0uS023TlPv5UZHNj9hF34a?usp=sharing

Comparison of the current preset (P4, tuning Low Latency, Single pass) to the proposed setting (P1, tuning Ultra Low Latency, Two-pass Quarter):

### H264 50

| Scene      | Config   | Encoding Latency        | Motion to Photons | Displayed on Graph |
|-----------|---------|-------------------------|-------------------|--------------------|
| Alyx      | Current | [1475, 1509]            | 55                | 42/45              |
| Skyrim_1  | Current | [2260, 2044]            | 68                | 40/50              |
| Skyrim_2  | Current | [2247, 2154]            | 69                | 44/51              |
| Alyx      | New     | [1051, 1052]            | 53                | 39/44              |
| Skyrim_1  | New     | [1575, 1467]            | 64                | 40/48              |
| Skyrim_2  | New     | [1508, 1555]            | 65                | 41/50              |

### H264 300

| Scene      | Config   | Encoding Latency        | Motion to Photons | Displayed on Graph |
|-----------|---------|-------------------------|-------------------|--------------------|
| Alyx      | Current | [1476, 1460]            | 56                | 43/48              |
| Skyrim_1  | Current | [2189, 2035]            | 71                | 46/51              |
| Skyrim_2  | Current | [2139, 2155]            | 74                | 48/56              |
| Alyx      | New     | [1610, 1579]            | 57                | 45/49              |
| Skyrim_1  | New     | [2091, 2050]            | 74                | 48/55              |
| Skyrim_2  | New     | [2030, 2165]            | 76                | 50/55              |

### H265 50

[...]

PR #644, opened 2025-11-29
Optimize NVENC configuration & Optimize the IDR handler.

ZokuTe:
After optimization, latency has decreased, and the issue mentioned in #626 and #612 where the image corruption could not be recovered has also been resolved. This PR mainly does the following:

- Non-reference frame tracking: Added support for tracking non-reference P-frames to avoid unnecessary IDR requests when non-ref frames are lost
- Enhanced IDR timeout handling: Added warning logs when IDR frames time out
- Optimized NVENC encoder settings: 
  - Doubled VBV buffer size for smoother streaming
  - Disabled lookahead for lower latency
  - Enabled non-reference P-frames support

These changes reduce unnecessary IDR frame requests and improve video stream stability, especially when frames are dropped or delayed.

### DIFF
--- a/server/encoder/idr_handler.cpp
+++ b/server/encoder/idr_handler.cpp
@@ -28,10 +28,6 @@ idr_handler::~idr_handler() = default;
 void default_idr_handler::on_feedback(const from_headset::feedback & f)
 {
 	std::unique_lock lock(mutex);
-	bool is_non_ref = non_ref_frames.erase(f.frame_index);
-	if (not non_ref_frames.empty() and f.frame_index > 500 and *non_ref_frames.begin() < f.frame_index - 500)
-		non_ref_frames.erase(non_ref_frames.begin(), non_ref_frames.lower_bound(f.frame_index - 500));
-
 	std::visit(utils::overloaded{
 	                   [](need_idr) {},
 	                   [](idr_received) {},
@@ -50,8 +46,8 @@ void default_idr_handler::on_feedback(const from_headset::feedback & f)
 			                   }
 		                   }
 	                   },
-	                   [this, &f, is_non_ref](running r) {
-		                   if (not f.sent_to_decoder and f.frame_index >= r.first_p and not is_non_ref)
+	                   [this, &f](running r) {
+		                   if (not f.sent_to_decoder and f.frame_index >= r.first_p and not is_non_ref_frame(f.frame_index))
 			                   state = need_idr{};
 	                   },
 	           },
@@ -89,7 +85,12 @@ bool default_idr_handler::should_skip(uint64_t frame_id)
 void default_idr_handler::set_non_ref(uint64_t frame_index)
 {
 	std::unique_lock lock(mutex);
-	non_ref_frames.insert(frame_index);
+	non_ref_frames[frame_index % non_ref_frames.size()] = frame_index;
+}
+
+bool default_idr_handler::is_non_ref_frame(uint64_t frame_index)
+{
+	return non_ref_frames[frame_index % non_ref_frames.size()] == frame_index;
 }
 
 default_idr_handler::frame_type default_idr_handler::get_type(uint64_t frame_index)
--- a/server/encoder/idr_handler.h
+++ b/server/encoder/idr_handler.h
@@ -22,7 +22,6 @@
 
 #include <cstdint>
 #include <mutex>
-#include <set>
 #include <variant>
 
 namespace wivrn
@@ -43,7 +42,6 @@ public:
 class default_idr_handler : public idr_handler
 {
 	std::mutex mutex;
-	std::set<uint64_t> non_ref_frames;
 	struct need_idr
 	{};
 	struct wait_idr_feedback
@@ -57,6 +55,7 @@ class default_idr_handler : public idr_handler
 		uint64_t first_p;
 	};
 	std::variant<need_idr, wait_idr_feedback, idr_received, running> state;
+	std::vector<uint64_t> non_ref_frames{512, uint64_t(-1)};
 
 public:
 	enum class frame_type
@@ -69,6 +68,7 @@ public:
 	void reset() override;
 	bool should_skip(uint64_t frame_id) override;
 	void set_non_ref(uint64_t frame_index);
+	bool is_non_ref_frame(uint64_t frame_index);
 	frame_type get_type(uint64_t frame_index);
 };
 } // namespace wivrn


xytovl:
@Kierek can you have a look? This reverts some of the changes done in #279 and I wonder if it behaves differently across hardware or if something else changed.

ZokuTe:
My machine is equipped with an RTX 4070 Max-Q. On my machine, disabling Multi-Pass at a bitrate of 200 Mbps can reduce latency for a single encoder, while enabling Multi-Pass can reduce latency when using multiple encoders.

xytovl:
The findings in the previous pull request was that multi-pass increases the encoding time a little, but decreases by about the same amount the decode time and in the end it would provide a better quality with the same end-to-end latency.

ZokuTe:
When NV_ENC_TWO_PASS_QUARTER_RESOLUTION is enabled, it looks like this. 
NV_ENC_TWO_PASS_QUARTER_RESOLUTION
As for why the chart shows a sawtooth pattern, I can’t figure it out either.

When multi-pass is disabled, it looks like this.
NV_ENC_MULTI_PASS_DISABLED
```json
  "encoders": [
    {
      "codec": "av1",
      "encoder": "nvenc",
      "width": 0.5
    },
    {
      "codec": "av1",
      "encoder": "nvenc",
      "offset_x": 0.5,
      "width": 0.5
    }
  ],
```

ZokuTe:
> @Kierek can you have a look? This reverts some of the changes done in #279 and I wonder if it behaves differently across hardware or if something else changed.


After testing, I found it matches the description in #279, so I will withdraw this modification.

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -135,8 +135,9 @@ NV_ENC_RC_PARAMS video_encoder_nvenc::get_rc_params(uint64_t bitrate, float fram
 	return {
 	        .rateControlMode = NV_ENC_PARAMS_RC_CBR,
 	        .averageBitRate = static_cast<uint32_t>(bitrate),
-	        .vbvBufferSize = static_cast<uint32_t>(bitrate / framerate),
-	        .vbvInitialDelay = static_cast<uint32_t>(bitrate / framerate),
+	        .maxBitRate = static_cast<uint32_t>(bitrate),
```
xytovl: Per nvEncodeAPI.h docs, this is ignored for CBR

xytovl:
>     * Enable Non-Reference P-Frames: This allows the encoder to generate P-frames that are not referenced by subsequent frames. If packet loss occurs, the client can discard these frames without affecting subsequent images, improving fault tolerance.

We currently issue an IDR in case of any packet loss, this is implemented in https://github.com/WiVRn/WiVRn/blob/master/server/encoder/idr_handler.cpp

For vulkan video encode, a specific implementation is used in https://github.com/WiVRn/WiVRn/blob/master/server/encoder/video_encoder_vulkan.cpp#L30

If we want to take advantage of the non-reference P frames, we would need to modify the idr handler so that it can ignore lost non-reference P frames.
The flag might still be interesting if the encoder can create a more efficient bitstream.

Kierek:
I was against disabling multiPass, but I see you resigned from that change.

Otherwise, looks good to me. It didn't affect my default settings much (AV1, 150mbit, dual encoders), though it maybe improved AV1 a little bit at 200mbit. 

Very high bitrate H264 (400+) is still unusable for me and I have no idea why, it worked fine with amd and vaapi, so no network issue.

ZokuTe:
> > ```
> > * Enable Non-Reference P-Frames: This allows the encoder to generate P-frames that are not referenced by subsequent frames. If packet loss occurs, the client can discard these frames without affecting subsequent images, improving fault tolerance.
> > ```
> 
> We currently issue an IDR in case of any packet loss, this is implemented in https://github.com/WiVRn/WiVRn/blob/master/server/encoder/idr_handler.cpp
> 
> For vulkan video encode, a specific implementation is used in https://github.com/WiVRn/WiVRn/blob/master/server/encoder/video_encoder_vulkan.cpp#L30
> 
> If we want to take advantage of the non-reference P frames, we would need to modify the idr handler so that it can ignore lost non-reference P frames. The flag might still be interesting if the encoder can create a more efficient bitstream.

I don’t think it’s necessary to modify the IDR handler. Enabling non-reference P-frames is beneficial and harmless. It does not increase decoding or encoding latency; on the contrary, it improves image stability. Having both features at the same time, I see no downside.

Even with an IDR handler, I still often see the image glitch and not recover, or get stuck with heavy artifacts. I don’t really know why, but it seems like non‑reference P‑frames help ease the problem a bit.

If the bitrate is too high, turning on non‑reference P‑frames usually lets the video bounce back after a freeze.

https://github.com/user-attachments/assets/7c4e0f5b-3ace-428b-9d0a-c4569ec4c4ff



Without them, it won’t recover.

https://github.com/user-attachments/assets/e6539394-22dc-4933-8d1d-72a3f4c090fd

ZokuTe:
I just ran into this problem again at 200 Mbps even with non‑reference P‑frames on. Looks like the improvement I thought I saw was just in my head.

https://github.com/user-attachments/assets/be1ac61e-1b56-4b4b-87d4-d38c29e9fbc2

xytovl:
Do you get any benefit with nvenc AV1 vs vulkan video h265?
I would rather improve the vulkan video solution, which on AMD is performing better than the vaapi one already.

I have not attempted vulkan video AV1 yet because h264/5 required significant effort to understand the codecs, AV1 would likely require a comparable amount of work. It is something that will eventually be done, especially has hardware encoders and decoders become more common.

caszuu:
Lightly tested on a quest 2 with an older GTX 1060 and at least comparing to `master`, latency is generally mostly same if not a little higher than on `master`.

H264 seems to be a little slower with ~5-8 extra ms of latency when running at 52Mbps, 100Mbps. When trying 200Mbps, both branches overload the decoder.

HEVC seems to be more latency stable that H264 overall and is nearly identical between branches.

Edit: this was done before the latest push, with it pulled high bandwidths now cause some persistent corruption when moving around and reconnect issues on my end.

ZokuTe:
> Do you get any benefit with nvenc AV1 vs vulkan video h265? I would rather improve the vulkan video solution, which on AMD is performing better than the vaapi one already.
> 
> I have not attempted vulkan video AV1 yet because h264/5 required significant effort to understand the codecs, AV1 would likely require a comparable amount of work. It is something that will eventually be done, especially has hardware encoders and decoders become more common.

Vulkan H.265 doesn’t work on my system at all; the stream starts out totally broken.
org meumeu wivrn local-20251201-013614

ZokuTe:
> Lightly tested on a quest 2 with an older GTX 1060 and at least comparing to `master`, latency is generally mostly same if not a little higher than on `master`.
> 
> H264 seems to be a little slower with ~5-8 extra ms of latency when running at 52Mbps, 100Mbps. When trying 200Mbps, both branches overload the decoder.
> 
> HEVC seems to be more latency stable that H264 overall and is nearly identical between branches.
> 
> Edit: this was done before the latest push, with it pulled high bandwidths now cause some persistent corruption when moving around and reconnect issues on my end.

Thanks for the detailed report. I found the cause of the persistent corruption and connection issues.

It was a logic bug in the IDR handler. When running at high bitrates, if the initial large IDR frame was lost due to network congestion, the server would get stuck in a state where it indefinitely sent non-reference P-frames without ever checking if the IDR needed to be resent. This caused the persistent corruption (client missing the reference frame) and eventual disconnects.

I've pushed a fix that.

xytovl:
> > Lightly tested on a quest 2 with an older GTX 1060 and at least comparing to `master`, latency is generally mostly same if not a little higher than on `master`.
> > H264 seems to be a little slower with ~5-8 extra ms of latency when running at 52Mbps, 100Mbps. When trying 200Mbps, both branches overload the decoder.
> > HEVC seems to be more latency stable that H264 overall and is nearly identical between branches.
> > Edit: this was done before the latest push, with it pulled high bandwidths now cause some persistent corruption when moving around and reconnect issues on my end.
> 
> Thanks for the detailed report. I found the cause of the persistent corruption and connection issues.
> 
> It was a logic bug in the IDR handler. When running at high bitrates, if the initial large IDR frame was lost due to network congestion, the server would get stuck in a state where it indefinitely sent non-reference P-frames without ever checking if the IDR needed to be resent. This caused the persistent corruption (client missing the reference frame) and eventual disconnects.
> 
> I've pushed a fix that.

The expectation of the IDR handler is to send IDR over TCP, then wait for a response before resuming the stream.
In the `wait_idr_feedback` state, it should skip frames until it got a response.
In the `encode` function, the `prefer_control` field of the result enforces TCP for I frames.

The discarding of frames is important, as TCP and UDP streams can get reordered, and if the n+1 frame is complete before the n one, we discard frame n, which could be an I-frame.
Under high network load, the big IDR is likely to be overtaken by the small P frame that follows (on UDP).

caszuu:
> > Lightly tested on a quest 2 with an older GTX 1060 and at least comparing to `master`, latency is generally mostly same if not a little higher than on `master`.
> > H264 seems to be a little slower with ~5-8 extra ms of latency when running at 52Mbps, 100Mbps. When trying 200Mbps, both branches overload the decoder.
> > HEVC seems to be more latency stable that H264 overall and is nearly identical between branches.
> > Edit: this was done before the latest push, with it pulled high bandwidths now cause some persistent corruption when moving around and reconnect issues on my end.
> 
> Thanks for the detailed report. I found the cause of the persistent corruption and connection issues.
> 
> It was a logic bug in the IDR handler. When running at high bitrates, if the initial large IDR frame was lost due to network congestion, the server would get stuck in a state where it indefinitely sent non-reference P-frames without ever checking if the IDR needed to be resent. This caused the persistent corruption (client missing the reference frame) and eventual disconnects.
> 
> I've pushed a fix that.

Thanks for the quick fix.

It *is better* than before but the issue still persists. Connects / reconnects now work but still stall for a few seconds after the app starts before showing the stream.

In the clip i try to change the bitrate over `wivrnctl` to better show what i mean. (i assume, perhaps wrongly, this also happens when the client connects) 

https://github.com/user-attachments/assets/091a3e85-1fda-458d-b8d9-9b0c711caaeb

Edit again: the last push seems to fix the (re)connect delay!

ZokuTe:
Thanks for testing. I just fixed a potential IDR bug and attempted to reduce latency by disabling PTD and manually specifying the pictureType. @caszuu  Could you please test it again? I'd like to know if these changes work well. For me, the former eliminated the occasional visual artifacts. The latter reduced the data volume and slightly lowered the latency.

ZokuTe:
> 
> The expectation of the IDR handler is to send IDR over TCP, then wait for a response before resuming the stream.
> In the `wait_idr_feedback` state, it should skip frames until it got a response.
> In the `encode` function, the `prefer_control` field of the result enforces TCP for I frames.
> 
> The discarding of frames is important, as TCP and UDP streams can get reordered, and if the n+1 frame is complete before the n one, we discard frame n, which could be an I-frame.
> Under high network load, the big IDR is likely to be overtaken by the small P frame that follows (on UDP).
> 

Fixed.

xytovl:
I am not sure about the changes to idr_handler.
The IDR itself should only happen on stream initialisation and in case of packet loss. In both cases we don't expect a smooth playback and the wait for ack should be acceptable.
In the proposed version, sending an IDR is followed by non-reference P frames on TCP, so those will only be added in transient state and there should be less than 10 each time. I doubt there is any need for them.

You are also adding the concept of framerate to the idr_handler class, which I don't think is necessary, the time out is just used as a safety mechanism in case the idr is discarded. Any constant is valid in that case and doesn't need to be a precise duration.

I am losing track of what issue the PR is fixing. I advise marking it as draft, doing all the testing needed and once we're clear with the actual fixes edit the first post to state the issue and how it is addressed. Of course don't hesitate to ask if there's anything unclear in the existing code or intended behaviour.

caszuu:
> Thanks for testing. I just fixed a potential IDR bug and attempted to reduce latency by disabling PTD and manually specifying the pictureType. @caszuu Could you please test it again? I'd like to know if these changes work well. For me, the former eliminated the occasional visual artifacts. The latter reduced the data volume and slightly lowered the latency.

The first change does work to eliminate *most* of the artifacts, some corruption *sometimes* appears from the `set-bitrate` command now but it's more of a rarity than a guarantee. All other cases seem to be artifact free. When running with the second change the headset only receives one frame before server asserts with

```
ERROR [encode] /home/caszu/WiVRn/server/encoder/video_encoder_nvenc.cpp:508: 20, 
 WARN [comp_wivrn_present_thread] encode error: nvenc error
ERROR [encode] /home/caszu/WiVRn/server/encoder/video_encoder_nvenc.cpp:508: 20, 
 WARN [comp_wivrn_present_thread] encode error: nvenc error
```

(`.enablePTD = 0` seems to be the cause)

ZokuTe:
The latest two commits fix bugs introduced due to a lack of understanding of IDR. This ensures that the video does not suffer from continuous corruption under extreme bitrates, even if the decoder struggles to handle the frames.

This is the performance of 400Mbps AV1 on Meta Quest 3.

https://github.com/user-attachments/assets/f888f898-f12d-451f-926c-a662790b639c

To some extent, it also reduces the data volume and alleviates network pressure.

@caszuu Could you please test it again?

ZokuTe:
> I am not sure about the changes to idr_handler. The IDR itself should only happen on stream initialisation and in case of packet loss. In both cases we don't expect a smooth playback and the wait for ack should be acceptable. In the proposed version, sending an IDR is followed by non-reference P frames on TCP, so those will only be added in transient state and there should be less than 10 each time. I doubt there is any need for them.
> 
> You are also adding the concept of framerate to the idr_handler class, which I don't think is necessary, the time out is just used as a safety mechanism in case the idr is discarded. Any constant is valid in that case and doesn't need to be a precise duration.
> 
> I am losing track of what issue the PR is fixing. I advise marking it as draft, doing all the testing needed and once we're clear with the actual fixes edit the first post to state the issue and how it is addressed. Of course don't hesitate to ask if there's anything unclear in the existing code or intended behaviour.

After re-reading idr_handler and nvenc, I realized I was indeed doing something stupid before, but this fix should be logically sound now.

caszuu:
> The latest two commits fix bugs introduced due to a lack of understanding of IDR. This ensures that the video does not suffer from continuous corruption under extreme bitrates, even if the decoder struggles to handle the frames.
> 
> This is the performance of 400Mbps AV1 on Meta Quest 3.
> redpandacompress_orgmeumeuwivrnlocal-20251201-170006-0.mp4
> 
> To some extent, it also reduces the data volume and alleviates network pressure.
> 
> @caszuu Could you please test it again?

Happy to report that all corruption is now fully fixed! Unfortunately for the data volume it is not as effective on my side. While hevc manages to keep its latency similar to master, on h264 latency now jumps from ~60ms to ~90ms probably due to following.

At least on my side, the download bandwidth varies dramatically based on headset movement sometimes jumping to twice the target which can potentially cause the decoder to temporarily get overloaded. (this happens on both hevc and h264)

https://github.com/user-attachments/assets/d5621872-7b65-4c1e-814b-01a31efa887b

For reference `master` keeps its bandwidth hard-fixed.

https://github.com/user-attachments/assets/547eb903-d778-450f-a427-2b3dd05c4243

ZokuTe:
```diff
--- a/server/encoder/video_encoder_nvenc.cpp
+++ b/server/encoder/video_encoder_nvenc.cpp
@@ -203,7 +203,6 @@ video_encoder_nvenc::video_encoder_nvenc(
        // Bitrate control
        config.rcParams = get_rc_params(bitrate, fps);
        config.rcParams.enableAQ = 1;
-       config.rcParams.zeroReorderDelay = 1;
        config.rcParams.enableNonRefP = 1;

        config.gopLength = NVENC_INFINITE_GOPLENGTH;
```
or
```diff
--- a/server/encoder/video_encoder_nvenc.cpp
+++ b/server/encoder/video_encoder_nvenc.cpp
@@ -202,7 +202,6 @@ video_encoder_nvenc::video_encoder_nvenc(

        // Bitrate control
        config.rcParams = get_rc_params(bitrate, fps);
-       config.rcParams.enableAQ = 1;
        config.rcParams.zeroReorderDelay = 1;
        config.rcParams.enableNonRefP = 1;
```
@caszuu  Could you apply this patch to the code and then test the latency again?

ZokuTe:
It seems that disabling zeroReorderDelay results in lower latency, which is quite counterintuitive.

ZokuTe:
@Kierek , @decruns Could you test the stability and latency performance?

caszuu:
> @caszuu Could you apply this patch to the code and then test the latency again?

Tried both patches and other variants but both `enableAQ` and `zeroReorderDelay` seem to have minimal effect on the latency.

But (for some reason) the change applied to `vbvBufferSize` and `vbvInitialDelay` is the cause of the data rate fluctuations, if i revert the change (leaving rest untouched) the data rate becomes very solid (sometimes even better than master) with comparable latency to master, sometimes maybe better but that is within error.

Could this be a generational difference then as i'm still using a Pascal card?

ZokuTe:
> > @caszuu Could you apply this patch to the code and then test the latency again?
> 
> Tried both patches and other variants but both `enableAQ` and `zeroReorderDelay` seem to have minimal effect on the latency.
> 
> But (for some reason) the change applied to `vbvBufferSize` and `vbvInitialDelay` is the cause of the data rate fluctuations, if i revert the change (leaving rest untouched) the data rate becomes very solid (sometimes even better than master) with comparable latency to master, sometimes maybe better but that is within error.
> 
> Could this be a generational difference then as i'm still using a Pascal card?

A bigger vbvBufferSize lets the encoder produce larger frames, which can cause bigger bitrate swings. Depending on the network and decoder performance, that might seriously hurt latency. I’ve changed it so the vbvBufferSize is set to the size of two frames. This way, you won’t see sudden, hard‑to‑recover blocky artifacts, and the bitrate stays more stable. @caszuu Please give it test.

caszuu:
Tested and everything now works perfectly, both bitrate and latency are now stable.

ZokuTe:
@xytovl I think there might not be any more changes, please review it.

server/encoder/video_encoder_nvenc.cpp
```diff
@@ -504,6 +511,9 @@ std::optional<video_encoder::data> video_encoder_nvenc::encode(uint8_t slot, uin
 	};
 	NVENC_CHECK(shared_state->fn.nvEncLockBitstream(session_handle, &buf_lock_params));
 
+	if (buf_lock_params.pictureType == NV_ENC_PIC_TYPE_NONREF_P)
+		idr_handler.set_non_ref(frame_index);
```
xytovl, line 515: I don't think non-ref P are worth the added complexity. Can we try with just the parameter tuning and no idr_handler changes?

ZokuTe: > I don't think non-ref P are worth the added complexity. Can we try with just the parameter tuning and no idr_handler changes?

I think non‑reference P‑frames help cut down error‑correction overhead, reduce the decoding dependency chain, and shave off a bit of latency. Plus, Vulkan Video can make use of non‑reference P‑frames too. All these benefits outweigh the extra complexity of adding another state to the idr_handler.

I’m not sure how complicated it is to configure Vulkan Video to insert non‑reference P‑frames, so I don’t really have the say on this change. If you think the complexity outweighs the benefits, we can just drop the modification.

xytovl: Vulkan video already has a resiliency mechanism, where frames are referenced only if the corresponding ack was received. For that it uses a specific implementation of the idr handler, the remaining issue is that the decoder apparently fails to skip the missing frame and seems to get confused with picture count.

If we keep the concept, it would be more efficient to store them in a fixed size std::vector. You can index them by frame number % size, store the frame number. If the frame number matches, then it's a non-ref frame. If it's a different number it's not. Constant time access, no memory allocation, no need to ever delete an element.
```c++
std::vector<uint64_t> non_ref_frames{512, uint64_t(-1)};

bool is_non_ref_frame(uint64_t id)
{
  return non_ref_frames[id % non_ref_frames.size()] == id;
}

void set_non_ref_frame(uint64_t id)
{
  non_ref_frames[id % non_ref_frames.size()] = id;
}
```

xytovl:
I think it's ok now, can you cleanup the commits into 2: one for tuning parameters and one for non-reference P frames?

ZokuTe:
> I think it's ok now, can you cleanup the commits into 2: one for tuning parameters and one for non-reference P frames?

Okay, I will do that.

xytovl:
You can just force-push on the branch, the pull request gets updated automatically

ZokuTe:
> You can just force-push on the branch, the pull request gets updated automatically

I will create two separate PRs to split these changes.

PR #648, opened 2025-12-03
Forward XR_ANDROID_face_tracking to server

ImSapphire:
Adds support for face tracking on capable Android XR headsets (Galaxy XR).

### DIFF
--- a/client/scenes/lobby_gui.cpp
+++ b/client/scenes/lobby_gui.cpp
@@ -1156,6 +1156,8 @@ void scenes::lobby::gui_first_run()
 		current_tab = tab::server_list;
 		config.first_run = false;
 		config.save();
+		ImGui::PopStyleVar(); // ImGuiStyleVar_ItemSpacing
+		return;
 	}
 
 	ImGui::PushStyleVar(ImGuiStyleVar_FrameRounding, 10);


common/wivrn_packets.h
```diff
@@ -38,7 +38,7 @@
 namespace wivrn
 {
 
-static constexpr int protocol_revision = 2;
+static constexpr int protocol_revision = 3;
```
xytovl: The revision is only useful if a change can't be detected by the hash function (change in enum values, constants). I usually reset it to 0 for other changes, or just ignore it.

server/driver/wivrn_android_face_tracker.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+		inout_value->face_expression_set_android.state = data.state;
+		inout_value->face_expression_set_android.is_valid = data.is_valid;
+		inout_value->face_expression_set_android.sample_time_ns = sample_time.count();
+
+		if (not data.is_valid)
+			return XRT_SUCCESS;
+
+		memcpy(&inout_value->face_expression_set_android.parameters, data.parameters.data(), sizeof(float) * data.parameters.size());
+		memcpy(&inout_value->face_expression_set_android.region_confidences, data.confidences.data(), sizeof(float) * data.confidences.size());
```
xytovl: I would prefer an `std::ranges::copy`, this removes the need for size and sizeof

server/driver/wivrn_htc_face_tracker.cpp
```diff
@@ -96,10 +96,10 @@ xrt_result_t wivrn_htc_face_tracker::get_face_tracking(enum xrt_input_name facia
 	if (facial_expression_type == XRT_INPUT_HTC_EYE_FACE_TRACKING)
 	{
 		cnx.set_enabled(to_headset::tracking_control::id::face, true);
-		auto [_, data] = face_list.get_at(at_timestamp_ns);
+		auto [sample_time, data] = face_list.get_at(at_timestamp_ns);
```
xytovl: The returned value is the extrapolation time, the difference between the requested time and the time of production.
I suppose `sample_time_ns` would be `at_timestamp_ns - sample_time`, the moment in time when it was produced on the headset. To be more accurate, the moment should be part of the data if it is available through OpenXR.

ImSapphire: The spec says (at least for ANDROID_face_tracking and FB_face_tracking2) that '`sampleTime` is an `XrTime` time at which the returned expressions are tracked or extrapolated to. It equals the time at which the expression weights were requested if the extrapolation at the time was successful.'

The way I'm interpreting it, `sample_time_ns` would just be `at_timestamp_ns`?

common/wivrn_packets.h
ImSapphire: Ah, I had incremented that when adding enum values, but I forgot that it detects changes in variants. Reset to 0

server/driver/wivrn_htc_face_tracker.cpp
xytovl: It should be state.sampleTime, converted to server time. We have it from the client already and are not doing any extrapolation so it remains the correct value.

server/driver/wivrn_session.cpp
```diff
@@ -1105,13 +1115,15 @@ xrt_result_t wivrn_session::get_roles(xrt_system_roles * out_roles)
 	return XRT_SUCCESS;
 }
 
+// FIXME: use feature_inc and feature_dec to toggle trackings
```
xytovl: I prefer the dynamic method where they get deactivated if unused for some time.
There's no need to do the polling if a tracker is created but is idle.

client/application.cpp
```diff
@@ -1238,6 +1238,8 @@ void application::initialize()
 	        XR_HTC_FACIAL_TRACKING_EXTENSION_NAME,
 	        XR_HTC_VIVE_XR_TRACKER_INTERACTION_EXTENSION_NAME,
 
+	        XR_ANDROID_FACE_TRACKING_EXTENSION_NAME,
+
```
xytovl: move it after EXT ones, the same order as OpenXR speces (KHR -> EXT -> vendor)

server/driver/wivrn_htc_face_tracker.cpp
ImSapphire: @xytovl should I just stuff it in the `tracking::android_face` packet?

PR #652, opened 2025-12-04
Optimize NVENC encoding settings

ZokuTe:
Optimized NVENC encoder settings:
- Doubled VBV buffer size for smoother streaming
- Disabled lookahead for lower latency

This pull request will close issue #612.

### DIFF
--- a/server/encoder/video_encoder_nvenc.cpp
+++ b/server/encoder/video_encoder_nvenc.cpp
@@ -135,8 +135,10 @@ NV_ENC_RC_PARAMS video_encoder_nvenc::get_rc_params(uint64_t bitrate, float fram
 	return {
 	        .rateControlMode = NV_ENC_PARAMS_RC_CBR,
 	        .averageBitRate = static_cast<uint32_t>(bitrate),
-	        .vbvBufferSize = static_cast<uint32_t>(bitrate / framerate),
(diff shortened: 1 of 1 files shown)


server/encoder/video_encoder_nvenc.cpp
```diff
@@ -491,8 +496,10 @@ std::optional<video_encoder::data> video_encoder_nvenc::encode(uint8_t slot, uin
 	{
 		case default_idr_handler::frame_type::i:
 			frame_params.encodePicFlags |= NV_ENC_PIC_FLAG_FORCEIDR | NV_ENC_PIC_FLAG_OUTPUT_SPSPPS;
+			frame_params.pictureType = NV_ENC_PIC_TYPE_IDR;
 			break;
 		case default_idr_handler::frame_type::p:
+			frame_params.pictureType = NV_ENC_PIC_TYPE_UNKNOWN;
```
xytovl, line 502: This was `NV_ENC_PIC_TYPE_P` in #644, is the difference intentional?

ZokuTe: > This was `NV_ENC_PIC_TYPE_P` in #644, is the difference intentional?

Directly setting `pictureType` explicitly to `NV_ENC_PIC_TYPE_P` may prevent the encoder from generating non‑reference P‑frames. This is a fix.

PR #653, opened 2025-12-04
Improve IDR handling and ignore dropped non-ref frames

ZokuTe:
- Handle dropped IDR frames by immediately transitioning back to `need_idr` state.
- Track non-reference P-frames in `video_encoder_nvenc` and prevent them from triggering unnecessary IDR requests when dropped.
- Add a timeout log for the IDR feedback.


server/encoder/idr_handler.cpp
```diff
@@ -51,6 +59,7 @@ void default_idr_handler::reset()
 	std::unique_lock lock(mutex);
 	U_LOG_D("IDR handler reset");
 	state = need_idr{};
+	non_ref_frames.assign(512, uint64_t(-1));
```
ZokuTe, line 62: This differs from #644; this change is a fix.

PR #654, opened 2025-12-05
Handle assignment of ANY_HAND_CONTROLLER devices from steamvr_lh

ImSapphire:


galister:
Not sure if I follow. Isn't ANY_HAND_CONTROLLER only used for Vive controllers? But those don't provide hand tracking?

PR #659, opened 2025-12-07
In-stream GUI bitrate adjustment

ImSapphire:


client/scenes/lobby_gui.cpp
```diff
@@ -646,16 +646,16 @@ void scenes::lobby::gui_settings()
...
+		auto intScale = int(current * 10);
 		const auto slider = ImGui::SliderInt(
 		        _("Resolution scale").append("##resolution_scale").c_str(),
 		        &intScale,
-		        50,
-		        350,
-		        fmt::format(_F("%d%% - {}x{} per eye"), int(width * current), int(height * current)).c_str());
+		        5,
+		        35,
+		        fmt::format(_F("{}0%% - {}x{} per eye"), intScale, int(width * current), int(height * current)).c_str());
```
xytovl: For this change it would be nice to do the change in localized messages as well.
And since we're doing the formatting ourselves, we may as well pass intScale * 10 and remove the hardcoded 0, so that we can change it to 5% increments if we want.

client/scenes/stream_gui.cpp
```diff
@@ -458,6 +460,13 @@ void scenes::stream::gui_settings(float predicted_display_period)
 		}
 	}
 
+	int bitrate_ = bitrate.load() / 10'000'000;
+	if (ImGui::SliderInt(_S("Bitrate (Mbps)"), &bitrate_, 1, 20, "%d0"))
+	{
+		bitrate.store(bitrate_ * 10'000'000);
+		send_settings_changed_packet(session, network_session.get(), config, bitrate_ * 10'000'000, predicted_display_period);
+	}
+
```
xytovl: Probably a big change, but I think it would be a better interaction to have the bitrate change similar to the foveation center update: use a small tooltip with thumbstick controls.

With the whole GUI we can't see the image center so users can't judge the quality difference.
Ideally we should also show the encode and decode times.

ImSapphire: Done

client/scenes/stream_gui.cpp
```diff
@@ -571,6 +579,27 @@ void scenes::stream::gui_settings(float predicted_display_period)
...
+void scenes::stream::gui_bitrate_settings(float predicted_display_period)
+{
+	ImGui::PushFont(nullptr, constants::gui::font_size_large);
+	ImGui::Text("%s", _S("Use the right thumbstick to change the bitrate"));
+	ImGui::Text("%s", _S("Press A to go back"));
+	ImGui::Text("%s", fmt::format(_F("Bitrate: {}Mbit/s"), bitrate / 1'000'000).c_str());
+	ImGui::PopFont();
+
+	// Maximum speed 1 rad/s
+	float delta = application::read_action_float(foveation_pitch).value_or(std::pair{0, 0}).second * 20'000'000.f * predicted_display_period;
```
xytovl: We should do some renaming:
1 rad/s -> 20Mb/s (or just say we limit the adjustment speed)
foveation_pitch -> settings_adjust

PR #661, opened 2025-12-08
Hand interaction cleanup

ImSapphire:


server/driver/wivrn_controller.cpp
```diff
@@ -511,8 +567,44 @@ make_binding_profiles()
...
+        {XRT_INPUT_HAND_AIM_POSE, XRT_INPUT_HAND_AIM_POSE},
+        {XRT_INPUT_HAND_PINCH_POSE, XRT_INPUT_HAND_PINCH_POSE},
+        {XRT_INPUT_HAND_PINCH_VALUE, XRT_INPUT_HAND_PINCH_VALUE},
+        {XRT_INPUT_HAND_PINCH_READY, XRT_INPUT_HAND_PINCH_READY},
+        {XRT_INPUT_HAND_POKE_POSE, XRT_INPUT_HAND_POKE_POSE},
+        {XRT_INPUT_HAND_AIM_ACTIVATE_READY, XRT_INPUT_HAND_AIM_ACTIVATE_READY},
+        {XRT_INPUT_HAND_AIM_ACTIVATE_VALUE, XRT_INPUT_HAND_AIM_ACTIVATE_VALUE},
+        {XRT_INPUT_HAND_GRASP_VALUE, XRT_INPUT_HAND_GRASP_VALUE},
+        {XRT_INPUT_HAND_GRASP_READY, XRT_INPUT_HAND_GRASP_READY},
+};
```
xytovl: Do we actually need a binding for hand interaction? I think Monado does it automatically for the profile that matches the name. There is no binding for touch on the default controller, so it should work here too.

server/driver/wivrn_controller.cpp
```diff
@@ -596,81 +687,92 @@ wivrn_controller::wivrn_controller(int hand_id,
...
+		SET_INPUT(INDEX, TRACKPAD_FORCE);
+		SET_INPUT(INDEX, TRACKPAD_TOUCH);
+		SET_INPUT(TOUCH_PRO, STYLUS_FORCE);
+
+		// Make sure everything is mapped
+		// for (const auto & item: inputs_array)
+		// {
+		//	assert(item.name);
+		// }
+
```
xytovl: We can't have that check any more, there will be holes for hand interaction specific inputs. I think it's ok to just remove this entirely.

PR #662, opened 2025-12-08
Transition app to STOPPING state on requested app exit

ImSapphire:
Also includes https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2685 as part of the Monado updates, which can be used to work around rendering bugs in Beat Saber.

### DIFF
--- a/server/driver/wivrn_session.h
+++ b/server/driver/wivrn_session.h
@@ -142,8 +142,8 @@ class wivrn_session : public xrt_system_devices
 	// run-time editable settings
 	thread_safe<from_headset::settings_changed> settings;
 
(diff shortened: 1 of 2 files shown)


server/driver/wivrn_session.h
```diff
@@ -141,7 +141,7 @@ class wivrn_session : public xrt_system_devices
 	std::shared_ptr<audio_device> audio_handle;
 
 	// when sessions shall be destroyed, key is timestap, value is client id
-	thread_safe<std::map<int64_t, int32_t>> session_loss;
+	thread_safe<std::map<uint32_t, int64_t>> session_loss;
```
xytovl, line 146: update the comment as well

PR #675, opened 2025-12-15
add instructions to help with steamvr_lh confusion

mitsuami-megane:
Most people get confused by the messages from lighthouse.so complaining about a missing lighthousedb or chaperone.

Many of them end up wasting hours of their lives trying to get a lighthousedb file, and then realize that the cause of the crash was just SteamVR being the wrong version.


xytovl:
The segfault error was for the previous WiVRn release paired with the current SteamVR.
The next breakage may be different, I am not sure the warning message will remain relevant.

galister:
Afaik current WiVRn still needs `previous` SteamVR.
It's also true that in an overwhelming number of cases now and in the past, it's been the lighthouse.so version mismatch that would cause segfaults there.

PR #696, opened 2025-12-27
Add controller offsets for Pico Neo 3

Sawkez:
Added lobby offsets for Pico Neo 3 controllers.
They don't match up exactly with the controller's physical position but are close to the ones seen in Pico OS.
pico-offset


xytovl:
Please format with clang-format, then I can merge

PR #708, opened 2025-12-30
Fix infinite loop in find_dir when reaching root directory

InconsolableCellist:
The previous loop condition `not copy.empty()` would never terminate at the filesystem root since root paths (e.g., "/" or "C:\") are never empty and their parent_path() returns themselves.

### DIFF
--- a/server/main.cpp
+++ b/server/main.cpp
@@ -146,12 +146,10 @@ static void append_delim(std::string & to, std::string_view what, char delim)
 // if it can't be found, return the full path
 static std::filesystem::path find_dir(const std::filesystem::path & d, const std::filesystem::path & needle)
 {
-	for (auto copy = d; ; copy = copy.parent_path())
+	for (auto copy = d; copy != copy.parent_path(); copy = copy.parent_path())
 	{
 		if (copy.filename() == needle)
 			return copy;
-		if (!copy.has_parent_path() || copy == copy.parent_path())
-			break;
 	}
 	return d;
 }


InconsolableCellist:
I've not yet nailed down generic steps to reproduce, but try making the path for your OpenVR runtime (the vrclient.so) inaccessible or invalid; it should then attempt to traverse up to `/` and get stuck, and this patch should fix that. It does in my environment: bazzite linux, wivrn `master`, trying to use the latest OpenComposite `vrclient.so` from their git repo

You can verify that the server is stuck in this condition with: 
1. Run `gdb wivrn-server` in the CLI, or check `running` in `wivrn-dashboard` and attach gdb; if you don't see the `Service published: <hostname>` in the log then:
2. ctrl-c gdb and do `thread apply all bt` and you should see a stack trace like:
```
Using host libthread_db library "/lib64/libthread_db.so.1".
WiVRn 25.12-19-g6356f37f starting
^C
Program received signal SIGINT, Interrupt.
...
(gdb) thread apply all bt

Thread 1 (Thread 0x7ffff5606780 (LWP 77397) "wivrn-server"):
#0  0x00007ffff69e2d39 in std::filesystem::__cxx11::path::operator=(std::filesystem::__cxx11::path const&) () from /lib64/libstdc++.so.6
#1  0x00007ffff69e419b in std::filesystem::__cxx11::path::parent_path() const () from /lib64/libstdc++.so.6
#2  0x00005555555ae55f in find_dir (d=filesystem::path "/var/home/user/Programs/OpenComposite" = {...}, needle=filesystem::path "io.github.wivrn.wivrn") at /home/user/WiVRn/server/main.cpp:149
#3  steam_command () at /home/user/WiVRn/server/main.cpp:170
#4  0x00005555555b2253 in inner_main (argc=argc@entry=1, argv=argv@entry=0x7fffffffdd88, show_instructions=true) at /home/user/WiVRn/server/main.cpp:890
#5  0x000055555559311d in main (argc=1, argv=0x7fffffffdd88) at /home/user/WiVRn/server/main.cpp:1013
(gdb) 
```
Note `find_dir` and the `__cx11::path::operator=`

ImSapphire:
> I've not yet nailed down generic steps to reproduce,

This happens when the OpenVR compat path starts with `/var` but doesn't have `io.github.wivrn.wivrn` in any parent paths.

server/main.cpp
```diff
@@ -146,10 +146,12 @@ static void append_delim(std::string & to, std::string_view what, char delim)
 // if it can't be found, return the full path
 static std::filesystem::path find_dir(const std::filesystem::path & d, const std::filesystem::path & needle)
 {
-	for (auto copy = d; not copy.empty(); copy = copy.parent_path())
+	for (auto copy = d; ; copy = copy.parent_path())
```
ImSapphire:
```suggestion
	for (auto copy = d; copy.has_parent_path(); copy = copy.parent_path())
```

server/main.cpp
```diff
@@ -146,10 +146,12 @@ static void append_delim(std::string & to, std::string_view what, char delim)
...
 	{
 		if (copy.filename() == needle)
 			return copy;
+		if (!copy.has_parent_path() || copy == copy.parent_path())
+			break;
```
ImSapphire:
```suggestion
```

ImSapphire (commented):
I think this would be a cleaner fix:

server/main.cpp
InconsolableCellist: Though I would've expected it to work, it seems that `has_parent_path()` returns true even on `/`. So when I applied this suggested commit and tested on my environment, I again got the same behavior as the original code. However, the condition `copy != copy.parent_path()` works, and then I was able to remove the ` if (!copy.has_parent_path() || copy == copy.parent_path()) break;`, making it cleaner. I'll commit that version to this PR and you can have a look.

InconsolableCellist:
Thanks, finally got a chance to test the proposed commit, and left a comment above, and a new commit ^

Issue #486, opened 2025-09-16
[Bug]: Easyeffects filtered application outputs produce incredible crackling


SpookySkeletons:
### Describe the bug

Audio which is first handled by easyeffects on the wivrn pipewire device produce a very noticeable constant crackling noise.

### Steps to Reproduce

Apply any filter to all audio outputs or individual apps in easyeffects and observe the issue

### WiVRn Version

25.8

### What GPU did the bug occur on?

AMD

### Vulkan Info

```text

```

### What HMD did the bug occur on?

Quest 3

### ADB Logs

```plain

```

GreenSwede:
I can confirm that this affects me as well. I didn't realize that EasyEffects was the culprit as I ran into a temprary fix by using this audio option:

<img width="430" height="78" alt="Image" src="https://github.com/user-attachments/assets/5507139d-790b-48f4-8664-5148732c269f" />

xytovl:
fixed in #716

xytovl:
previous fix reverted, fixed in #717

PR #716, opened 2026-01-01
Fix pipewire sink setup

chicco-carone:
This pull request improves the PipeWire audio backend by explicitly configuring stream latency and quantum size for both speaker and microphone devices, leading to more consistent and predictable audio performance. This was made to address the issue in #486 which after my testing was fixed.
The main issue that easyeffects had was that before it used a variable quantum size which would de-sync the audio and cause the crackling issue.

### DIFF
--- a/server/audio/audio_pipewire.cpp
+++ b/server/audio/audio_pipewire.cpp
@@ -108,7 +108,7 @@ struct pipewire_device : public audio_device
 			// Smaller buffers (<5ms) risk underruns, larger ones (>10ms) add perceptible latency
 			uint32_t quantum_size = (desc.speaker->sample_rate * 5) / 1000;
 			uint32_t frame_size = desc.speaker->num_channels * sizeof(int16_t);
-			
+
 			std::string rate_str = std::format("1/{}", desc.speaker->sample_rate);
 			std::string latency_str = std::format("{}/{}", quantum_size, desc.speaker->sample_rate);
 
@@ -151,8 +151,8 @@ struct pipewire_device : public audio_device
 			        .rate = desc.speaker->sample_rate,
 			        .channels = desc.speaker->num_channels,
 			};
-			
-			const spa_pod *params[1];
+
+			const spa_pod * params[1];
 			params[0] = spa_format_audio_raw_build(&b, SPA_PARAM_EnumFormat, &audio_info);
 
 			// Stream flags:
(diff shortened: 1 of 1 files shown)


server/audio/audio_pipewire.cpp
```diff
@@ -120,29 +128,37 @@ struct pipewire_device : public audio_device
 			                "Audio/Sink",
 			                PW_KEY_MEDIA_ROLE,
 			                "Game",
+			                PW_KEY_NODE_RATE,
+			                rate_str,
+			                PW_KEY_NODE_LATENCY,
+			                latency_str,
+			                "node.force-quantum",
```
galister: let's use `PW_KEY_NODE_FORCE_QUANTUM` here

https://docs.pipewire.org/group__pw__keys.html#gaa50ce0bca22904a417a4e9d05f8face6

server/audio/audio_pipewire.cpp
```diff
@@ -168,28 +192,36 @@ struct pipewire_device : public audio_device
 			                "Audio/Source",
 			                PW_KEY_MEDIA_ROLE,
 			                "Game",
+			                PW_KEY_NODE_RATE,
+			                rate_str,
+			                PW_KEY_NODE_LATENCY,
+			                latency_str,
+			                "node.force-quantum",
```
galister: `PW_KEY_NODE_FORCE_QUANTUM`

xytovl:
Are all the parameters and flags required? If you have a quick explanation on why they are needed, that would be helpful as comments

server/audio/audio_pipewire.cpp
```diff
@@ -104,6 +104,14 @@ struct pipewire_device : public audio_device
 			        .sample_rate = info.speaker->sample_rate,
 			};
 
+			uint32_t quantum_size = (desc.speaker->sample_rate * 5) / 1000;
+			uint32_t frame_size = desc.speaker->num_channels * sizeof(int16_t);
+			
+			char rate_str[32];
+			snprintf(rate_str, sizeof(rate_str), "1/%u", desc.speaker->sample_rate);
+			char latency_str[32];
+			snprintf(latency_str, sizeof(latency_str), "%u/%u", quantum_size, desc.speaker->sample_rate);
```
xytovl: Prefer [`std::format`](https://en.cppreference.com/w/cpp/utility/format/format.html) and `std::string`

chicco-carone:
Ok I added some comments for the new pipewire stream flags and fixed the formattation for the string

xytovl:
Just do a clang-format and I think it will be fine. Thanks for investigating and fixing the issue.

chicco-carone:
i just did it and also i've been testing it without issues for one hour

xytovl:
The change breaks some setups and also breaks device switching, I have reverted it.

chicco-carone:
What were the issues in the setup that where there were problems? Maybe I can try and replicate it and update this to work for both cases

xytovl:
switching devices from gnome settings would not have any effect, switching from qpwgraph would stop audio from the application.
People on discord said they had no sound at all.

galister:
I would think that `PW_STREAM_FLAG_DRIVER` might be the issue, does the crackling still occur with the same changes, but without `PW_STREAM_FLAG_DRIVER`?

PR #727, opened 2026-01-10
client: implement lying down mode

Kate6:
I find that with a lot of VR games, playing while lying down really helps keep the motion sickness in control.  This effectively lets you decouple the VR world from the real world direction of gravity so the in-world "forward" becomes the real world ceiling.


xytovl:
Technically this is not allowed per OpenXR specification, where LOCAL stage is gravity aligned.

At first glance the PR seems to be fine, I am however working on a separate feature in branch prediction-pacing, which will conflict with your work. I should be able to see soon-ish if prediction-pacing works, I'll have your PR on hold during that time.

xytovl:
The changes to tracking have been pushed to master, the file where client side tracking was heavily reworked, so I believe it won't merge and you will have to reimplement the changes.

CherrieTheShifter:
Yes this

AGENTS.md
xytovl, line 1: I strongly oppose LLM for software development, additional docs are welcome but not duplicates of existing ones.

PR #730, opened 2026-01-10
Add support for XR_FB_hand_tracking_aim

rustedsword:
This extension extends hand tracking by adding support for finger pinches and hand aim.

The Monado patch must go to Monado of course, but the upstream has moved to a python script. So, if this MR is good then I'll push it to Monado.

I've tested this on Quest 3.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.13-bin.zip
```
xytovl, line 3: This should be in a separate commit

server/driver/wivrn_controller.cpp
```diff
@@ -621,6 +621,7 @@ wivrn_controller::wivrn_controller(xrt_device_name name,
         pinch_ext(hand_id == 0 ? device_id::LEFT_PINCH_POSE : device_id::RIGHT_PINCH_POSE),
         poke_ext(hand_id == 0 ? device_id::LEFT_POKE : device_id::RIGHT_POKE),
         joints(hand_id),
+        hand_aim(hand_id),
```
xytovl: Shouldn't we merge `hand_joints_list` and `hand_aim_history`? They will always be used together

common/wivrn_packets.h
```diff
@@ -380,11 +380,21 @@ struct hand_tracking
...
 	};
+	struct aim_data
+	{
+		XrHandTrackingAimFlagsFB status;
+		XrPosef aim_pose;
+		float pinch_strength_index;
+		float pinch_strength_middle;
+		float pinch_strength_ring;
+		float pinch_strength_little;
+	};
```
xytovl, line 413: The `hand_tracking` packet is already very close to the network MTU.
Can you check with a wireshark capture that packets don't get fragmented?

gradle/wrapper/gradle-wrapper.properties
rustedsword: I removed it once, but it just do not want to go :) Need more sleep >_>

common/wivrn_packets.h
rustedsword: image

rustedsword: what packets to look for, I can't see hand_tracking for some reason -_-

rustedsword: Got it. But this is without patch:

image

rustedsword: This is with patch :(
image

Funny thing is that I made it with separate packets initially :/

xytovl: It would be better to keep it in the same packet.
We can find tricks to squeeze it in the existing one, quaternions could be stored as 3 floats instead of 4 and then we reconstruct the 4th one.

We could theoretically save 2 bytes by merging hand_id and the 2 bools used for optionals into a single byte (but the serialization library can't do that currently), use 2 bytes instead of 4 for the `XrHandTrackingAimFlagsFB` flags. It all depends on how many bytes we're chasing.

xytovl: And we have to remember that encrypted packets have 4 additional bytes for the initialization vector but the dissector is unable to read those. So the total amount of required savings is 4 bytes more than the cleartext version.

patches/monado/0009-Add-FB_hand_tracking_aim-support.patch
xytovl, line 1: Is there an open merge request to Monado?
I prefer having as few downstream patches as possible.

rustedsword: Not yet. I wanted to know if this MR would be useful at all before trying to merge something in Monado.
Also, this patch was made against **monado-rev**. Upstream needs a different patch due to changes there.

(It could help a little if you update Monado version :)

common/wivrn_packets.h
xytovl: @Meumeu started testing a 32 bit encoding of quaternions that appears to have sufficient accuracy, once implemented and merged this will free up space in the hand tracking packet to avoid fragmentation.

server/driver/wivrn_controller.cpp
rustedsword: Made the combined version (but didn't rename hand_joints_list to something 'common') . I've mixed feelings about this, because this extension will be not supported everywhere, but 10 samples of joints + aim stored together are using less memory than two histories.

common/wivrn_packets.h
Meumeu: It's done now, it frees 312 bytes in the hand tracking packet

xytovl:
I have just pushed to master changes to tracking, which conflict with this merge request, you should try to rebase or redo the changes on top of master.

The hand tracking packet was also reduced thanks to optimizations on quaternion encodings, your data should now fit in an UDP packet.

client/scenes/stream_tracking.cpp
```diff
@@ -537,21 +546,51 @@ void scenes::stream::tracking()
...
+									        .aim_pose = {
+									                .position = result.aim->aim_pose.position,
+									                .orientation = pack(result.aim->aim_pose.orientation),
+									        },
+									        .pinch_strength_index = result.aim->pinch_strength_index,
+									        .pinch_strength_middle = result.aim->pinch_strength_middle,
+									        .pinch_strength_ring = result.aim->pinch_strength_ring,
+									        .pinch_strength_little = result.aim->pinch_strength_little,
+									};
+								}
```
xytovl: We should move this logic inside `locate_hands` and make it return a `from_headset::hand_tracking` object or take a reference to such an object that it would fill.

client/scenes/stream_tracking.cpp
```diff
@@ -607,9 +646,12 @@ void scenes::stream::tracking()
 
 			for (const auto & i: hands)
 			{
-				auto & packet = packets[packet_count++];
-				packet.clear();
-				wivrn_session::stream_socket_t::serialize(packet, i);
+				if (i.joints or i.aim)
```
xytovl: The `if (i.joints)` was removed so that server knows immediately that tracking becomes invalid. Just send every hand tracking packet, even if empty.

server/driver/hand_joints_list.cpp
```diff
@@ -120,13 +171,42 @@ static xrt_hand_joint_set convert_joints(const std::optional<std::array<from_hea
...
+		        .joints = convert_joints(tracking.joints),
+		        .aim = convert_aim(tracking.aim),
+		};
 		add_sample(
 		        tracking.production_timestamp,
 		        tracking.timestamp,
-		        convert_joints(tracking.joints),
+		        data,
 		        offset);
+	}
```
xytovl: You can write it inline:

```cpp
		add_sample(
		        tracking.production_timestamp,
		        tracking.timestamp,
		        hand_tracking_data{
		        		.joints = convert_joints(tracking.joints),
		        		.aim = convert_aim(tracking.aim),
				},
		        offset);
```

server/driver/wivrn_controller.cpp
```diff
@@ -825,9 +825,20 @@ xrt_result_t wivrn_controller::get_hand_tracking(xrt_input_name name, int64_t de
...
+
+			// Populate FB_hand_tracking_aim data
+			out_value->aim.has_aim = data.aim.valid;
+			if (data.aim.valid)
+			{
+				out_value->aim.status = data.aim.status;
+				out_value->aim.aim_pose = data.aim.aim_pose;
+				std::copy(std::begin(data.aim.pinch_strength),
+				          std::end(data.aim.pinch_strength),
+				          std::begin(out_value->aim.pinch_strength));
```
xytovl: `std::ranges::copy` is better

server/driver/hand_joints_list.cpp
```diff
@@ -27,32 +28,82 @@ static_assert(XRT_HAND_JOINT_COUNT == XR_HAND_JOINT_COUNT_EXT);
...
+	hand_aim_data result;
+	result.valid = true;
+	// Use the status from the closer sample
+	result.status = t < 0.5f ? a.status : b.status;
+
+	// Interpolate pose using slerp for orientation and lerp for position
+	math_quat_slerp(&a.aim_pose.orientation, &b.aim_pose.orientation, t, &result.aim_pose.orientation);
+	result.aim_pose.position.x = a.aim_pose.position.x * (1 - t) + b.aim_pose.position.x * t;
+	result.aim_pose.position.y = a.aim_pose.position.y * (1 - t) + b.aim_pose.position.y * t;
+	result.aim_pose.position.z = a.aim_pose.position.z * (1 - t) + b.aim_pose.position.z * t;
```
xytovl: `std::lerp` is more readable, or even better `m_vec3_lerp` from Monado if we have the correct type.

server/driver/hand_joints_list.cpp
```diff
@@ -27,32 +28,82 @@ static_assert(XRT_HAND_JOINT_COUNT == XR_HAND_JOINT_COUNT_EXT);
...
+
+	// Interpolate pinch strengths
+	for (int i = 0; i < 4; i++)
+	{
+		result.pinch_strength[i] = a.pinch_strength[i] * (1 - t) + b.pinch_strength[i] * t;
+	}
```
xytovl: Use `std::lerp`, if possible replace the `4` with `std::size()`, or use `std::arrays` with enumerate or zip.

CherrieTheShifter:
Yes this

server/driver/hand_joints_list.cpp
rustedsword: Changed to m_vec3_lerp

server/driver/hand_joints_list.cpp
rustedsword: Changed to std::lerp  + std::size()

server/driver/wivrn_controller.cpp
rustedsword: Done

server/driver/hand_joints_list.cpp
rustedsword: Done

client/scenes/stream_tracking.cpp
rustedsword: Done

client/scenes/stream_tracking.cpp
rustedsword: Done

client/scenes/stream_tracking.cpp
```diff
@@ -182,46 +182,59 @@ class locate_spaces_functor
...
+static void locate_hands(xr::hand_tracker & hand, XrSpace space, XrTime time, from_headset::hand_tracking & out)
 {
-	auto joints = hand.locate(space, time);
+	auto result = hand.locate(space, time);
 
-	if (joints)
+	if (!result)
+		return;
+
+	std::array<from_headset::hand_tracking::pose, XR_HAND_JOINT_COUNT_EXT> poses;
```
xytovl, line 192: Could be `auto & poses = out.joints.emplace()`

patches/monado/0009-Add-FB_hand_tracking_aim-support.patch
xytovl: Is the current monado-rev good enough to have an upstream MR?

server/driver/hand_joints_list.cpp
```diff
@@ -120,13 +168,39 @@ static xrt_hand_joint_set convert_joints(const std::optional<std::array<from_hea
...
+	return {
+	        .valid = true,
+	        .status = input_aim->status,
+	        .aim_pose = xrt_cast(pose),
+	        .pinch_strength = {
+	                input_aim->pinch_strength_index,
+	                input_aim->pinch_strength_middle,
+	                input_aim->pinch_strength_ring,
+	                input_aim->pinch_strength_little,
+	        },
```
xytovl, line 190: In the XR spec and the client side, it's 4 named fields.

I would prefer a consistent representation, array or fields doesn't really matter, it may depend on feedback once the Monado side is merged.

xytovl:
The last comments are really minor and the code looks fine to merge.
Is there an PCVR app I could use to do some manual testing?

PR #732, opened 2026-01-11
Update flake to latest nixpkgs

Detegr:
- `invalidateFetcherByDrvHash` is now under testers
- Trim `monado-rev` to not break `fetchFromGitlab` if the file has trailing whitespace


ImSapphire:
Thanks! I was trying to get around to this myself, but I was confused by the fetcher error.

PR #755, opened 2026-02-01
fix(dashboard): add missing includes

damaestro:
The close() function requires <unistd.h> to be included. This was not caught on older toolchains which transitively included it through other headers, but newer GCC/glibc requires explicit includes.


damaestro:
This caused a FTBFS when we did the F44 mass rebuild https://bugzilla.redhat.com/show_bug.cgi?id=2435188

PR #762, opened 2026-02-03
feat(translations): update ru translation

8osm:


xytovl:
You should not merge master into your branch, it will confuse github and block the pull request.

It's safer to rebase instead and force push to the branch, if you're not familiar with git just notify me when translations are correct and I'll fix the history.

8osm:
> You should not merge master into your branch, it will confuse github and block the pull request.
> 
> It's safer to rebase instead and force push to the branch, if you're not familiar with git just notify me when translations are correct and I'll fix the history.

The one time I trusted github to work properly...

Ill fix it in a bit

Issue #465, opened 2025-08-24
[Bug]: Play for Dream headset support


purraricat:
### Describe the bug

![Image](https://github.com/user-attachments/assets/2cb4b49d-107d-485f-b88b-af7938363538)

Controllers do not work on the Play for Dream headset.
Pinching fingers with hand tracking does not click buttons.
Trying physically to press button with a finger does not work. 

### Steps to Reproduce

Installed v25.8 WiVRn-standard-release.apk on Play for Dream headset
Run the app
No controllers
Switched to hand tracking 
Hand pointer moves and highlights buttons 
Does not allow to click any button

### Expected behavior

Be able to click the buttons.

### WiVRn Version

v25.8

### Installation method

installed the APK from github

### What GPU did the bug occur on?

AMD

### What HMD did the bug occur on?

Other

xytovl:
Branch https://github.com/WiVRn/WiVRn/tree/feat/pfd-support adds the interaction profile for the Play For Dream headset.

Can you try it and report if controller works on that version? You need both the updated APK and server.

Maybe headset logs can help understand why hand tracking doesn't work as expected, can you run `adb logcat > /tmp/logcat.txt`, start the application, try to interact with buttons then attach the produced file on this issue?

purraricat:
Can you link the apk and SteamOS flatpak? Because I can not find it on that branch.

xytovl:
They're in https://github.com/WiVRn/WiVRn/actions/runs/17242289719

purraricat:
Controllers still do not appear in the test app. 

But I have figured out the hand tracking. I put the pointer on the button and move my hand to press empty space at the current position and the pointer presses the button.

So I got into to the app. But server menu is pined to the bottom on my view, so I can not reach it to click any buttons now.
I got the logcat too. It's the last run of the app, I clicked all buttons on the controllers and clicked couple buttons with my hand tracking in the app.

![Image](https://github.com/user-attachments/assets/e4b45d39-cec1-46e2-be66-3821ef77046b)

[logcatWiVR.txt](https://github.com/user-attachments/files/21992710/logcatWiVR.txt)

xytovl:
If the headset provides both hand and controller positions, we will show the hands. That's probably the reason why it never shows controllers.

An other possibility is this:
```
[...]

PR #763, opened 2026-02-04
Improve README

Bluscrn99:
Ajust the README so it's nicer and easier to read (in my opinion)


README.md
```diff
@@ -41,35 +68,34 @@ Avahi must be running:
...
-You can set an application to be started automatically when you connect your headset in the dashboard settings or [manually](docs/configuration.md#application)
+You can set an application to be started automatically when your headset is connected, in the dashboard settings or [manually](docs/configuration.md#application).
 
 ### Application list
 When the headset is connected and no XR application is running, it will show an application launcher. Applications in that list are sourced from:
 - Steam games that are flagged as VR. Steam may need to be restarted for the list to be updated when new games are installed.
 - .desktop files that contain `X-WiVRn-VR` in the `Categories` section. Files are searched in [standard locations](https://specifications.freedesktop.org/desktop-entry/latest/file-naming.html#desktop-file-id) which usually include `~/.local/share/applications` and `/usr/share/applications/`.
 
-### OpenVR and Steam games
+### OpenVR/Steam games
```
xytovl: The section is meant for OpenVR games and when running under Steam both OpenVR and OpenXR games. I don't think the slash conveys it better than the "and"

README.md
```diff
@@ -1,33 +1,60 @@
...
+| Quest 3 | ✓ | Android, tested |
+| Quest 3s | ✓ | Android, tested |
+| Quest Pro | ✓ | Android, tested |
+| Pico Neo 3 | ✓ | Android, tested |
+| Pico 4 | ✓ | Android, tested |
+| HTC Vive Focus 3 | ✓ | Android, tested |
+| HTC Vive XR Elite | ✓ | Android, tested |
+| Other Android VR | ? | Cannot know |
+| Any Non-Android VR | ✖ | Client is Android |
+| Any Non-VR Android | ✖ | VR required |
```
xytovl: We now also support the Samsung Galaxy XR.

We might want to add Play for Dream as unsupported due to https://github.com/WiVRn/WiVRn/issues/465

For the HTC line, we try to support them as good as we can, but the headset runtime just doesn't perform well enough and the experience is not good. We may need to add a note about that

There's a Linux non-Android client, which is used for debugging and currently has no hardware decoding or audio.

ImSapphire: imo the 'Reason' column should be replaced with 'Notes', for most headsets it could be left blank.

README.md
Bluscrn99: changed this back

PR #767, opened 2026-02-06
Log result on failed calls

ImSapphire:


server/driver/wivrn_session.cpp
```diff
@@ -349,7 +357,7 @@ xrt_result_t wivrn::wivrn_session::create_session(std::unique_ptr<wivrn_connecti
 	auto xret = comp_main_create_system_compositor(&self->hmd, &ctf, &self->app_pacers, out_xsysc);
 	if (xret != XRT_SUCCESS)
 	{
-		U_LOG_E("Failed to create system compositor");
+		U_LOG_E("Failed to create system compositor: %s", xrt_result_to_string(xret).c_str());
 		return xret;
 	}
 	self->system_compositor = *out_xsysc;
```
xytovl, line 363: We could use `U_LOG_CHK_WITH_RET`

server/driver/wivrn_session.cpp
```diff
@@ -379,7 +387,7 @@ xrt_result_t wivrn::wivrn_session::create_session(std::unique_ptr<wivrn_connecti
 		}
 		if (res != XRT_SUCCESS)
 		{
-			U_LOG_W("failed to initialize eye tracker");
+			U_LOG_W("failed to initialize eye tracker: %s", xrt_result_to_string(res).c_str());
```
xytovl, line 390: Could be `U_LOG_CHK_ONLY_PRINT`

xytovl:
Actually I'm not sure about the other comments for the Monado macros, they don't seem to be much more readable, I'm just not a big fan of `xrt_result_to_string` and all the C macros

ImSapphire:
> I'm just not a big fan of `xrt_result_to_string` and all the C macros

yeah I agree, not sure how to make it better though

xytovl:
We may want to do something based on spdlog/libfmt, server logging is a bit of a mess currently.
let's go with this solution for now and we'll refactor later

PR #784, opened 2026-02-11
Re-enable microphone on reconnect

ImSapphire:


server/audio/audio_pipewire.cpp
```diff
@@ -342,6 +344,8 @@ void pipewire_device::mic_process(void * self_v)
 void pipewire_device::mic_state_changed(void * self_v, pw_stream_state old, pw_stream_state state, const char * error)
 {
 	auto self = (pipewire_device *)self_v;
+	self->mic_state = state;
+	U_LOG_I("Microphone state changed from %s to %s (error: %s)", magic_enum::enum_name(old).data(), magic_enum::enum_name(state).data(), error != nullptr ? error : "<unknown>");
```
xytovl, line 348: string_view may not be null-terminated, we have to go through a temporary std::string when using C logging apis

server/audio/audio_pipewire.cpp
```diff
@@ -61,6 +61,7 @@ struct pipewire_device : public audio_device
 	std::atomic<size_t> mic_buffer_size_bytes;
 	audio_data mic_current;
 	std::unique_ptr<pw_stream, deleter> microphone;
+	pw_stream_state mic_state{PW_STREAM_STATE_UNCONNECTED};
```
xytovl: This is accessed by multiple threads, should be an `std::atomic<std::underlying_type_t<pw_stream_state>>`

server/audio/audio_pipewire.cpp
xytovl: ok, https://github.com/Neargye/magic_enum/blob/master/doc/reference.md#enum_name states it's null-terminated. We can keep it like this.

PR #794, opened 2026-02-16
server: reconnect refactor revisited

caszuu:
This PR is an attempt at making a few fixes to the reconnect feature and making its existence a bit less detrimental to the rest of the project.

The primary goal is to make it less fragile by merging the first-time connects and reconnects under the same code path as much as possible. A headset connection now looks something like:
 - first-time connect: `create_session()` -> `start()` -> `resume_session()`
 - dis/reconnect: `pause_session()` -> `reconnect()` -> `resume_session()`

Headset descriptors are then sent from `resume_session()` making sure the headset receives all info necessary.

---

Fixed issues are:
 - missing `from_headset::headset_info_packet` checks on reconnect (incomplete, but works)
 - settings not being applied on reconnect
 - wrong openxr application session states on reconnect
 - refresh rate and user presence reset on reconnect
 - missing `try` / `catch` blocks for monado called code paths
 - refuse pair enable requests when server is active
   - this already did nothing, but no logs notified the user that the server didn't switch to pairing mode
- partial workaround for quest swapchain corruptions
  - fixed only by trial and error, from what i can tell wivrn does everything correctly but the runtime can still return unusable swapchains (causing image corruption, gpu page faults or full device losses)
  - by simply not recreating swapchains on scene changes the chance of getting a unusable swapchains are mostly mitigated, but the bug is probably *still present* 
- fixed a tracking thread softlock when `XrFrameState::predictedDisplayPeriod` is zero

Known issues still:
 - ~~a rare bug where inputs can get randomly disabled for some clients~~
   - have not gotten it since, probably fixed?
 - ~~the client sometimes struggles to (re)start decoding after connecting or waking up from sleep freezing the client~~
   - most likely a version of the tracking thread softlock

### DIFF
--- a/client/scenes/stream.cpp
+++ b/client/scenes/stream.cpp
@@ -766,7 +766,7 @@ void scenes::stream::render(const XrFrameState & frame_state)
 	if (state_ == state::shutdown)
 		application::pop_scene();
 
-	display_time_phase = frame_state.predictedDisplayTime % frame_state.predictedDisplayPeriod;
+	display_time_phase = frame_state.predictedDisplayPeriod ? frame_state.predictedDisplayTime % frame_state.predictedDisplayPeriod : 0;
 	display_time_period = frame_state.predictedDisplayPeriod;
 	real_display_period = last_display_time ? frame_state.predictedDisplayTime - last_display_time : frame_state.predictedDisplayPeriod;
 	last_display_time = frame_state.predictedDisplayTime;


xytovl:
The idea looks ok, the disabled input issue may be a race condition with `update_client_states`.

If we want to send warnings/errors, I think we should add localizable messages in the server, using language/country from the info packet.

server/driver/wivrn_comp_target.h
```diff
@@ -101,14 +101,18 @@ struct wivrn_comp_target : public comp_target
...
 
 	wivrn_comp_target(wivrn::wivrn_session & cnx, struct comp_compositor * c);
 	~wivrn_comp_target();
 
 	void on_feedback(const from_headset::feedback &, const clock_offset &);
 	void reset_encoders();
-	void set_bitrate(uint32_t bitrate_bps);
 
+	void pause_encoders();
+	void resume_encoders();
```
xytovl: Could be just pause/resume (same for audio), this may later be used for more than encoders.

server/driver/wivrn_comp_target.h
```diff
@@ -101,14 +101,18 @@ struct wivrn_comp_target : public comp_target
 	std::optional<wivrn_foveation> foveation;
 
 	std::atomic<float> requested_refresh_rate;
+	bool skip_encoding;
```
xytovl: Is used by more than one thread, should be atomic

client/scenes/lobby.cpp
```diff
@@ -882,10 +882,23 @@ void scenes::lobby::render(const XrFrameState & frame_state)
 
 	if (next_scene)
 	{
-		if (!next_scene->alive())
```
xytovl, line 885: Can we get rid of `alive()` and `exiting`? The state should be sufficient to handle both.

Or keep `alive()` as an accessor to state.

client/scenes/lobby.cpp
```diff
@@ -882,10 +882,23 @@ void scenes::lobby::render(const XrFrameState & frame_state)
...
+		{
+			case scenes::stream::state::streaming:
+				application::push_scene(next_scene);
+				break;
+
+			case scenes::stream::state::shutdown:
+				next_scene.reset();
+				[[fallthrough]];
+
+			case scenes::stream::state::shutdown_with_errors:
```
xytovl: It would be better to have a single`shutdown` state, then consume errors and when it's empty discard the scene.

server/driver/wivrn_session.cpp
```diff
@@ -1066,62 +1111,149 @@ void wivrn_session::quit_if_no_client()
...
-	if (not tcp)
+	if (prev_info.microphone and info.microphone)
 	{
-		request_stop();
-		return;
+		refuse |= prev_info.microphone->num_channels != info.microphone->num_channels;
+		refuse |= prev_info.microphone->sample_rate != info.microphone->sample_rate;
 	}
-	try
+	warn |= (bool)prev_info.microphone != (bool)info.microphone;
```
xytovl: Maybe out of scope, but we should be able to update the audio objects on the fly.

client/scenes/lobby.cpp
caszuu: This would probably be ideal. Although i was / am getting a few issues with what i assume are race conditions with other threads changing `state_` after it has been set to `shutdown`.

Currently it's *kind of* locked by the `exiting` atomic and putting a mutex just for it seems silly, so i'm unsure how to fix this.

client/scenes/lobby.cpp
caszuu: This is practically already done and i just didn't realize, will fix

client/scenes/lobby.cpp
xytovl: I see, one way to solve this is to have a (private) set_state method, which ensures shutdown remains final.

For the implementation, it's either a mutex or an atomic enum with compare and exchange.

server/driver/wivrn_session.cpp
caszuu: I think that we should at least add audio resets in another PR, this one is getting quite big. ^^;
For now the validate function checks if the session can be safely resumed without any changes.

client/scenes/lobby.cpp
caszuu: done.

caszuu:
This should be all of the server-side changes. (at least for this PR) The client-side still has a few issues (primarily the texture / swapchain corruption issue) but those can at least be worked around by restarting the client app.

client/scenes/lobby_gui.cpp
```diff
@@ -249,17 +249,47 @@ void scenes::lobby::gui_enter_pin(locked_notifiable<pin_request_data> & pin_requ
 
 void scenes::lobby::gui_connected(XrTime predicted_display_time)
 {
-	if (not next_scene)
```
xytovl, line 252: Add an `assert(next_scene)`

xytovl:
> This should be all of the server-side changes. (at least for this PR) The client-side still has a few issues (primarily the texture / swapchain corruption issue) but those can at least be worked around by restarting the client app.

Those corruption issues may happen if we are trying to draw on a swapchain that has been destroyed. This was the case where next_scene was deleted while it didn't finish rendering. The fix was to move the actual deletion on the next frame, maybe there's a similar issue remaining.

caszuu:
Not sure, it's possible since one time the the bug managed to overlay what i assume is the projection layer over the imgui window.

https://github.com/user-attachments/assets/825de889-d8df-43f6-b614-92d20a59f0a0

xytovl:
We recently fixed the `DebugWithValidationLayer` APK, it can help find the issue. It will catch usage of destroyed descriptors even if the driver would not fail.

caszuu:
This should probably be it for this PR. Found and fixed some minor bugs and tried to cleanup the commits a little. (issue status updated in top comment)

Also sorry for having you re-review this so many times, it wasn't intentional and is appreciated. ^^;

client/scenes/lobby_gui.cpp
caszuu: done

client/scenes/stream_tracking.cpp
```diff
@@ -351,7 +351,7 @@ void scenes::stream::tracking()
 				frame_duration = std::max<XrDuration>(5'000'000, display_time_period);
 				// Wait for next frame
 				XrTime now = instance.now();
-				pattern_begin = display_time_phase + (std::max(pattern_begin + frame_duration, now) / frame_duration) * frame_duration;
+				pattern_begin = std::min<XrTime>(5'000'000, display_time_phase) + (std::max(pattern_begin + frame_duration, now) / frame_duration) * frame_duration;
```
xytovl: phase is between 0 and frame time, the clamp value should be higher, I'd say 100ms (10 fps) is safer.

xytovl: Do you have evidence that `display_time_phase` is the issue? It is initialized to 0 and written in `scenes::stream::render` where it cannot be larger than `XrFrameState::predictedDisplayPeriod`

caszuu: This was very much a late night fix. 

I only noticed that near the session state transition from READY to SYNCHRONIZED the `display_time_phase` could pretty much contain `now()` (looking back it was probably `XrFrameState::predictedDisplayTime`) and reliably put the tracking thread to sleep for hours.

Will fix when I get home.

caszuu: Turns out it's only a simple divide-by-zero with `XrFrameState::predictedDisplayPeriod` bug, which then caused the thread to sleep for the full `predictedDisplayTime` amount of ns.

I assume this is technically allowed by OpenXR since it says `predictedDisplayPeriod` **may** be different from the display refresh rate, but not specifying how.

xytovl:
It looks fine now, I did some quick tests and it works.

The commits are quite nicely laid out, just the cleanup one that should be merged into previous ones it seems. If that's too annoying we can just merge like this.

server/driver/wivrn_session.h
```diff
@@ -140,7 +141,7 @@ class wivrn_session : public xrt_system_devices
 	bool connected();
 	const from_headset::headset_info_packet & get_info() const
 	{
-		return connection->info();
+		return headset_info;
```
ImSapphire, line 144: why was this moved from the wivrn_connection? On reconnect, connection->info will return the new one, so this seems unnecessary

caszuu: Primarily because the `headset_info` can now differ from the currently connected headset to some degree (see all the fields in `validate_headset_info()`), but it's also simpler for storing the "reference" headset info to check against in `reconnect()`

ImSapphire: ah, I see

ImSapphire (approved):

PR #807, opened 2026-02-25
Enable intra refresh for NVENC

ZokuTe:
This can solve the severe artifacts that occur during long-term streaming. Compared to implementing image quality detection on the client side, this solution is simpler.


xytovl:
Reading the nvEncodeAPI.h, there's a `NV_ENC_CAPS_SUPPORT_INTRA_REFRESH` capability. Unless it is supported on every known GPU, we should add a check before enabling the feature.

PR #809, opened 2026-02-25
Add port and hostname config options

Schmarni-Dev:
Adds a `port` config option to allow multiple WiVRn servers to run on the same host (assuming they run under different users).
Also adds a `hostname` option to override the name displayed in the server list, to allow differentiating different servers on the same host in a stable way.

### DIFF
--- a/dashboard/settings.cpp
+++ b/dashboard/settings.cpp
@@ -489,7 +489,7 @@ void Settings::set_openvr(const QString & value)
 	else
 		m_jsonSettings["openvr-compat-path"] = value.toStdString();
 	if (old != value)
-		tcpOnlyChanged();
+		openvrChanged();
 }
 
 QList<Settings::video_codec> Settings::allowedCodecs() const


xytovl:
Those options would need to be added in docs.

Dashboard should also be updated for firewall checks (dashboard/firewall.cpp), it could just disable checks if the port is not the default one.
Bonus points for exposing the port/hostname configuration options in the settings tab.

Schmarni-Dev:
i disabled the firewall warning when using a custom port, idk how to expose the port option through the settings ui, and idk if its helpful to expose the hostname without exposing the port

dashboard/settings.cpp
```diff
@@ -429,6 +429,40 @@ void Settings::set_tcpOnly(const bool & value)
 		tcpOnlyChanged();
 }
 
+int Settings::port() const
+{
+	auto it = m_jsonSettings.find("port");
+	if (it != m_jsonSettings.end() and it->is_number_integer())
+		return *it;
+	return 9757;
```
xytovl: Use the `default_port` constant from `wivrn_config.h`

dashboard/settings.cpp
```diff
@@ -429,6 +429,40 @@ void Settings::set_tcpOnly(const bool & value)
...
+	return "";
+}
+
+void Settings::set_hostname(const QString & value)
+{
+	auto old = hostname();
+	if (value == "")
+		m_jsonSettings.erase("hostname");
+	else
+		m_jsonSettings["port"] = value.toStdString();
```
xytovl: should be `"hostname"`

dashboard/settings.cpp
Schmarni-Dev: is there a good way to access that value in qml? i just hardcode the default port there (right next to it being hardcoded in the text)

Schmarni-Dev:
i noticed that `set_openvr` calls `tcpOnlyChanged` instead of `openvrChanged`, should i fix that here or in a separate pr?

dashboard/settings.cpp
xytovl: It's possible to have a property annotated with CONSTANT, meaning that it won't change (no signal, no setter). Then in cpp you provide the implementation to read a value.

See the `flatpak` property for an example.

xytovl:
> i disabled the firewall warning when using a custom port, idk how to expose the port option through the settings ui, and idk if its helpful to expose the hostname without exposing the port

It's fine, disabling the check prevents a broken behaviour (false positives and scripting to enable the default port only). Having a GUI to set port and hostname is independent and can be done later.

xytovl:
> i noticed that `set_openvr` calls `tcpOnlyChanged` instead of `openvrChanged`, should i fix that here or in a separate pr?

Just do a separate commit in this pr, it's fine

dashboard/settings.cpp
xytovl, line 124: Needs signals for new properties

xytovl:
One last thing, can you squash commits into one for openvrchanged fix, and as you prefer one for all the rest or one for hostname and one for port?

PR #823, opened 2026-03-06
add community links to readme

galister:
seen someone saying that it's hard to find the discord server for wivrn.

i think it can't hurt to add to the readme.


xytovl:
The matrix button renders as a simple link, but the discord one has logo + number of members, is that intended?

xytovl:
> The matrix button renders as a simple link, but the discord one has logo + number of members, is that intended?

I now see it correctly, it must have been an error on my network or something transient.

Issue #741, opened 2026-01-22
[Bug]: when passthrough mode activates or headset loses connection wivrn server gets perminantly stuck on 72 hz untill the server gets restarted


Redd56:
### Describe the bug

whenever i accidently start passthough mode or i lose connection to the wivrn server due to my wifi just randomly dying the wivrn server swaps to sending a 72hz signal

### Steps to Reproduce

use wivrn
swap to passthough
see 72hz once you get back

### Expected behavior

for wivrn to stick to whatever the original fps was

### WiVRn Version

wivrn server 25.12

### Installation method

aur

### Server Config

{
  "application": [
    "wlx-overlay-s"
  ],
  "bitrate": 150000000,
  "debug-gui": false,
  "encoder": "vaapi",
  "encoders": [
    {
      "codec": "h265",
      "encoder": "vaapi",
      "width": 0.5
    },
    {
      "codec": "h265",
      "encoder": "vaapi",
      "offset_x": 0.5,
      "width": 0.5
    }
  ],
  "hid-forwarding": false,
  "openvr-compat-path": "/opt/xrizer",
  "scale": 0.5099999904632568,
  "use-steamvr-lh": false
}

### Server Logs

wivrn-server
WiVRn 25.12 starting
Service published: reddspc
Client connected
Server started, PID 4050063
 INFO [ipc_server_main_common]  'v25.12' starting up...
Launching wlx-overlay-s
With args:
    "wlx-overlay-s"
Failed to open log file (path: Os { code: 2, kind: NotFound, message: "No such file or directory" }): /tmp/wlx.log
  2026-01-22T10:50:37.365925Z  INFO wlx_overlay_s: Welcome to wlx-overlay-s version 25.4.2-50c9449!
    at src/main.rs:98

  2026-01-22T10:50:37.365991Z  INFO wlx_overlay_s: It is Thu Jan 22 02:50:37 2026.
    at src/main.rs:103

DEBUG [comp_main_create_system_compositor] Doing init 0x55f0ee704c40
 WARN [compositor_check_and_prepare_xdev] Had to fill in meshuv on xdev 'WiVRn HMD', this should be done in the driver.
 INFO [create_device] Created device and COMPUTE queue with QUEUE_GLOBAL_PRIORITY_REALTIME.
 INFO [vk_print_device_info] Device info:
	name: AMD Radeon RX 7800 XT (RADV NAVI32)
	vendor: 0x1002
	product: 0x747e
[...]

PR #838, opened 2026-03-14
client: sync display and stream refresh rate after wake up

caszuu:
This is a quick patch for resetting refresh rate on meta headsets, which set their refresh rate to 72 when waking up.

Though ideally wivrn should be also able to return to the user specified refresh rate in general when the runtime changes it (perhaps after a small delay?), fixing #741 in return.

### DIFF
--- a/client/scenes/stream.cpp
+++ b/client/scenes/stream.cpp
@@ -1319,6 +1319,10 @@ void scenes::stream::on_xr_event(const xr::event & event)
 			});
 			break;
 		case XR_TYPE_EVENT_DATA_SESSION_STATE_CHANGED:
+			// sync display refresh rate with video stream when regaining focus
+			if (event.state_changed.state == XR_SESSION_STATE_READY and video_stream_description)
+				session.set_refresh_rate(video_stream_description->fps);
+
 			// Override session state if the GUI is interactable
 			if (event.state_changed.state == XR_SESSION_STATE_FOCUSED and is_gui_interactable())
 				network_session->send_control(from_headset::session_state_changed{


xytovl:
Are you confident we don't have a more direct mechanism to know about the changed refresh rate?

Also, did you reproduce the issue and confirm this fixes it? I don't think it's a dangerous change but it shouldn't be needed.

caszuu:
Well we have the refresh rate change events, but so far wivrn respects those and forwards them to the server.

On my end the bug does not manifest and i had to fake it, but i suspect only because the runtime fps change gets reverted by the video stream. This might not be the case if the pc is too slow at starting the stream and the 72 event gets to the server in time.

This issue is patched specifically because it has appeared quite a few times already:
https://discord.com/channels/1065291958328758352/1225819663024259082/1482443232473255946
https://discord.com/channels/1065291958328758352/1225819663024259082/1479786505416343562

xytovl:
My concern is that if we can't reproduce the issue, we need someone who is affected to test and confirm this fixes it.

For now we don't know why the headset changes refresh rate nor when it does the change.

moi952:
I have a Quest 3, and I'm willing to volunteer to test it.

I don't know if I have the issue; I started testing this tool today, and it's fantastic! It's a real change from SteamVR.

How do I see the headset's FPS?

I enjoy helping with open-source projects. I might have the problem, or maybe not, but if I can help, let me know :)

xytovl:
> I have a Quest 3, and I'm willing to volunteer to test it.
> 
> I don't know if I have the issue; I started testing this tool today, and it's fantastic! It's a real change from SteamVR.
> 
> How do I see the headset's FPS?
> 
> I enjoy helping with open-source projects. I might have the problem, or maybe not, but if I can help, let me know :)

The issues reported on discord are about the game switching to 72FPS, which may or may not be the headset refresh rate.
There isn't currently a method to show the game or headset refresh rate, so you need a game that is able to display it itself.

If on latest released version it sets the refresh rate to 72 when resuming from sleep and that with this branch you don't get the issue, this will confirm it is indeed a valid fix.

moi952:
> > I have a Quest 3, and I'm willing to volunteer to test it.
> > I don't know if I have the issue; I started testing this tool today, and it's fantastic! It's a real change from SteamVR.
> > How do I see the headset's FPS?
> > I enjoy helping with open-source projects. I might have the problem, or maybe not, but if I can help, let me know :)
> 
> The issues reported on discord are about the game switching to 72FPS, which may or may not be the headset refresh rate. There isn't currently a method to show the game or headset refresh rate, so you need a game that is able to display it itself.
> 
> If on latest released version it sets the refresh rate to 72 when resuming from sleep and that with this branch you don't get the issue, this will confirm it is indeed a valid fix.

This would allow for debugging, a bit like Sunshine/Moonlight; you can see the host framerate and the framerate actually displayed by the client.

Redd56:
ive got an easy way to test if you still need a tester, im just gonna need a bit of time or instructions on how to manualy build and run it as i normally run the aur version

xytovl:
The change is on the APK only, you can grab it from https://github.com/WiVRn/WiVRn/actions/runs/24497029194?pr=838 when CI is done building.
On the PC, build wivrn-full-git so you get a compatible version.

PR #842, opened 2026-03-18
Forward skeletal body tracking extensions

### LINKED ISSUE #421: [Feature Request]: Forward XR_FB_body_tracking from client
### Feature description

Implement forwarding of FB_body_tracking data from client

### Reason for the feature

I have an application which I would like to have consume FB_body_tracking data directly, however WiVRn does not forward this extension, it only generates trackers from this data.

### Proposed implementation

Monado already has this extension implemented, so it's only a matter of forwarding the tracking data from clients, and exposing the FB_body_tracking functions. I believe it should be simple to expose FB_body_tracking using the Pico and HTC body tracking extensions as data sources aswell.

Ideally, when an application enables FB_body_tracking, that application should no longer be able to enumerate the emulated trackers.

### Additional context

_No response_

ImSapphire:
Closes #421

### DIFF
--- a/client/configuration.cpp
+++ b/client/configuration.cpp
@@ -205,8 +205,6 @@ configuration::configuration(xr::system & system, xr::session & session)
 
 		if (auto val = root["fb_lower_body"]; val.is_bool())
 			fb_lower_body = val.get_bool();
-		if (auto val = root["fb_hip"]; val.is_bool())
-			fb_hip = val.get_bool();
 
 		if (auto val = root["virtual_keyboard_layout"]; val.is_string())
 			virtual_keyboard_layout = val.get_string().value();
@@ -320,7 +318,6 @@ void configuration::save()
 	json << ",\"passthrough_enabled\":" << std::boolalpha << passthrough_enabled;
 	json << ",\"mic_unprocessed_audio\":" << std::boolalpha << mic_unprocessed_audio;
 	json << ",\"fb_lower_body\":" << std::boolalpha << fb_lower_body;
-	json << ",\"fb_hip\":" << std::boolalpha << fb_hip;
 	json << ",\"enable_stream_gui\":" << std::boolalpha << enable_stream_gui;
 	for (auto & [key, value]: features)
 		json << "," << key << ":" << std::boolalpha << value;
--- a/client/configuration.cpp
+++ b/client/configuration.cpp
@@ -191,8 +191,6 @@ configuration::configuration(xr::system & system, xr::session & session)
 
 		if (auto val = root["fb_lower_body"]; val.is_bool())
 			fb_lower_body = val.get_bool();
-		if (auto val = root["fb_hip"]; val.is_bool())
-			fb_hip = val.get_bool();
 
 		if (auto val = root["virtual_keyboard_layout"]; val.is_string())
 			virtual_keyboard_layout = val.get_string().value();
@@ -309,7 +307,6 @@ void configuration::save()
 	json << ",\"forward_mouse\":" << std::boolalpha << forward_mouse;
 	json << ",\"forward_gamepad\":" << std::boolalpha << forward_gamepad;
 	json << ",\"fb_lower_body\":" << std::boolalpha << fb_lower_body;
-	json << ",\"fb_hip\":" << std::boolalpha << fb_hip;
 	json << ",\"enable_stream_gui\":" << std::boolalpha << enable_stream_gui;
 	for (auto & [key, value]: features)
 		json << "," << key << ":" << std::boolalpha << value;
--- a/client/configuration.cpp
+++ b/client/configuration.cpp
@@ -189,8 +189,24 @@ configuration::configuration(xr::system & system, xr::session & session)
 		if (auto val = root["forward_gamepad"]; val.is_bool())
 			forward_gamepad = val.get_bool();
 
-		if (auto val = root["fb_lower_body"]; val.is_bool())
-			fb_lower_body = val.get_bool();
+		if (auto val = root["body_parts"]; val.is_object())
+		{
+			for (const auto & [b, name]: magic_enum::enum_entries<wivrn::from_headset::body_part_mask>())
+			{
+				const auto & body_part = val[name];
+				if (!body_part.is_bool())
+					continue;
+
+				if (body_part.get_bool())
+				{
+					body_part_mask |= std::to_underlying(b);
+				}
+				else
+				{
+					body_part_mask &= ~std::to_underlying(b);
+				}
+			}
+		}
 
 		if (auto val = root["virtual_keyboard_layout"]; val.is_string())
 			virtual_keyboard_layout = val.get_string().value();
@@ -306,7 +322,21 @@ void configuration::save()
 	json << ",\"forward_keyboard\":" << std::boolalpha << forward_keyboard;
 	json << ",\"forward_mouse\":" << std::boolalpha << forward_mouse;
 	json << ",\"forward_gamepad\":" << std::boolalpha << forward_gamepad;
-	json << ",\"fb_lower_body\":" << std::boolalpha << fb_lower_body;
+
+	std::stringstream body_part_ss;
+	for (const auto & [b, name]: magic_enum::enum_entries<wivrn::from_headset::body_part_mask>())
+	{
+		bool enabled = (body_part_mask & std::to_underlying(b)) != 0;
+		body_part_ss << "\"" << name << "\":" << std::boolalpha << enabled << ",";
+	}
+
+	std::string body_parts_str = body_part_ss.str();
+	if (!body_parts_str.empty())
+	{
+		body_parts_str.pop_back(); // Remove last comma
+		json << ",\"body_parts\":{" << body_parts_str << "}";
+	}
+
 	json << ",\"enable_stream_gui\":" << std::boolalpha << enable_stream_gui;
 	for (auto & [key, value]: features)
 		json << "," << key << ":" << std::boolalpha << value;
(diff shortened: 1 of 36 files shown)


common/wivrn_packets.h
```diff
@@ -402,7 +413,84 @@ struct hand_tracking
...
+	// excluding root
+	struct fb_joints
+	{
+		std::array<packed_pose, XR_BODY_JOINT_COUNT_FB - 1> joints;
+	};
+	struct meta_joints
+	{
+		std::array<packed_pose, XR_FULL_BODY_JOINT_COUNT_META - 1> joints;
+	};
+	std::optional<std::variant<fb_joints, meta_joints>> joints;
```
xytovl: optional<variant<...>> could be variant<std::monostate, ...>

common/wivrn_packets.h
```diff
@@ -402,7 +413,84 @@ struct hand_tracking
...
-struct body_tracking
+struct meta_body
+{
+	enum flags : uint8_t
+	{
+		orientation_valid = 1 << 0,
+		position_valid = 1 << 1,
+		orientation_tracked = 1 << 2,
+		position_tracked = 1 << 3,
+	};
```
xytovl: Maybe have a single definition of pose flags for all packets

client/xr/body_tracker.cpp
```diff
@@ -18,18 +18,23 @@
...
-		if (fb_body_properties.supportsBodyTracking)
+		bool body_supported = system.fb_body_tracking_properties().supportsBodyTracking;
+		bool full_body_supported = instance.has_extension(XR_META_BODY_TRACKING_FULL_BODY_EXTENSION_NAME) and instance.has_extension(XR_META_BODY_TRACKING_FIDELITY_EXTENSION_NAME)
+		                                   ? system.meta_body_tracking_properties().supportsFullBodyTracking
+		                                   : false;
+		if (full_body_supported)
+			return body_tracker_type::meta;
+		if (body_supported)
 			return body_tracker_type::fb;
 	}
```
xytovl: I would use early return without the intermediate variables:

```cpp
        if (instance.has_extension(XR_FB_BODY_TRACKING_EXTENSION_NAME))                     
        {                                                                                                      
                if (instance.has_extension(XR_META_BODY_TRACKING_FULL_BODY_EXTENSION_NAME) and                                                                                
                    instance.has_extension(XR_META_BODY_TRACKING_FIDELITY_EXTENSION_NAME) and                                                                                    
                    system.meta_body_tracking_properties().supportsFullBodyTracking)
                        return body_tracker_type::meta;
                if (system.fb_body_tracking_properties().supportsBodyTracking)                                                                 
                        return body_tracker_type::fb;                                                   
        } 
```

client/xr/fb_body_tracker.cpp
```diff
@@ -82,44 +80,82 @@ std::optional<std::array<wivrn::from_headset::body_tracking::pose, wivrn::from_h
...
-		}
-		if (joint_loc.locationFlags & XR_SPACE_LOCATION_POSITION_VALID_BIT)
-		{
-			pose.flags |= wivrn::from_headset::body_tracking::position_valid;
-			pose.flags |= wivrn::from_headset::body_tracking::position_tracked;
-		}
-
-		poses[num_poses++] = pose;
+		spdlog::warn("xrGetBodySkeletonFB returned {}", xr::to_string(res));
+		return {};
```
xytovl: It doesn't look like it should return an optional, the empty case is an error case and should be handled by an exception

common/wivrn_packets.h
```diff
@@ -402,26 +423,85 @@ struct hand_tracking
...
+		XrVector3f position;
+		packed_quaternion orientation;
+		uint8_t flags;
+	};
+	struct packed_pose
+	{
+		struct
+		{
+			int16_t x, y, z;
+		} position;
```
xytovl, line 488: Could use a comment stating it's in 10th of mm relative to root

xytovl:
If I understand correctly, this will break compatibility with xrizer.
Ideally we should only merge into master when xrizer is able to process the new form of data, and update the version we ship in the flatpak.

ImSapphire:
yeah, I was thinking of re-adding the xdevs, then having a setting for which should be enabled that can be changed by the client, but I am not sure how that UI would look

xytovl:
Wouldn't it be easier to add support for full body tracking in xrizer instead?

ImSapphire:
Ideally I would do that, but some people may want to mix Quest IOBT + SlimeVR or lighthouse trackers, which wouldn't work when using the skeletal body tracking extensions

CherrieTheShifter:
yes this

client/scenes/gui_common.cpp
```diff
@@ -116,6 +119,84 @@ bool refresh_rate(
...
+	        std::make_pair(from_headset::body_part_mask::left_knee, _C("virtual body tracker selection", "Left knee")),
+	        std::make_pair(from_headset::body_part_mask::right_knee, _C("virtual body tracker selection", "Right knee")),
+	        std::make_pair(from_headset::body_part_mask::left_foot, _C("virtual body tracker selection", "Left foot")),
+	        std::make_pair(from_headset::body_part_mask::right_foot, _C("virtual body tracker selection", "Right foot")),
+	};
+
+	bool changed = false;
+	bool button_pressed = ImGui::ArrowButton("##OpenBodyPartMenu", ImGuiDir_Down);
+	ImGui::SameLine(0.f, ImGui::GetStyle().ItemInnerSpacing.x);
+	ImGui::Text("%s", _S("Virtual body trackers"));
```
xytovl, line 146: The arrow button doesn't behave as one would expect as it doesn't expand below.
A gear may be more appropriate, and I don't know if there's common terms to describe what it is. A tooltip might be needed to explain it's mostly a compatibility thing.

client/scenes/gui_common.cpp
```diff
@@ -116,6 +119,84 @@ bool refresh_rate(
...
+
+	if (button_pressed)
+		ImGui::OpenPopup("body part menu");
+
+	const auto & popup_layer = imgui_ctx.layers()[1];
+	const glm::vec2 popup_layer_center = popup_layer.vp_origin + popup_layer.vp_size / 2;
+	ImGui::SetNextWindowPos({popup_layer_center.x, popup_layer_center.y}, ImGuiCond_Always, ImVec2(0.5f, 0.5f));
+	ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, {8, 8});
```
xytovl: The panel felt quite small when I tested, maybe some more (horizontal) padding?

xytovl:
Overall I think it's fine, UI polish could be done in a separate pull request if we want.

Are the virtual trackers meant to remain long term? I don't think they belong to the OpenXR layer and it's mostly a hack to get them to translate to OpenVR.

ImSapphire:
> Are the virtual trackers meant to remain long term? I don't think they belong to the OpenXR layer and it's mostly a hack to get them to translate to OpenVR.

Yeah, it's mostly for compatibility to be removed later. Right now, trying to implement tracker emulation from the skeleton in xrizer would be a bit tough, xrizer doesn't have any kind of configuration or UI that could be extended to allow toggling trackers, and device mixing would be difficult to implement.

client/scenes/gui_common.cpp
ImSapphire: I tried to use a gear or gears icon but couldn't get it to fit in the button, and trying to scale the font size using PushFont would scale down the button as well.. I'm not sure if it's possible to offset and/or scale the font without it changing the button size too?

client/scenes/gui_common.cpp
ImSapphire: changed to use the normal window padding, I was trying to make the X button closer to the edge, but it doesn't seem too easy with ImGui

xytovl:
I don't remember what was left to do.
Going through the code quickly it looks fine, if functionally it's complete I'll just do a quick check on UI as that's where last comments were.

ImSapphire:
> I don't remember what was left to do. Going through the code quickly it looks fine, if functionally it's complete I'll just do a quick check on UI as that's where last comments were.

It should be functionally complete at this point, I didn't test after the rebase though since I didn't have time. I'll test it here soon

client/configuration.cpp
```diff
@@ -308,8 +306,7 @@ void configuration::save()
 	json << ",\"forward_keyboard\":" << std::boolalpha << forward_keyboard;
 	json << ",\"forward_mouse\":" << std::boolalpha << forward_mouse;
 	json << ",\"forward_gamepad\":" << std::boolalpha << forward_gamepad;
-	json << ",\"fb_lower_body\":" << std::boolalpha << fb_lower_body;
-	json << ",\"fb_hip\":" << std::boolalpha << fb_hip;
+	json << ",\"body_part_mask\":" << body_part_mask;
```
xytovl: If we later do a change in the enum values, we will have to migrate the configuration file.
We should store by name if possible.

ImSapphire: hmmm, what would be a clean way to do that? I can't think of anything that wouldn't be extremely messy

ImSapphire: actually I suppose I can use magic_enum to do it automatically for new enum entries. I'll experiment

xytovl: Something like
```json
{
  "body_parts": {
    "chest": true,
    "hip": false, // this may be useful so that we can have defaults that are not all off
    ...
  }
}
```
Both for reading and writing you would iterate over the exising enum values.

ImSapphire: done

PR #844, opened 2026-03-20
Update Monado

ImSapphire:
User-facing fixes and improvements:
- Fix previous layers not being discarded when an empty frame is submitted ([a232d15eecaa3e82959401bcff6f2f354a8098fb](https://gitlab.freedesktop.org/monado/monado/-/commit/a232d15eecaa3e82959401bcff6f2f354a8098fb))
- `LH_STANDBY_ON_EXIT` environment variable to power down lighthouse devices (controllers, trackers) on exit ([16a030f4c628c75b2d19d39c036f343ced9271e2](https://gitlab.freedesktop.org/monado/monado/-/commit/16a030f4c628c75b2d19d39c036f343ced9271e2))


xytovl:
is the interaction profile changed event still correctly changed with the updated patches?

ImSapphire:
Yup, seems to be working fine

xytovl:
It would be nice to have a mechanism to signal WayVR that the no-layer case has been fixed.
If we don't, we'll end up with many people confused as with the cylinder fix.

PR #849, opened 2026-03-23
Add wivrn-version dependency to common-base

Scrumplex:
This causes the packaging in nixpkgs-xr to fail. Not sure why other builds aren't affected.

See https://github.com/nix-community/nixpkgs-xr/pull/669

https://buildbot.nix-community.org/#/builders/4015/builds/442/steps/1/logs/stdio

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -12,6 +12,7 @@ add_library(wivrn-common-base STATIC EXCLUDE_FROM_ALL
     utils/xdg_base_directory.cpp
     ${CMAKE_BINARY_DIR}/version.cpp
 )
+add_dependencies(wivrn-common-base wivrn-version)
 
 add_library(wivrn-common STATIC EXCLUDE_FROM_ALL
     crypto.cpp


xytovl:
We already have the cpp file as a byproduct, it should not be necessary to also have the target itself as a dependency

Scrumplex:
I am not quite sure why this works fine in your CI, but not in our Nix build. Perhaps different cmake versions? We use CMake 4 in nixpkgs.

xytovl:
~~What is the error message you are getting?~~ edit: it's in the first message

Recent changes to version.cpp include: caa191ff06d0945f539d3789acca5f81f30dcf48 and 45907329c7a39a97f38de2b456b58b00d15ac012

xytovl:
Should be fixed by branch `fix/version.cpp`, can you confirm?

We had automoc enabled globally when building dashboard, and it seems it doesn't play nice with generated files.

xytovl:
fb7703cb76c028521c284086078533184375355b merged to master, should fix the build failure

Scrumplex:
I still get an issue with the dependency on version.cpp:

> ```
> wivrn> make[2]: *** No rule to make target 'version.cpp', needed by 'common/CMakeFiles/wivrn-common-base.dir/__/version.cpp.o'.  Stop.
> ```

https://buildbot.nix-community.org/#/builders/4015/builds/447/steps/1/logs/stdio

According to https://cmake.org/cmake/help/latest/command/add_custom_target.html

> By default nothing depends on the custom target. Use the [add_dependencies()](https://cmake.org/cmake/help/latest/command/add_dependencies.html#command:add_dependencies) command to add dependencies to or from other targets.

It does mention some semantic differences between the Ninja and Makefile generators, but nothing that implies that byproducts create automatic dependencies.

xytovl:
I have been able to reproduce the issue, it only happens with high enough parallelism (`-j12` is fine, `-j24` breaks).
It is a cmake bug: https://gitlab.kitware.com/cmake/cmake/-/work_items/19005 with generators other than Ninja.

The pull request should work for the `wivrn-common-base` target, while the `aux_util` target is deep enough in build dependencies that it it unlikely to happen.

PR #858, opened 2026-03-30
Implement main compositor

xytovl:
Replace the patches to Monado's main compositor with a local implementation, reusing code when relevant.
    
The layer squasher is very close to comp_render_cs, while distortion is completely replaced by foveation.

### DIFF
--- a/server/compositor/compositor.cpp
+++ b/server/compositor/compositor.cpp
@@ -730,6 +730,13 @@ void compositor::update_foveation_center_override(const from_headset::override_f
 	foveation.update_foveation_center_override(center);
 }
 
+void compositor::resume()
+{
+	for (auto & encoder: encoders)
+		encoder->reset();
+	send_video_stream_description();
+}
+
 void compositor::on_feedback(const from_headset::feedback & feedback, const clock_offset & o)
 {
 	uint8_t stream = feedback.stream_index;
--- a/server/compositor/compositor.h
+++ b/server/compositor/compositor.h
@@ -168,6 +168,8 @@ public:
 	void update_tracking(const from_headset::tracking &);
 	void update_foveation_center_override(const from_headset::override_foveation_center &);
 
+	void resume();
+
 	void on_feedback(const from_headset::feedback &, const clock_offset &);
 };
 
--- a/server/driver/wivrn_session.cpp
+++ b/server/driver/wivrn_session.cpp
@@ -405,9 +405,9 @@ void wivrn_session::resume_session()
 			        .to = target_fps,
 			});
 		}
-
-		// FIXME: resume compositor
 	}
+
+	compositor.resume();
 
 	if (get_info().user_presence)
 	{


server/compositor/compositor.cpp
```diff
@@ -0,0 +1,747 @@
...
+#ifndef NDEBUG
+	        .initialized = true,
+	        .recursive = false,
+#endif
+	};
+}
+
+const comp_swapchain_image & get_layer_image(const comp_layer & layer, uint32_t swapchain_index, uint32_t image_index)
+{
+	return ((struct comp_swapchain *)(comp_layer_get_swapchain(&layer, swapchain_index)))->images[image_index];
```
ImSapphire: perhaps use reinterpret_cast here, i find it a bit hard to read this
```suggestion
	return reinterpret_cast<struct comp_swapchain *>(comp_layer_get_swapchain(&layer, swapchain_index))->images[image_index];
```

server/compositor/compositor.cpp
```diff
@@ -0,0 +1,747 @@
...
+{
+	try
+	{
+		session.send_control(to_headset::refresh_rate_change{.fps = hz});
+		return XRT_SUCCESS;
+	}
+	catch (std::exception & e)
+	{
+		U_LOG_W("refresh rate change failed: %s", e.what());
+		return XRT_ERROR_IPC_FAILURE;
```
ImSapphire: this may not be a good error code to return, if this error code makes it to oxr (not sure if that's the case) the app will lose the instance

server/compositor/compositor.cpp
```diff
@@ -0,0 +1,747 @@
...
+	        std::get<vk::PhysicalDeviceVulkan12Features>(vk.feat).timelineSemaphore,
+	        true, // in Vulkan 1.2
+	        vk.has_instance_ext(VK_EXT_DEBUG_UTILS_EXTENSION_NAME),
+	        log_level);
+	vk::detail::resultCheck(vk::Result(res), "vk_init_from_given");
+
+	auto xxx = (PFN_vkGetInstanceProcAddr)vk.instance.getProcAddr("vkGetInstanceProcAddr");
+	auto y = xxx(*vk.instance, "vkGetPhysicalDeviceExternalFenceProperties");
+	assert(y);
+
```
ImSapphire: seems unused

server/compositor/layer_squasher.cpp
```diff
@@ -0,0 +1,846 @@
...
+				m_relation_chain_resolve(&xrc, &rel);
+				world_poses[i] = rel.pose;
+			}
+		}
+	}
+};
+
+const comp_swapchain_image & get_layer_image(const comp_layer & layer, uint32_t swapchain_index, uint32_t image_index)
+{
+	return ((struct comp_swapchain *)(comp_layer_get_swapchain(&layer, swapchain_index)))->images[image_index];
```
ImSapphire:
```suggestion
	return reinterpret_cast<struct comp_swapchain *>(comp_layer_get_swapchain(&layer, swapchain_index))->images[image_index];
```

server/compositor/layer_squasher.cpp
```diff
@@ -0,0 +1,846 @@
...
+}
+
+const struct comp_swapchain_image &
+get_layer_depth_image(const comp_layer & layer, uint32_t swapchain_index, uint32_t image_index)
+{
+	return ((struct comp_swapchain *)(comp_layer_get_depth_swapchain(&layer, swapchain_index)))->images[image_index];
```
ImSapphire:
```suggestion
	return reinterpret_cast<struct comp_swapchain *>(comp_layer_get_depth_swapchain(&layer, swapchain_index))->images[image_index];
```

server/compositor/compositor.cpp
xytovl: ops, old debugging code

server/compositor/compositor.cpp
xytovl: Changed to ignore the error, according to https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_FB_display_refresh_rate
> Note that this is only a request and does not guarantee the system will switch to the requested display refresh rate.

The error should only happen if the application requests a change while the headset is not connected.

PR #862, opened 2026-04-03
add option for half fps

galister:


client/scenes/stream.cpp
```diff
@@ -232,7 +232,7 @@ std::shared_ptr<scenes::stream> scenes::stream::create(std::unique_ptr<wivrn_ses
 		if (self->instance.has_extension(XR_FB_DISPLAY_REFRESH_RATE_EXTENSION_NAME))
 		{
 			info.available_refresh_rates = self->session.get_refresh_rates();
-			info.settings.preferred_refresh_rate = config.preferred_refresh_rate;
+			info.settings.preferred_refresh_rate = config.preferred_refresh_rate / (float)config.fps_divider;
```
xytovl: We may have to split refresh rate and target frame rate.

resume_session is sending a display refresh rate changed event that may be incorrect.

CherrieTheShifter:
yes this

galister:
now sending both preferred_refresh_rate and fps_divider as is, and it gets evaluated on the server only.

server/driver/wivrn_session.cpp
```diff
@@ -755,13 +772,16 @@ void wivrn_session::operator()(from_headset::user_presence_changed && event)
...
 	push_event(
 	        {
 	                .display = {
 	                        .type = XRT_SESSION_EVENT_DISPLAY_REFRESH_RATE_CHANGE,
-	                        .from_display_refresh_rate_hz = event.from,
-	                        .to_display_refresh_rate_hz = event.to,
+	                        .from_display_refresh_rate_hz = old_effective,
+	                        .to_display_refresh_rate_hz = new_effective,
 	                },
 	        });
```
xytovl, line 771: I don't think we should lie to the application about the refresh rate.
The value should be the actual display refresh rate. Divider shall be used for pacing only

galister: Makes sense, adjusted.

galister:
anything else i can polish on this one?

xytovl:
I've just tested it and it seems to work fine, the "(native)" text may be too much, the "(xx with spacewarp)" on halved ones should be sufficient.

galister:
removed Native from the labels.

xytovl:
There's still a logic issue due to the previous assumption that frame rate == refresh rate.

We need to introduce the divider in the compositor, so that we send actual refresh rate to the extension, but keep frame rates for timing and pacer

xytovl:
I've done some extensive changes, @galister can you give it a quick review?

galister:
looks good!

xytovl:
actually merged, I rebased and pushed through cli

PR #863, opened 2026-04-03
server: wait for headset before sending frames

caszuu:
Make sure we have a valid clock offset (and therefore view poses) before starting to encode frames.

This is really only visible when reconnecting, where the server is able to start streaming immediately before tracking had a chance to react.


CherrieTheShifter:
peak

PR #889, opened 2026-04-26
fix(encoder): match passthrough encoder dimensions to eyes on nvenc

damaestro:
Work around an NVIDIA driver bug that requires the passthrough/alpha encoder to share dimensions with the eye encoders. Keep the half-height alpha layout for non-nvenc backends.

Having the third encoder stream a different size has some very negative effects on single engine generation cards. This appears to help a lot on my GTX 3080ti.

### DIFF
--- a/server/encoder/encoder_settings.cpp
+++ b/server/encoder/encoder_settings.cpp
@@ -293,15 +293,17 @@ std::array<encoder_settings, 3> get_encoder_settings(wivrn::vk_bundle & bundle, 
 
 	auto width = align(info.stream_eye_width, 64);
 	auto height = align(info.stream_eye_height, 64);
-	// Ensure we don't try to encode too large images (only for left/right, ignore alpha)
-	for (size_t i = 0; i < 2; ++i)
-		check_video_size(res[i].encoder_name, res[i].codec, width, height);
+	// Ensure we don't try to encode too large images
+	for (auto & encoder: res)
+		check_video_size(encoder.encoder_name, encoder.codec, width, height);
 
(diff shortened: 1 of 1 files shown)


xytovl:
If it is related to the number of encoder engines, then it should depend on `NV_ENC_CAPS_NUM_ENCODER_ENGINES`.

According to https://developer.nvidia.com/video-encode-decode-support-matrix there's quite a few cards with 2 or more engines in 40 and 50 series.

damaestro:
Due to reports with people in the 4x series having this issue, now I'm not certain this is related to number of engines. The different size encoding must cause some terrible code path following in the driver.

`NV_ENC_CAPS_NUM_ENCODER_ENGINES` would be nice to use to adjust our frame submission strategy. Right now it appears as if we submit in serial. For cards with more than one engine we might be able to take advantage and submit in parallel.

In addition to this tuning, we should also make it so we can completely disable the alpha encoding channel. Any encoder context still has some sync overhead and for those of us that wont be using this feature, it could be a small performance win.

xytovl:
Parallel encoding would be easier on vulkan, where we could simply use different vulkan encode queues.
Doing so in nvenc would require a rather significant rework on the encoder interface and synchronization.

I don't know how the existence of alpha encoder could introduce a performance penalty, we are not calling it when passthrough is not enabled. You can try and disable it so you can test if that's indeed an issue.

damaestro:
The encoder is created on the card and has to be maintained. Even when not doing work there is overhead. I have not measured how much overhead. I'll see about getting some data when I find more time.

xytovl:
This doesn't seem to be necessary, closing

PR #895, opened 2026-05-03
Compositor: defer GPU wait/query/GC to encoder thread

ZokuTe:
This PR addresses two major performance regressions introduced in the 22a819b8 architecture.
---
Previously, `compositor::layer_commit` performed synchronous waits on the Vulkan timeline semaphore and ran swapchain garbage collection while on Monado’s `multi_main_loop` system thread. 
- Issue: This created a **3–5 ms stall** per layer commit, preventing the `comp_multi` scheduled slot from draining. This resulted in "Frame late" and "GPU work waiting to be displayed" warnings.
- Fix: Moved these blocking operations to the **encoder thread**. `layer_commit` now returns immediately after publishing the encode request, restoring microsecond-level latency to the system thread.
---
The rewritten architecture stopped feeding `SUBMIT_END` timestamps to `pacer::mark_timing_point`.
- Issue:*Without feedback, `mean_wake_up_to_present_ns` remained stuck at its 1ms default. This caused `predict_frame` to calculate inaccurate wake-up times, consistently waking the application too late.
- Fix: Re-implemented the timing feedback path, allowing the pacer to converge on the real GPU latency and ensuring accurate application wake-up predictions.
---
- Indexed Query Pool: Expanded the query pool and implemented per-image slot indexing (offset `i * 3`) to prevent frame N+1 from overwriting frame N’s timestamps during asynchronous processing.
- Deferred Wait: In-flight images now carry a `sem_value` so the encoder thread can wait on the specific timeline value before recycling the image.

### DIFF
--- a/server/compositor/compositor.cpp
+++ b/server/compositor/compositor.cpp
@@ -323,8 +323,8 @@ xrt_result_t compositor::layer_commit(xrt_graphics_sync_handle_t sync_handle)
 	cmd_pool.reset();
 	cmd.begin({.flags = vk::CommandBufferUsageFlagBits::eOneTimeSubmit});
 
-	cmd.resetQueryPool(*query_pool, 0, 3);
-	cmd.writeTimestamp(vk::PipelineStageFlagBits::eTopOfPipe, *query_pool, 0);
+	cmd.resetQueryPool(*query_pool, i * 3, 3);
+	cmd.writeTimestamp(vk::PipelineStageFlagBits::eTopOfPipe, *query_pool, i * 3 + 0);
 
 	bool flip_y = false;
 	std::array<vk::ImageView, 2> src;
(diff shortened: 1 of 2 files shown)


xytovl:
layer_commit should remain synchronous, like the Monado implementation.
There is a single command buffer, ubo and squasher/foveation data. We only wait for completion of the compositor part (where sem_info is signlaed), so the encoder part (present and encode) can run asynchronously and they have their own private data.

3-5ms execution time for squasher + foveation is too high, can you check with the debug gui what squasher and foveation GPU time look like?

The `SUBMIT_END` fix might be useful, although I noticed that it converges to very low values (under 1ms), so it didn't make any noticeable difference.

ZokuTe:
Although they are useful in theory, in-depth testing has proven that these fixes are completely useless.

PR #896, opened 2026-05-03
Support for Play for Dream MR

### LINKED ISSUE #465: [Bug]: Play for Dream headset support

amwatson:
- Add compositor layer flag to passthrough to prevent blackscreen on PFD (technically required but not enforced on other runtimes)
- Fix #465 : I think it's a bug/spec violation in Play for Dream's compositor: OpenXR spec says imageRect.offset should use top-left origin for Vulkan (spec as of 2021 is like ["the origin for ImageRec should match your graphics API"](https://registry.khronos.org/OpenXR/specs/1.1/man/html/XrSwapchainSubImage.html) and the [relevant graphics API for Vulkan specifies top-left](https://registry.khronos.org/OpenXR/specs/1.0/man/html/XrSwapchainImageVulkan2KHR.html), relevant discussion is https://github.com/KhronosGroup/OpenXR-Docs/issues/86). However, PFD's runtime uses bottom-left origin for both OpenGL and Vulkan. As a result it's sampling the wrong part of the texture atlas.
    - Note on enum naming: the company is called "Play for Dream" and the HMD model is "MR", thus, "Play for Dream MR"
- Make controllers identifiable by skipping over `grip_surface/pose` despite what PFD's OpenXR API version says.


amwatson:
[e2275e6](https://github.com/WiVRn/WiVRn/pull/896/commits/e2275e6c9472b4c438e2ce3ba3937753247d0a1d) is one of multiple ways to solve a problem -- maintainers can suggest an alternative way if preferred.

**Problem**: Play for Dream's runtime identifies itself as "oculus 1.0.0" (fascinating) but returns API version 1.1.54 from `xrGetInstanceProperties`. Per the OpenXR spec, a runtime advertising API version >= 1.1 is required to include `grip_surface/pose` in the allowlist of every interaction profile that supports grip/pose. PFD's runtime does not: `grip_surface/pose` is absent from those allowlists despite negotiating API 1.1.54.

The current code injects `grip_surface/pose` into interaction profiles when api_version >= 1.1.0, which is valid per spec. Because of PFD's non-conformance, `xrSuggestInteractionProfileBindings` returns XR_ERROR_PATH_UNSUPPORTED for the whole binding call, causing all pose bindings to be discarded. The result is that `xrLocateSpace` returns no valid location flags for either controller, and controller tracking is completely broken.

**Solution**: Probe for `grip_surface/pose` support explicitly before injecting it, and only inject if the probe succeeds. 

This avoids hardcoding any assumption about PFD's runtime specifically. The probe adds some complexity for all devices -- which is only needed for PFD MR, due to spec non-conformance -- on init, which is the tradeoff.

Alternative solutions include just skipping injection for Play for Dream MR or downgrading Play for Dream MR to OpenXR 1.0 regardless of what the negotiated version says. These add less code but hard-code in the assumption that Play for Dream MR incorrectly negotiates API 1.1 and doesn't support `grip_surface/pose`.

ImSapphire:
i'm of the opinion that it should just be quirked for PFD specifically. perhaps we should not even try to request openxr 1.1 on PFD if it's broken

ImSapphire:
is there any way to report these spec violations to PFD? these are pretty egregious and should be fixed in the runtime

amwatson:
> is there any way to report these spec violations to PFD? these are pretty egregious and should be fixed in the runtime

I was wondering that. I'll ask the evangelist for the right contact. They have a "support" github, but idk

xytovl:
My main concern is that we are spreading the `model` enum in too many places. Initially this was only used to adjust the resolution, as the "recommended" value is not the most suitable for streaming, then many things were added.

What I would do instead is to have a `quirks` structure which holds individual quirks (resolution override, OpenXR 1.1 disable, sRGB etc) and permission data. The constructor would have settings based on the model and possibly environment variables to set a specific value so it can be tested without the matching hardware.
Functionally it would remain the same, but we won't need to extend the model enum.

There's 3 options to do this:
* I can write the structure on master, then you could rebase your work
* I can add a commit to the pull request
* You do it on the pull request

amwatson:
Sure, happy to

client/xr/passthrough.cpp
```diff
@@ -60,7 +60,7 @@ xr::passthrough_fb::passthrough_fb(instance & inst, session & s) :
 
 	composition_layer = XrCompositionLayerPassthroughFB{
 	        .type = XR_TYPE_COMPOSITION_LAYER_PASSTHROUGH_FB,
-	        .flags = 0,
+	        .flags = XR_COMPOSITION_LAYER_BLEND_TEXTURE_SOURCE_ALPHA_BIT,
```
xytovl, line 63: Is PFD using the FB passthrough extension?

The passthrough layer should not use alpha.

client/application.cpp
```diff
@@ -565,6 +600,52 @@ static std::vector<interaction_profile> interaction_profiles{
         },
 };
 
+// Needed for Play for Dream MR spec violation.
+// Explanation: Per OpenXR spec, API versions 1.1+ must support grip_surface/pose if they support
+// grip/pose. Play for Dream MR currently negotiates an API version above 1.1, but doesn't support
+// grip_surface/pose, so we need to probe for it explicitly. This will likely change when
+// Play for Dream updates its runtime to 1.1.
```
xytovl, line 607: If you force OpenXR 1.0, does the palm_pose extension work?

amwatson:
@xytovl 

> My main concern is that we are spreading the `model` enum in too many places. Initially this was only used to adjust the resolution, as the "recommended" value is not the most suitable for streaming, then many things were added.
> 
> What I would do instead is to have a `quirks` structure which holds individual quirks (resolution override, OpenXR 1.1 disable, sRGB etc) and permission data. The constructor would have settings based on the model and possibly environment variables to set a specific value so it can be tested without the matching hardware. Functionally it would remain the same, but we won't need to extend the model enum.
> 
> There's 3 options to do this:
> 
> * I can write the structure on master, then you could rebase your work
> * I can add a commit to the pull request
> * You do it on the pull request

#899

client/xr/passthrough.cpp
amwatson: > Is PFD using the FB passthrough extension?

 PFD can use either the `XrCompositionLayerPassthroughFB` or [`XR_YVR_passthrough`](https://developer.yvrdream.com/yvrdoc/native_EN/UserManual/Archive/MobileSDK/AdvancedDevelopment.html?q=passthrough)
 
 > The passthrough layer should not use alpha.
 
Just to make sure we're on the same page: are you trying to do something other than the [default passthrough layer behavior on Meta Quest](https://developers.meta.com/horizon/documentation/native/android/mobile-passthrough/#submit-passthrough-layers-to-the-compositor)?
> The passthrough layers generated by the system may contain non-opaque alpha values when styles or surface-projected Passthrough are used. Those alpha values are ignored if XR_COMPOSITION_LAYER_BLEND_TEXTURE_SOURCE_ALPHA_BIT is not set. It is recommended to set the flag unless an application specifically needs to ignore alpha values

client/application.cpp
amwatson: Yeah, it works fine. PFD should prooobably Just Work if you request 1.0.0 as the `apiVersion` during `xrCreateInstance()` and then ensure you're using the 1.0 version of the API at every call, if that's your preference/what you're already doing for other devices anyway.

client/xr/passthrough.cpp
xytovl: We have support for `XR_FB_passthrough` as the Meta runtime initially didn't support the standard OpenXR `XR_ENVIRONMENT_BLEND_MODE_ALPHA_BLEND`.

We use the `XrCompositionLayerPassthroughFB` in order to emulate environment alpha blend, so it is the bottommost layer and there should not be a difference with or without the alpha flag.

If `XR_ENVIRONMENT_BLEND_MODE_ALPHA_BLEND` is available, it should be preferred as we don't use any of the odd features one can have with the layer.

client/application.cpp
xytovl: Forcing OpenXR 1.0 is already done for HMD headsets, I think it's better that way

client/xr/passthrough.cpp
amwatson: The [available flags](https://registry.khronos.org/OpenXR/specs/1.0/man/html/XrCompositionLayerFlagBits.html) are:
- `XR_COMPOSITION_LAYER_UNPREMULTIPLIED_ALPHA_BIT`
- `XR_COMPOSITION_LAYER_BLEND_TEXTURE_SOURCE_ALPHA_BIT`
- `XR_COMPOSITION_LAYER_CORRECT_CHROMATIC_ABERRATION_BIT`

For clarity: we need to set *something* in this field ([XrCompositionLayerPassthroughFB](https://registry.khronos.org/OpenXR/specs/1.0/man/html/XrCompositionLayerPassthroughFB.html) says "flags must not be 0", and PFD's runtime, unlike Quest's, actually enforces this). Note that the requirement and the interpretation of these flags on passthrough layers differ from other layer types.

If not `XR_COMPOSITION_LAYER_BLEND_TEXTURE_SOURCE_ALPHA_BIT`, we can alternatively set a nonsense flag like `XR_COMPOSITION_LAYER_UNPREMULTIPLIED_ALPHA_BIT` or `XR_COMPOSITION_LAYER_CORRECT_CHROMATIC_ABERRATION_BIT`, which are guaranteed [no-ops on Quest specifically](https://developers.meta.com/horizon/documentation/native/android/mobile-passthrough/). 

Or, if you don't like this idea, I can just leave passthrough unchanged for Quest and instead use [XR_YVR_passthrough](https://developer.yvrdream.com/yvrdoc/native_EN/UserManual/Archive/MobileSDK/AdvancedDevelopment.html?q=passthrough) on Play for Dream's passthrough, which is pretty straightforward.

xytovl: You're right, let's keep `XR_COMPOSITION_LAYER_BLEND_TEXTURE_SOURCE_ALPHA_BIT`.

The spec still doesn't make much sense, but that's common with FB extensions.

amwatson:
> is there any way to report these spec violations to PFD? these are pretty egregious and should be fixed in the runtime

@ImSapphire I've got a thread with the devs on Discord

amwatson:
Got in touch with the Play for Dream dev team and they are fixing both issues by "the end of the month"

> Hi guys, we will fix these two issues in version 4.5.0 at the end of this month. Thanks again.

Will have to do a messy merge to continue dev but should have time this weekend or next week

EasonZxp:
I verified this on the upcoming PFDM 4.5.0 release. The controller support works correctly there, and I could not reproduce the Vulkan origin issue mentioned above.

Since #896 is fairly old now, I opened a separate PR for the controller support changes. #933

amwatson:
Closing as the important changes were instead resolved within the runtime. Passthrough flag might be good to resolve as a separate PR

PR #897, opened 2026-05-04
Updated 3 new translations in pt_BR

Patola:
maximum/minimum translations. Fixes #446


ThomasEricB:
thank you for the translations @Patola

PR #899, opened 2026-05-04
Add runtime-initialized 'hmd_traits' struct to streamline hmd-specific behavior checks

amwatson:
Previously, HMD-specific behavior (including workarounds for per-device runtime bugs) was scattered across in-function switch statements on `guess_model()`. This revision consolidates all HMD-specific behavior into a single `hmd_traits` struct that is initialized once at client startup.

## Design choices: 
- Explicit over implicit accesses: e.g. `runtime_hmd_traits().needs_srgb_conversion` instead of a global `needs_srgb_conversion()` function so performance and dependencies are clear at the caller level (with two important exceptions: fallback lazy-init in `runtime_hmd_traits()` and implicit use of `guess_model()` inside `initialize_runtime_hmd_traits()`).
- Global `runtime_hmd_traits` with one-time-init at top of `Application::Application()`: Populate all properties once during app init so init cost (e.g. model sysprop reads) is realized only once and never in the frame loop.   
- Sysprop/envvar overrides of traits at runtime: 
    - Sysprop (Android) and environment variable (linux) overrides can be used to simulate different hardware traits without access to that hardware. 
    - Checking once on init means you can't change the traits dynamically once the app is launched but 1) that wouldn't work well 2) sysprop reads are expensive enough that adding them to a hotpath can add lots of cycles no one wants.

## Testing
**Available Devices:** Oculus Quest, Meta Quest 2, Meta Quest 3, Lynx R1, Oculus Quest Pro, Play for Dream MR, Pico 4, Vive Focus 3  
**Missing Devices:** Samsung Galaxy XR, Pico Neo 3, Pico 4 Pro, Pico 4 Enterprise, Pico 4s, Vive XR Elite, Vive Focus Vision
**Devices Tested:** Meta Quest 3, Play for Dream MR, Pico 4
 
| Test | Steps | Expected |
|---|---|---|
| Trait correctness | Launch client, inspect logcat | `HMD traits:` log line matches output from debug log added to old version |
| Sysprop overrides | Set override sysprops, launch client, inspect logcat | `HMD trait overrides:` log line reflects override values |
| Permission prompts | Uninstall, install, launch local client | User is prompted with correct permissions requests for device |

Note: Since I'm not testing all supported devices, I additionally ran a homemade "static analysis" script to compare old/new codes versions to verify per-traits amount to the same thing across both versions

Test script for overrides [here](https://gist.github.com/amwatson/8f5efc4a15a9d8df93b5162edc255082)
Test script for devices configurations (assumes debug sysprops for devices I didn't add to the revision) [here](https://gist.github.com/amwatson/05b0e6202c7a616769d985f5676a2e35)

### DIFF
--- a/client/hardware.cpp
+++ b/client/hardware.cpp
@@ -25,7 +25,6 @@
 #include <cstring>
 #include <glm/ext/quaternion_trigonometric.hpp>
 #include <limits>
-#include <stdexcept>
 #include <string>
 
 #include <spdlog/spdlog.h>
@@ -35,12 +34,13 @@
 
 #include <sys/system_properties.h>
 
-static std::string get_property(const char * property)
+static std::optional<std::string> get_property(const char * property)
 {
 	auto info = __system_property_find(property);
-	std::string result;
 	if (not info)
-		return result;
+		return std::nullopt;
+
+	std::string result;
 	wrap_lambda cb = [&result](const char * name, const char * value, uint32_t serial) {
 		result = value;
 	};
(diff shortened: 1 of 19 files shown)


client/hardware.h
```diff
@@ -54,16 +55,34 @@ enum class feature
...
+std::string model_name();
+void initialize_runtime_hmd_traits();
+// Initialized once at startup (initialize_runtime_hmd_traits()).
+// If the HMD is recognized, returns traits specific to that model; otherwise,returns
+// default behavior.
+// Fallback: if called before initialize_runtime_hmd_traits() (avoid this!), dynamically
+// constructs itself and complains in logs.
+const hmd_traits & runtime_hmd_traits();
+const char * permission_name_for_hmd(const hmd_traits & traits, const feature f);
+XrViewConfigurationView override_view_for_hmd(const hmd_traits & traits, XrViewConfigurationView);
```
ImSapphire: I think this should be a class that uses our `singleton` utility, which exposes the fields via const accessor methods, instead of having these global functions.

ImSapphire (changes requested):

client/hardware.h
ImSapphire: It could also just be a class stored inside `application`, with a `const hmd_traits & application::get_hmd_traits()` accessor.

client/hardware.h
```diff
@@ -54,16 +55,34 @@ enum class feature
...
 
-bool need_srgb_conversion(model);
+struct hmd_traits
+{
+	const char * controller_profile = "generic-trigger-squeeze";
+	const char * controller_ray_model = "assets://ray.glb";
+	XrVersion max_openxr_api_version = XR_API_VERSION_1_1;
+	uint32_t panel_width_override = 0;
+	bool needs_srgb_conversion = true;
+	const hmd_permissions * permissions = nullptr;
```
xytovl: This could be a `magic_enum::containers::array<feature, const char*>`

client/hardware.h
xytovl: global variable with static initialization may be easier as long as we don't access it in other static initalized contexts.

As we control the whole application, there is no need for the fallback initialization with logs. If we don't want it, just add an assertion at start that it is initialized.

xytovl: Those could be const methods

client/hardware.cpp
```diff
@@ -122,222 +125,246 @@ model guess_model()
...
-	throw std::range_error("invalid model " + std::to_string((int)m));
+#ifdef __ANDROID__
+	const std::string value = get_property(android_sysprop_name);
+	if (!value.empty())
+		return value;
+	return std::string(fallback ? fallback : "");
+#else
+	const char * value = std::getenv(env_name);
+	if (value != nullptr && *value != '\0')
+		return std::string(value);
```
xytovl: If the environment variable is the empty string, I would return that instead of default value.

client/hardware.cpp
```diff
@@ -122,222 +125,246 @@ model guess_model()
...
+{
+	if (g_hmd_traits.has_value())
+		return;
+
+	hmd_traits q = get_hmd_traits(guess_model());
+	const uint32_t panel_width = env_u32("WIVRN_QUIRK_PANEL_WIDTH", "debug.wivrn.quirk_panel_width");
+	const bool had_panel_override = panel_width > 0;
+	if (had_panel_override)
+		q.panel_width_override = panel_width;
+	const bool disable_openxr_1_1 = env_bool("WIVRN_QUIRK_DISABLE_OPENXR_1_1", "debug.wivrn.quirk_disable_openxr_1_1", q.max_openxr_api_version < XR_API_VERSION_1_1);
```
xytovl: It looks like the `env_` functions could return `std::optional` so we don't need magic values.

client/hardware.h
```diff
@@ -54,16 +55,34 @@ enum class feature
...
+};
 
-// Return nullptr if no permission is required
-const char * permission_name(feature f);
+model guess_model();
```
xytovl: The `model` enum should no longer be needed outside of the cpp file.
Even in the implementation it may be easier to initialize the traits without the intermediate enum.

client/hardware.h
amwatson: > As we control the whole application, there is no need for the fallback initialization with logs

Sure, will remove, thanks.

> global variable with static initialization

Unfortunately, you can't static-initialize `hmd_traits`, as it depends on variables known only at runtime (e.g. model)

amwatson: You mean inside the class? Like const methods inside the `hmd_traits` struct?

client/hardware.cpp
amwatson: Yeah, good callout. Will do

client/hardware.cpp
amwatson: Sure, thanks!

client/hardware.h
xytovl: methods inside hmd_traits, so you call
```cpp
runtime_hmd_traits().override_view(view);
```

For permissions, the array could be a member variable so you access it through
```cpp
runtime_hmd_traits().permissions[f];
```

xytovl: Sorry I was unclear, I meant a global variable that is initialized as part of program startup. Just to skip the parens when accessing it.

client/hardware.h
xytovl: I have pushed to https://github.com/WiVRn/WiVRn/tree/refactor/quirks how I see it (untested for the moment).

client/hardware.h
xytovl: It turns out we can't cleanly use the logger with a global variable, I'll revert to the accessor + explicit initialization.

xytovl:
@amwatson if you want to finish this pull request, you can integrate my changes on https://github.com/WiVRn/WiVRn/tree/refactor/quirks or I can just merge that one as I think it's good enough (and includes all from the current PR).

amwatson:
I have a draft for the refactor you requested (adds hmd_traits as a member of application). Heads-up: eliminating the hardware file is a little extensive. If you want, I'll make that a separate PR if you want to merge this one first.

Otherwise, I can test and add the refactor tomorrow -- either a full or partial (less extensive) elimination of guess_model()

xytovl:
Some changes I did on the branch that should be replicated:
* e8a532c1a0fda6901541921bffeb60c0d1ad1a86 to fully remove the model enum
* 57f81f2192207a1db436b143f927bf0438b54e4a to workaround a bug on Pico
* b741fd19cac781553f2f5bf651d58eccf592fcb2 so that overrides are done automatically when a new member is added (+ d5fe2eb262847feb1b54546f0a48489f0655542d for CI as it needs a newer boost version)

amwatson:
Cool, my internet is spotty, but will do,and should be good for EoD. Thanks!

amwatson:
Moved hmd_traits into application class and removed hardware.h/cpp. 

Taste-wise, `model_name` could be a static in `wivrn_client.cpp` since it's only called in one place -- the reason I put it in `hmd_traits` is so we don't add android-specific behaviors to the client code.

xytovl:
Looks fine to me, just get clang-format happy and we can merge. Thanks for the PRs.

amwatson:
@xytovl oops, thanks. Should be good now!

ImSapphire (approved):

PR #913, opened 2026-05-16
feat: add OVR_COMPAT_EXTRA_SEARCH_PATH for distro-prepended compat paths

damaestro:
Add an extra option to allow downstream packaging to append search paths for OVR compatibility paths. This allows packagers to add distribution paths to the search path while keeping upstream standard search paths.


xytovl:
I am not sure this is a good idea, the default values are not very important. They mainly help those who build from sources.

If you think distros shouldn't override the defaults and only append, then I would handle that in the build system: concatenate both lists before the configure_file. In C++ the only change would be to skip empty entries

damaestro:
Yeah, makes sense. I put this up for discussion.

PR #914, opened 2026-05-16
Fix vulkan encode QFOT layouts (NVIDIA green frame)

damaestro:
The compositor's release barrier left old/newLayout defaulted to eUndefined while the encoder's acquire used eGeneral -> eVideoEncodeSrcKHR (or eTransferSrcOptimal on the NVIDIA tmp_image path), which the spec forbids for an exclusive-mode queue-family ownership transfer. NVIDIA honoured the mismatch by treating the encode-queue contents as undefined (zeros), giving Y=0/CbCr=0 -- solid green.

Also fix optimal_transfer(): the inner `auto to = ...` shadowed the parameter, so the maintenance9 check returned mask & (1 << mask) instead of mask & (1 << dst_qfi) and was effectively always false.

Release/acquire now agree on eGeneral -> eGeneral; the encoder issues a second non-QFOT barrier to reach eVideoEncodeSrcKHR before encodeVideoKHR.

### DIFF
--- a/server/compositor/compositor.cpp
+++ b/server/compositor/compositor.cpp
@@ -442,6 +442,14 @@ xrt_result_t compositor::layer_commit(xrt_graphics_sync_handle_t sync_handle)
 			                .srcAccessMask = vk::AccessFlagBits2::eShaderStorageWrite,
 			                .dstStageMask = vk::PipelineStageFlagBits2::eAllCommands,
 			                .dstAccessMask = vk::AccessFlagBits2::eMemoryRead,
+			                // For a queue-family ownership transfer in EXCLUSIVE
+			                // sharing mode, the release barrier's old/new layout
+			                // must match the encoder-side acquire. Both sides use
+			                // eGeneral → eGeneral here; the encoder issues a
+			                // separate non-QFOT barrier to transition to its
+			                // required layout (eVideoEncodeSrcKHR, etc.).
+			                .oldLayout = vk::ImageLayout::eGeneral,
+			                .newLayout = vk::ImageLayout::eGeneral,
 			                .srcQueueFamilyIndex = vk.queue_family_index,
 			                .dstQueueFamilyIndex = encoder->target_queue,
 			                .image = images[i].image,
--- a/server/compositor/compositor.cpp
+++ b/server/compositor/compositor.cpp
@@ -442,6 +442,16 @@ xrt_result_t compositor::layer_commit(xrt_graphics_sync_handle_t sync_handle)
 			                .srcAccessMask = vk::AccessFlagBits2::eShaderStorageWrite,
 			                .dstStageMask = vk::PipelineStageFlagBits2::eAllCommands,
 			                .dstAccessMask = vk::AccessFlagBits2::eMemoryRead,
+			                // For a queue-family ownership transfer in EXCLUSIVE
+			                // sharing mode, the release barrier's old/new layout
+			                // must match the encoder-side acquire. newLayout is
+			                // the encoder's target_layout; per the VkImageMemoryBarrier2
+			                // spec the layout transition is executed exactly once
+			                // between the queues, so this single QFOT barrier covers
+			                // both the queue-family transfer and the layout
+			                // transition the encoder needs.
+			                .oldLayout = vk::ImageLayout::eGeneral,
+			                .newLayout = encoder->target_layout,
 			                .srcQueueFamilyIndex = vk.queue_family_index,
 			                .dstQueueFamilyIndex = encoder->target_queue,
 			                .image = images[i].image,
--- a/server/encoder/video_encoder_vulkan.cpp
+++ b/server/encoder/video_encoder_vulkan.cpp
@@ -638,14 +638,18 @@ void wivrn::video_encoder_vulkan::present_image(vk::Image y_cbcr, vk::SemaphoreS
 		compositor_sem.stageMask = vk::PipelineStageFlagBits2::eCopy;
 
 		beman::inplace_vector::inplace_vector<vk::ImageMemoryBarrier2, 2> im_barriers;
+		// y_cbcr was written by the compositor in eGeneral and stays in
+		// eGeneral here. eGeneral is a valid source layout for copyImage,
+		// so no layout transition is needed; the QFOT acquire below uses
+		// eGeneral → eGeneral to match the compositor's release.
 		vk::ImageLayout y_cbcr_layout{vk::ImageLayout::eGeneral};
 		if (need_transfer)
 		{
 			im_barriers.push_back(vk::ImageMemoryBarrier2{
 			        .dstStageMask = vk::PipelineStageFlagBits2KHR::eCopy,
 			        .dstAccessMask = vk::AccessFlagBits2::eTransferRead,
-			        .oldLayout = y_cbcr_layout,
-			        .newLayout = vk::ImageLayout::eTransferSrcOptimal,
+			        .oldLayout = vk::ImageLayout::eGeneral,
+			        .newLayout = vk::ImageLayout::eGeneral,
 			        .srcQueueFamilyIndex = vk.queue_family_index,
 			        .dstQueueFamilyIndex = target_queue,
 			        .image = y_cbcr,
@@ -655,7 +659,6 @@ void wivrn::video_encoder_vulkan::present_image(vk::Image y_cbcr, vk::SemaphoreS
 			                             .baseArrayLayer = stream_idx,
 			                             .layerCount = 1},
 			});
-			y_cbcr_layout = vk::ImageLayout::eTransferSrcOptimal;
 		}
 		im_barriers.push_back(vk::ImageMemoryBarrier2{
 		        .dstStageMask = vk::PipelineStageFlagBits2KHR::eCopy,
@@ -739,13 +742,15 @@ void wivrn::video_encoder_vulkan::present_image(vk::Image y_cbcr, vk::SemaphoreS
 	else
 	{
 		compositor_sem.stageMask = vk::PipelineStageFlagBits2::eVideoEncodeKHR;
+		// QFOT acquire (must match compositor's eGeneral → eGeneral release)
+		// followed by a non-QFOT layout transition to eVideoEncodeSrcKHR,
+		// which vkCmdEncodeVideoKHR requires.
+		beman::inplace_vector::inplace_vector<vk::ImageMemoryBarrier2, 2> im_barriers;
 		if (need_transfer)
 		{
-			vk::ImageMemoryBarrier2 video_barrier{
-			        .dstStageMask = vk::PipelineStageFlagBits2KHR::eVideoEncodeKHR,
-			        .dstAccessMask = vk::AccessFlagBits2::eVideoEncodeReadKHR,
+			im_barriers.push_back(vk::ImageMemoryBarrier2{
 			        .oldLayout = vk::ImageLayout::eGeneral,
-			        .newLayout = vk::ImageLayout::eVideoEncodeSrcKHR,
+			        .newLayout = vk::ImageLayout::eGeneral,
 			        .srcQueueFamilyIndex = vk.queue_family_index,
 			        .dstQueueFamilyIndex = target_queue,
 			        .image = y_cbcr,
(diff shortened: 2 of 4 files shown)


damaestro:
There are still performance differences between the `nvenc` and `vulkan` backends, but this seems to be similar to existing differences with what we had before.

server/compositor/compositor.cpp
```diff
@@ -442,6 +442,14 @@ xrt_result_t compositor::layer_commit(xrt_graphics_sync_handle_t sync_handle)
...
 			                .dstStageMask = vk::PipelineStageFlagBits2::eAllCommands,
 			                .dstAccessMask = vk::AccessFlagBits2::eMemoryRead,
+			                // For a queue-family ownership transfer in EXCLUSIVE
+			                // sharing mode, the release barrier's old/new layout
+			                // must match the encoder-side acquire. Both sides use
+			                // eGeneral → eGeneral here; the encoder issues a
+			                // separate non-QFOT barrier to transition to its
+			                // required layout (eVideoEncodeSrcKHR, etc.).
+			                .oldLayout = vk::ImageLayout::eGeneral,
+			                .newLayout = vk::ImageLayout::eGeneral,
```
xytovl: This looks like a bug in the driver: https://docs.vulkan.org/refpages/latest/refpages/source/VkImageMemoryBarrier2.html#_description
> If oldLayout is not equal to newLayout, then the memory barrier defines an [image layout transition](https://docs.vulkan.org/spec/latest/chapters/synchronization.html#synchronization-image-layout-transitions) for the specified image subresource range. If this memory barrier defines a [queue family ownership transfer operation](https://docs.vulkan.org/spec/latest/chapters/synchronization.html#synchronization-queue-transfers), the layout transition is only executed once between the queues.

Leaving them both undefined should be the same as setting them to the actual image layout.

server/encoder/video_encoder_vulkan.cpp
```diff
@@ -638,14 +638,18 @@ void wivrn::video_encoder_vulkan::present_image(vk::Image y_cbcr, vk::SemaphoreS
 		compositor_sem.stageMask = vk::PipelineStageFlagBits2::eCopy;
 
 		beman::inplace_vector::inplace_vector<vk::ImageMemoryBarrier2, 2> im_barriers;
+		// y_cbcr was written by the compositor in eGeneral and stays in
+		// eGeneral here. eGeneral is a valid source layout for copyImage,
+		// so no layout transition is needed; the QFOT acquire below uses
+		// eGeneral → eGeneral to match the compositor's release.
 		vk::ImageLayout y_cbcr_layout{vk::ImageLayout::eGeneral};
```
xytovl: Then remove the `y_cbcr_layout` variable if it's a constant.

Using General layout may have lower performance on some GPUs, which is why I added the transition. It may be irrelevant for general -> transfer src and this case is NVIDIA only so I can't really test.

server/compositor/compositor.cpp
xytovl: > If srcQueueFamilyIndex and dstQueueFamilyIndex are equal and oldLayout and newLayout are also equal, the layout values are ignored and the image contents are preserved regardless of the values of oldLayout, newLayout, and the current layout of the image.

ok, maybe we still need them, but we should avoid the double barrier for transfer and layout transition. We may solve it like the `target_queue` member and have a `target_layout`

xytovl:
> There are still performance differences between the `nvenc` and `vulkan` backends, but this seems to be similar to existing differences with what we had before.

Were you able to see where the performance difference comes from? Are encode times higher (in csv we log encode_start as the beginning of the encode method, but using vulkan it may start earlier, it's more accurate to check submit to encode duration)?

damaestro:
> > There are still performance differences between the `nvenc` and `vulkan` backends, but this seems to be similar to existing differences with what we had before.
> 
> Were you able to see where the performance difference comes from? Are encode times higher (in csv we log encode_start as the beginning of the encode method, but using vulkan it may start earlier, it's more accurate to check submit to encode duration)?

I started working on instrumenting with perfetto. I've ran out of time today but the instrumentation is looking pretty good. We are actually able to capture mesa + monado + wivrn. These traces target just wivrn:

image

Load these traces via https://ui.perfetto.dev/ to see in detail.
[pftraces.zip](https://github.com/user-attachments/files/27946017/pftraces.zip)

I don't think the perf differences should block this fix. They are possibly existing and can be worked on in another PR?

https://github.com/WiVRn/WiVRn/issues/922

xytovl:
https://github.com/WiVRn/WiVRn/tree/feat/multiple-vulkan-encode-queues should be all we need to use multiple vulkan encode queues.
I also added VK_KHR_unified_image_layouts to skip layout transitions when possible

damaestro:
> https://github.com/WiVRn/WiVRn/tree/feat/multiple-vulkan-encode-queues should be all we need to use multiple vulkan encode queues. I also added VK_KHR_unified_image_layouts to skip layout transitions when possible

There is one bug on that branch it seems:

```
diff --git a/server/encoder/video_encoder_nvenc.cpp b/server/encoder/video_encoder_nvenc.cpp
index f04f3d45..7a86d3b3 100644
--- a/server/encoder/video_encoder_nvenc.cpp
+++ b/server/encoder/video_encoder_nvenc.cpp
@@ -418,8 +418,8 @@ video_encoder_nvenc::video_encoder_nvenc(
                auto props = vk.physical_device.getProperties();
                timestamp_period_ns = props.limits.timestampPeriod;
                auto qprops = vk.physical_device.getQueueFamilyProperties();
-               if (vk.queue_family_index < qprops.size())
-                       timestamp_supported = qprops[vk.queue_family_index].timestampValidBits > 0;
+               if (vk.queue.family_index < qprops.size())
+                       timestamp_supported = qprops[vk.queue.family_index].timestampValidBits > 0;
                if (timestamp_supported)
                {
                        timestamp_pool = vk.device.createQueryPool(vk::QueryPoolCreateInfo{
diff --git a/server/encoder/video_encoder_vulkan.cpp b/server/encoder/video_encoder_vulkan.cpp
index d10e68e9..809ba4a6 100644
--- a/server/encoder/video_encoder_vulkan.cpp
+++ b/server/encoder/video_encoder_vulkan.cpp
@@ -573,8 +573,8 @@ void wivrn::video_encoder_vulkan::init(const vk::VideoCapabilitiesKHR & video_ca
                timestamp_period_ns = props.limits.timestampPeriod;
                // Need timestamps to be valid on the video-encode queue.
                auto qprops = vk.physical_device.getQueueFamilyProperties();
-               if (vk.encode_queue_family_index < qprops.size())
-                       timestamp_supported = qprops[vk.encode_queue_family_index].timestampValidBits > 0;
+               if (encode_queue.family_index < qprops.size())
+                       timestamp_supported = qprops[encode_queue.family_index].timestampValidBits > 0;
                if (timestamp_supported)
                {
                        timestamp_pool = vk.device.createQueryPool(vk::QueryPoolCreateInfo{
```

However, interesting results. 

Screenshot From 2026-05-18 19-23-01


[q-vulkan.zip](https://github.com/user-attachments/files/27980195/q-vulkan.zip)

damaestro:
> However, interesting results.

It looks like this sorts out  part of the issue, but oddly not with the encode latency. It helps with the flow of encoded frames.

  Encoder side (server) — essentially identical

  ```  
  ┌─────────────────────┬───────────────┬───────────────┬──────────────────────┐
  │        slice        │   pre-queue   │    queued     │          Δ           │
  ├─────────────────────┼───────────────┼───────────────┼──────────────────────┤
  │ encodeVideoKHR n    │ 1924          │ 1912          │ ~same                │
  ├─────────────────────┼───────────────┼───────────────┼──────────────────────┤
  │ encode mean         │ 7.78 ms       │ 7.84 ms       │ ~same                │
  ├─────────────────────┼───────────────┼───────────────┼──────────────────────┤
  │ encode max          │ 10.52 ms      │ 8.26 ms       │ ~2.3 ms tail removed │
  ├─────────────────────┼───────────────┼───────────────┼──────────────────────┤
  │ encoder_work iter   │ 15.59 ms      │ 15.68 ms      │ ~same                │
  ├─────────────────────┼───────────────┼───────────────┼──────────────────────┤
  │ SendData mean / max │ 0.139 / 0.332 │ 0.144 / 0.416 │ ~same                │
  └─────────────────────┴───────────────┴───────────────┴──────────────────────┘
```

  The encoder throughput is unchanged, but queued has cut the worst-case encode outlier from 10.5 ms down
   to 8.3 ms. Tail is much tighter.

  Receive / decode / display side — major differences

```
  ┌───────────────┬────────────────┬─────────────┬──────┐
  │     slice     │ pre-queue rate │ queued rate │  Δ   │
  ├───────────────┼────────────────┼─────────────┼──────┤
  │ receive_begin │ 952 /s         │ 1083 /s     │ +14% │
  ├───────────────┼────────────────┼─────────────┼──────┤
  │ receive_end   │ 901 /s         │ 1031 /s     │ +14% │
  ├───────────────┼────────────────┼─────────────┼──────┤
  │ decode_begin  │ 901 /s         │ 1031 /s     │ +14% │
  ├───────────────┼────────────────┼─────────────┼──────┤
  │ decode_end    │ 775 /s         │ 904 /s      │ +17% │
  ├───────────────┼────────────────┼─────────────┼──────┤
  │ blit          │ 744 /s         │ 904 /s      │ +22% │
  ├───────────────┼────────────────┼─────────────┼──────┤
  │ display       │ 744 /s         │ 904 /s      │ +22% │
  └───────────────┴────────────────┴─────────────┴──────┘
```

  Two things stand out:

  1. Frame drops between decode_end and blit are eliminated. In pre-queue, 11,757 frames finish decoding
  but only 11,292 reach blit — ~465 decoded frames (≈4%) are discarded before display. In queued, every
  decoded frame reaches blit/display (13,682 → 13,682).
  2. More receive/decode events per second overall. queued is moving 14–22% more packets/frames through
  the receive→decode→display chain in the same 15-second window. Combined with point 1, this looks like
  the queued pipeline is keeping the client busy and not stalling, whereas pre-queue was leaving the
  receiver/decoder under-fed and then dropping frames before blit.

damaestro:
@xytovl let's merge your branch with that small bug fix and then we can iterate on the vulkan encoder off `main`.

I'd like to get the perfetto work rebased on main after the merge and put up a PR to add the better tooling. :)

xytovl:
> > https://github.com/WiVRn/WiVRn/tree/feat/multiple-vulkan-encode-queues should be all we need to use multiple vulkan encode queues. I also added VK_KHR_unified_image_layouts to skip layout transitions when possible
> 
> There is one bug on that branch it seems:
> 
> ```
> diff --git a/server/encoder/video_encoder_nvenc.cpp b/server/encoder/video_encoder_nvenc.cpp
> index f04f3d45..7a86d3b3 100644
> --- a/server/encoder/video_encoder_nvenc.cpp
> +++ b/server/encoder/video_encoder_nvenc.cpp
> @@ -418,8 +418,8 @@ video_encoder_nvenc::video_encoder_nvenc(
>                 auto props = vk.physical_device.getProperties();
>                 timestamp_period_ns = props.limits.timestampPeriod;
>                 auto qprops = vk.physical_device.getQueueFamilyProperties();
> -               if (vk.queue_family_index < qprops.size())
> -                       timestamp_supported = qprops[vk.queue_family_index].timestampValidBits > 0;
> +               if (vk.queue.family_index < qprops.size())
> +                       timestamp_supported = qprops[vk.queue.family_index].timestampValidBits > 0;
>                 if (timestamp_supported)
>                 {
>                         timestamp_pool = vk.device.createQueryPool(vk::QueryPoolCreateInfo{
> diff --git a/server/encoder/video_encoder_vulkan.cpp b/server/encoder/video_encoder_vulkan.cpp
> index d10e68e9..809ba4a6 100644
> --- a/server/encoder/video_encoder_vulkan.cpp
> +++ b/server/encoder/video_encoder_vulkan.cpp
> @@ -573,8 +573,8 @@ void wivrn::video_encoder_vulkan::init(const vk::VideoCapabilitiesKHR & video_ca
>                 timestamp_period_ns = props.limits.timestampPeriod;
>                 // Need timestamps to be valid on the video-encode queue.
>                 auto qprops = vk.physical_device.getQueueFamilyProperties();
> -               if (vk.encode_queue_family_index < qprops.size())
> -                       timestamp_supported = qprops[vk.encode_queue_family_index].timestampValidBits > 0;
> +               if (encode_queue.family_index < qprops.size())
> +                       timestamp_supported = qprops[encode_queue.family_index].timestampValidBits > 0;
>                 if (timestamp_supported)
>                 {
>                         timestamp_pool = vk.device.createQueryPool(vk::QueryPoolCreateInfo{
> ```

This may be uncommited changes on your side, I don't have the `timestamp_supported` variables.

xytovl:
I don't know exactly what your traces are capturing, the separate encode queues should cause left and right eye to be encoded concurrently. GPU times for each eye would remain unchanged but the total time for both should be cut in half.

xytovl:
An other element that may be interesting is to check how `vk::VideoEncodeUsageInfoKHR` impacts NVIDIA.
Long ago, I had to check in `get_video_caps` if `getVideoCapabilitiesKHR` fails when usage info is provided and when it does disable it entirely.
The video profile info is used to get the capabilities, but also to create the images and the video session. I suppose the tuning parameters are only important for the video session itself, maybe relinking the `vk::VideoEncodeUsageInfoKHR` before the call to `::init` works and gives the hint to do low latency streaming of rendered content.

damaestro:
> I don't know exactly what your traces are capturing, the separate encode queues should cause left and right eye to be encoded concurrently. GPU times for each eye would remain unchanged but the total time for both should be cut in half.

I'll cleanup the tracing code once we merge this line of work from your branch and put up a PR.

PR #915, opened 2026-05-17
Implement auto-connect feature

openlyst:
Adds a new settings option, when a vr headset gets plugged in and paired it will auto connect to it using USB.

image


dashboard/qml/SettingsPage.qml
```diff
@@ -127,6 +127,10 @@ Kirigami.ScrollablePage {
                     toolTipText: i18n("Allows the use of lighthouse-based controllers and trackers.\nRequires SteamVR to be installed.\nDevices must be be powered on before connecting to WiVRn.\nAn external tool such as motoc is needed for calibration.")
                 }
             }
+            Controls.CheckBox {
+                id: auto_connect_usb
+                text: i18n("Auto connect from USB")
+            }
```
xytovl: We could move that out of advanced block, next to the application autostart

PR #926, opened 2026-05-20
Fix dashboard settings race master (#925)

### LINKED ISSUE #925: [Bug]: Dashboard settings show stale/default values until reopening — race between D-Bus properties and SettingsPage load
### Describe the bug

When `wivrn-dashboard` is launched and it spawns `wivrn-server` itself (i.e., the server was not already running via systemd), opening the Settings page right away shows incorrect values:

- "Encoder" reverts to *Auto* even though `~/.config/wivrn/config.json` has `"encoder": "vaapi"`
- "Autostart application" appears empty even though `"application"` is set
- "Enable SteamVR tracked devices support" checkbox appears unchecked even though `"use-steamvr-lh": true` is saved and the feature actually works at runtime
- "OpenVR compatibility library" is blank
The underlying configuration on disk is correct and the server actually honors it (Lighthouse devices show up, etc.). It is purely a UI/state-sync problem.

Root cause is a chain of two race conditions:

1. **`dashboard/wivrn_server.cpp` — `refresh_server_properties()`**
   `on_server_dbus_registered()` is triggered as soon as the server owns the D-Bus name `io.github.wivrn.Server`, but the server's `on_name_acquired` callback in `server/main.cpp` is the place that actually calls `wivrn_server_set_json_configuration()` and `g_dbus_interface_skeleton_export()`. The dashboard's `GetAll` request can therefore arrive before the interface is exported / properties are populated. The reply comes back with `isError() == true` or as an empty `QVariantMap` (no `JsonConfiguration` key), and the dashboard silently treats this as "the server has no configuration", never retrying.

2. **`dashboard/qml/SettingsPage.qml` — `Component.onCompleted`**
   Even if (1) were fixed, the page only ever loads settings once in `Component.onCompleted`. If the user opens Settings before the asynchronous `JsonConfiguration` D-Bus property has been received, `Settings.load(WivrnServer)` parses an empty string and the UI is populated with default values. There is no `Connections { target: WivrnServer; function onJsonConfigurationChanged() ... }` to recover when the property eventually arrives.

The combined symptom is that the very first time the user opens Settings after launching the dashboard (without `wivrn.service` already running), all values appear wrong. Closing and reopening the Settings page sometimes "fixes" it because by then the server has populated the property.

The server log around the failed `GetAll` shows nothing wrong on the server side — it simply finishes starting normally.

Relevant code:
- https://github.com/WiVRn/WiVRn/blob/v26.2.3/dashboard/wivrn_server.cpp#L274-L288
- https://github.com/WiVRn/WiVRn/blob/v26.2.3/dashboard/qml/SettingsPage.qml#L247-L252
- https://github.com/WiVRn/WiVRn/blob/v26.2.3/server/main.cpp#L787-L860

### Steps to Reproduce

1. Have a `~/.config/wivrn/config.json` with at least one non-default value, e.g.:
   ```json
   {
     "encoder": "vaapi",
     "use-steamvr-lh": true,
     "application": ["wayvr"]
   }
   ```
2. Ensure `wivrn-server` is **not** running (`systemctl --user stop wivrn`, kill any stray `wivrn-server` process).
3. Launch `wivrn-dashboard` (let it start the server itself).
4. As soon as the main window appears, click **Settings**.
5. Observe: the Encoder is "Auto", the SteamVR checkbox is unchecked, OpenVR / Autostart are empty.
6. Close Settings, reopen — values are now usually correct (because the D-Bus property has arrived in the meantime).

To verify the underlying config is fine:
```bash
systemctl --user start wivrn
busctl --user get-property io.github.wivrn.Server \
       /io/github/wivrn/Server \
       io.github.wivrn.Server JsonConfiguration
# → returns the full, correct JSON
```

### Expected behavior

When the Settings page opens, all controls (Encoder, Autostart application, OpenVR library, all advanced checkboxes including "Enable SteamVR tracked devices support") must reflect the configuration that the server has actually loaded from `config.json`, regardless of whether `JsonConfiguration` happened to arrive over D-Bus before or after the page was constructed.

### WiVRn Version

26.2.3 + 1f71ade(master)

### Installation method

Arch Linux, wivrn-dashboard 26.2.3-1 from AUR

### Server Config

_No response_

### Server Logs

_No response_

AMD

### Vulkan Info

```text

```

HTC Vive XR Elite

### ADB Logs

```plain

```

clansty:
This PR fixed #925 

Test on my machine

### DIFF
--- a/dashboard/wivrn_server.cpp
+++ b/dashboard/wivrn_server.cpp
@@ -25,6 +25,7 @@
 #include <QApplication>
 #include <QClipboard>
 #include <QDesktopServices>
+#include <QTimer>
 #include <QtLogging>
 #include <cassert>
 #include <memory>
@@ -52,7 +53,14 @@ wivrn_server::wivrn_server(QObject * parent) :
 
 	const QStringList services = QDBusConnection::sessionBus().interface()->registeredServiceNames();
 	if (services.contains("io.github.wivrn.Server"))
+	{
+		// Server is already up. Move to a transitional state so Main.qml's
+		// onCompleted does not see Status::Stopped and spawn a duplicate
+		// process. The state will transition to Started after the initial
+		// properties have been fetched (see refresh_server_properties).
+		m_serverStatus = Status::Starting;
 		on_server_dbus_registered();
+	}
 }
 
 wivrn_server::~wivrn_server()
@@ -220,8 +228,6 @@ void wivrn_server::on_server_ready_read_standard_output()
 
 void wivrn_server::on_server_dbus_registered()
 {
-	serverStatusChanged(m_serverStatus = Status::Started);
-
 	if (server_interface)
 		server_interface->deleteLater();
 	if (server_properties_interface)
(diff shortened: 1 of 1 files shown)


xytovl:
Thanks for investigating the issue.

I think a cleaner fix would be to delay the ServerStatus change to Started after dbus properties are populated.

dashboard/wivrn_server.cpp
xytovl: Remove both (?!) `serverStatusChanged` calls here, and move them inside `refresh_server_properties`'s lambda, after the successful call to `on_server_properties_changed`.

This should replace the functionality of your first commit and make it safer to use any function in the qml when server is started.

clansty:
Changed. Tested in both scenarios (systemd --user and non systemd)

PR #927, opened 2026-05-20
Attempting some fixes.

ZokuTe:
It didn't really solve the problem, but it seems to have improved somewhat.

### DIFF
--- a/server/encoder/ffmpeg/video_encoder_va.cpp
+++ b/server/encoder/ffmpeg/video_encoder_va.cpp
@@ -432,7 +432,7 @@ video_encoder_va::video_encoder_va(wivrn::vk_bundle & vk,
 	}
 }
 
-void video_encoder_va::present_image(vk::Image y_cbcr, vk::SemaphoreSubmitInfo, uint8_t slot, uint64_t)
+void video_encoder_va::present_image(vk::Image y_cbcr, vk::SemaphoreSubmitInfo compositor_sem, uint8_t slot, uint64_t)
 {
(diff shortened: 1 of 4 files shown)


server/encoder/ffmpeg/video_encoder_va.cpp
```diff
@@ -540,9 +540,12 @@ void video_encoder_va::present_image(vk::Image y_cbcr, vk::SemaphoreSubmitInfo,
 	vk::CommandBufferSubmitInfo cmd_info{
 	        .commandBuffer = *cmd,
 	};
+	compositor_sem.stageMask = vk::PipelineStageFlagBits2::eTransfer;
 
 	vk.device.resetFences(*in[slot].fence);
 	vk.queue.queue.submit2(vk::SubmitInfo2{
+	                               .waitSemaphoreInfoCount = 1,
+	                               .pWaitSemaphoreInfos = &compositor_sem,
```
xytovl, line 548: We don't need semaphores for commands submitted on the same queue.
command buffers are executed in submission order

ZokuTe: I don't know why, but this does alleviate the issue of the client inexplicably requesting a large number of IDRs.

xytovl: The semaphore should not have any effect, it may introduce timing changes or require additional synchronization.
Are you able to see why there's IDRs? base idr_handler now has debug logs to know what causes them.

PR #930, opened 2026-05-24
Allow access to XDG_DATA_HOME SlimeVR socket path for Flatpak

ImSapphire:
Used when running SlimeVR in Steam: https://github.com/SlimeVR/SlimeVR-Server/blob/61687a27ca1d5ee4eee98dc4b244b0687536d58e/server/core/src/main/java/io/eiren/util/OperatingSystem.kt#L42


xytovl:
This time we should remember to do a pull request in https://github.com/flathub/io.github.wivrn.wivrn/ too with just the permission changes, so that we don't get blocked on the next release.

xytovl (approved):

ImSapphire:
flathub/io.github.wivrn.wivrn#40

PR #933, opened 2026-05-25
Add PFDM MR controller support

EasonZxp:


common/wivrn_packets.h
```diff
@@ -160,6 +160,7 @@ enum class interaction_profile : uint8_t
 	meta_touch_controller_rift_cv1,
 	meta_touch_controller_quest_1_rift_s,
 	meta_touch_controller_quest_2,
+	yvr_touch_controller_yvr,
```
xytovl, line 163: There's a limitation in the versioning system, when adding values to enums, you should change the protocol revision variable.

xytovl:
Can you try and synchronize with https://github.com/WiVRn/WiVRn/pull/896 from @amwatson to have full support?

common/wivrn_packets.h
EasonZxp: Have bumped the version. Thanks.

client/CMakeLists.txt
```diff
@@ -290,7 +290,7 @@ wivrn_webxr_controller(PROFILE pico-4                  DESTINATION ${ASSETS_DIR}
 wivrn_webxr_controller(PROFILE pico-4u                 DESTINATION ${ASSETS_DIR}/controllers TARGET wivrn)
 wivrn_webxr_controller(PROFILE pico-neo3               DESTINATION ${ASSETS_DIR}/controllers TARGET wivrn)
 wivrn_webxr_controller(PROFILE samsung-galaxyxr        DESTINATION ${ASSETS_DIR}/controllers TARGET wivrn)
-
+wivrn_webxr_controller(PROFILE yvr-touch-v2            DESTINATION ${ASSETS_DIR}/controllers TARGET wivrn)
```
xytovl, line 293: Does the headset support `XR_EXT_render_model`?

I don't own any that has the extension, but it would be nice to support that instead of embedded models.

EasonZxp: Currently, the PFDM MR runtime does not yet support XR_EXT_render_model. :(

xytovl:
We'll eventually have to add the interaction profile on server side too, but Quest Touch emulation should be fine for now

PR #934, opened 2026-05-25
Add FB hand mesh support with GLB fallback

EasonZxp:
Use the XR_FB_HAND_TRACKING_MESH extension to render hand models when supported, and fall back to the bundled GLB hand models otherwise. This fixes incorrect hand model rendering on PFDM MR device, Have tested with PFDM MR V4.5.0 version.


client/xr/hand_tracker.cpp
```diff
@@ -87,6 +93,129 @@ std::optional<std::array<xr::hand_tracker::joint, XR_HAND_JOINT_COUNT_EXT>> xr::
...
+		data.vertex_uvs.resize(mesh.vertexCountOutput);
+		data.vertex_blend_indices.resize(mesh.vertexCountOutput);
+		data.vertex_blend_weights.resize(mesh.vertexCountOutput);
+		data.indices.resize(mesh.indexCountOutput);
+	}
+	catch (const std::bad_alloc &)
+	{
+		spdlog::warn("Hand mesh allocation failed for {} vertices and {} indices", mesh.vertexCountOutput, mesh.indexCountOutput);
+		return nullptr;
+	}
```
xytovl: You don't have to catch `bad_alloc`, it only happens when the requested size is absurd

PR #952, opened 2026-05-28
CMake: Auto-detect optional libraries which use pkgconfig

JoshuaVandaele:
This allows configure-time detection of available libraries and toggles functionality ON or OFF automatically. This was primarily made for libsystemd, but I extended this to other optional libraries that also use pkgconfig.

If the value of the option is set to AUTO (default), the feature will be toggled depending on the presence of the required libraries on the system. If the value of the option is set to ON or OFF, the old behavior remains.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ check_pie_supported()
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 include(FetchContent)
+include(FindOptional)
 include(ExternalProject)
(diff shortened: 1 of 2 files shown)


xytovl:
Automatic dependencies have been removed in 211922ea02d5f0eae7bfaf5048a9369a59a9d392

It made it harder for packagers to know what is being built and users could silently miss features.

JoshuaVandaele:
Ah, fair enough I suppose.

Issue #959, opened 2026-05-31
[Feature Request]: Detect and offer a fix for NVIDIA P2-state performance limiting during NVENC encoding


unimpressive-user:
### Feature description

When an NVIDIA GPU is detected, WiVRn could check whether the NVIDIA application profile configuration contains the settings required to exclude wivrn-server from the default P2-state performance limitation that is applied while NVENC is in use.

### Reason for the feature

NVIDIA drivers limit the GPU to the P2 performance state during NVENC encoding.
While this behavior is intended to improve stability, it can also reduce gaming performance.
In my testing, excluding wivrn-server from this limitation resulted in a noticeable performance improvement in GPU heavy games without introducing any stability issues. 

### Proposed implementation

If the required configuration file is missing, or if it does not contain the necessary profile entries, the WiVRn Dashboard could display a notification similar to the existing firewall warning.
The notification would explain that NVIDIA's default NVENC behavior may reduce game performance and offer a one-click fix.

Configuration file path: `$HOME/.nv/nvidia-application-profiles-rc`

My configuration file: [nvidia-application-profiles-rc.txt](https://github.com/user-attachments/files/28440596/nvidia-application-profiles-rc.txt)

xytovl:
The only documentation I can find is https://download.nvidia.com/XFree86/Linux-x86/384.59/README/profiles.html and seems to be out of date.

Does this happen when using vulkan video encode too?

unimpressive-user:
Vulkan encoder is working fine. 
Also this was my source for the config https://github.com/NVIDIA/open-gpu-kernel-modules/issues/333#issuecomment-3669477571

xytovl:
I don't think it's worth it adding a dashboard check and especially messing with user configuration.

I plan on recommending users to try vulkan encode on NVIDIA starting on the next release and if feedback is positive I'll switch the default on the following one.

unimpressive-user:
Yeah, that's understandable.
Could it be added in to the docs? Just a few lines about NVENC can be affected by the P2-state performance limitation, the potential impact on game performance, and how users can work around it if they encounter the issue.

xytovl:
docs or maybe the troubleshoot tab are fine (although I think that one has to go, it's not very up to date).

Can you make a pull request? You seem to know more about those power state, what the issue really is and what the configuration does.

unimpressive-user:
Yeah, I can do a PR. Just want to look into it a bit more first.

PR #965, opened 2026-06-01
NVIDIA NVENC P-states issue docs

unimpressive-user:
https://github.com/WiVRn/WiVRn/issues/959
This adds a new documentation page describing a P-state behavior seen when using NVIDIA NVENC.
I’m not fully confident in the technical framing of the explanation, so feedback on correctness, wording, and whether the workaround is appropriate would be especially appreciated.

### DIFF
--- a/docs/nvenc.md
+++ b/docs/nvenc.md
@@ -1,0 +1,52 @@
+# Notes for NVIDIA NVENC encoder
+
+### Introduction
+NVIDIA GPUs use performance states (P-states), which influence how the driver manages GPU core and memory clocks. P0 is generally the highest-performance state, while P8 is typically one of the lowest-power states. 
+
+WiVRn's NVIDIA NVENC encoder implementation relies on CUDA, which may cause NVIDIA drivers to keep the GPU in the P2 performance state instead of allowing it to enter P0.
+This behavior is associated with CUDA workloads and may be used by NVIDIA drivers to improve stability.
+
+NVIDIA drivers support an application profile setting named `CudaNoStablePerfLimit`, users have reported that it resolves the issue. [open-gpu-kernel-modules](https://github.com/NVIDIA/open-gpu-kernel-modules/issues/333#issuecomment-3669477571)
+
+Some distributions already ship application profiles that exclude specific applications from these performance limits. You can see examples of how they apply this in [Pop!_OS](https://github.com/pop-os/nvidia-graphics-drivers/blob/master/debian/application-profiles/cuda-no-stable-perf-limit) or [CachyOS](https://github.com/CachyOS/CachyOS-PKGBUILDS/blob/master/nvidia/nvidia-utils/cuda-no-stable-perf-limit).
+
+### Identifying the issue
+
+To check whether your system is affected, monitor the GPU P-State using nvidia-smi or another GPU monitoring tool while playing a game with the NVIDIA NVENC encoder enabled. If the GPU remains in P2 under load while using NVENC, but reaches P0 with the Vulkan encoder or after applying this profile, the system is likely affected.
+
+### Workaround instructions
+> [!NOTE] 
+> If your GPU already enters P0 while using the NVIDIA encoder, this workaround is unlikely to provide any benefit.
+
+1. Create the NVIDIA configuration directory
+
+    The application profiles configuration file is located at:
+
+    `$HOME/.nv/nvidia-application-profiles-rc`
+
+    If the `.nv` directory does not already exist, create it:
+
+    `mkdir -p "$HOME/.nv"`
+
+2. Create the configuration file
+
+    Open `$HOME/.nv/nvidia-application-profiles-rc` in your preferred text editor and add the following contents:
+```
+{
+    "profiles": [
+        {
+            "name": "CudaNoStablePerfLimit",
+            "settings": ["0x166c5e", 0 ]
+        }
+    ],
+    "rules": [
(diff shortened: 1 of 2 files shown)


xytovl:
The docs seem relevant, the CachyOS and PopOS config seem to add the config for popular software that uses the encoder so it's likely the right thing to do.

I can't say if the details are correct, but look correct.

As just a file in the docs folder, the main benefit would be for search engines to find the file or to point during discussions.
Adding a link from the main readme in the troubleshooting question would increase visibility.

@damaestro as you helped on docs/packaging/NVIDIA in the past, feel free to add comments if you have any

damaestro:
Let's go a little bit further so we can adjust this profile in the future.

docs/nvenc.md
```diff
@@ -0,0 +1,52 @@
...
+
+    If the `.nv` directory does not already exist, create it:
+
+    `mkdir -p "$HOME/.nv"`
+
+2. Create the configuration file
+
+    Open `$HOME/.nv/nvidia-application-profiles-rc` in your preferred text editor and add the following contents:
+```
+{
```
damaestro, line 35:
```
{
    "profiles": [
        {
            "name": "WiVRnServer",
            "settings": [
                {
                  "key"   : "0x166c5e",
                  "value" : 0
                },
                {
                  "key"   : "GLVRRAllowed",
                  "value" : false
                },
                {
                  "key"   : "VKDirectGSYNCAllowed",
                  "value" : false
                },
                {
                  "key"   : "VKDirectGSYNCCompatibleAllowed",
                  "value" : 0
                },
                {
                  "key"   : "GLShowGraphicsOSD",
                  "value" : false
                }
            ]
        }
    ],
    "rules": [
        { "pattern" : { "feature" : "procname", "matches" : "wivrn-server" }, "profile" : "WiVRnServer" }
    ]
}
```

damaestro: Let's

1. Name it something WiVRn specific, so we can set multiple options. The profile you have defined already is defined in the upstream `nvidia-application-profiles-rc`.
2. When we look at the 595.71.05 driver settings, I noticed a lot of important things have the "No VRR/OSD" profile attached. Above is the "No VRR/OSD" and "CudaNoStablePerfLimit" profiles combined as a policy for our purposes.

docs/nvenc.md
```diff
@@ -0,0 +1,52 @@
...
+        { "pattern" : { "feature" : "procname", "matches" : "wivrn-server" }, "profile" : "CudaNoStablePerfLimit" }
+    ]
+}
+
+```
+3. Save the file
+
+    Save the file and `restart wivrn-server`. The profile should be applied automatically by the NVIDIA driver.
+
+`0x166c5e` is an internal NVIDIA driver setting ID for CUDA performance limiting behavior and `0` disables this limitation.
```
damaestro: Create a table and explain each option we have set. The `0x166c5e` option can be referenced as "CUDA - Force P2 State".

damaestro (dismissed):
Good addition for those using a distro that has not done any of this tuning and/or the default upstream settings.

Please see my comments and add a section about monitoring p-state. Include something like:

```
watch nvidia-smi -q -d PERFORMANCE
```

docs/nvenc.md
unimpressive-user: Hi, I replaced the old configuration with the one requested by you.

xytovl: Are you sure VRR/GSYNC/OSD options have any effect on WiVRn?
wivrn-server doesn't spawn any GUI (unless using the debug window) so there's no swapchain and I don't see what those settings could do in that case.

docs/nvenc.md
unimpressive-user: Hi, I think it may be worth mentioning that I relied on the following NVIDIA documentation when creating the table.
[Source 1](https://download.nvidia.com/XFree86/Linux-x86_64/610.43.02/README/profiles.html), [Source 2](https://download.nvidia.com/XFree86/Linux-x86_64/610.43.02/README/openglenvvariables.html), [Source 3](https://download.nvidia.com/XFree86/Linux-x86_64/610.43.02/README/configtwinview.html)

docs/nvenc.md
damaestro: I just trolled through `/usr/share/nvidia/nvidia-application-profiles-rc` to see what might make sense. We can see that window managers, steam, vrmonitor, etc all have the VRR/OSD profile.

obs (most similar) only has the `CudaNoStablePerfLimit` profile defined.

I also am looking at the desktop compositors, they all have "No VidMem Reuse" which we might need to consider. (`"settings": [ "GLVidHeapReuseRatio", 0 ]`).

caszuu: afaik `GLVidHeapReuseRatio` purely affects OpenGL programs. It controls how much not-in-use vram a program can keep reserved from the rest of the system and setting it to 0 is a workaround for opengl based compositors sometimes holding 5x the vram they'd otherwise need.

I don't think it will affect wivrn in any way.

damaestro: I suggest we move forward with setting this like obs, and only enable the `0x166c5e` option for now. Keep the multi-option format and the custom named profile for any future reconsideration.

damaestro (changes requested):
Sorry for the back and forth. Let's only set `0x166c5e` for now.

damaestro (approved):

PR #970, opened 2026-06-06
Add WiVRn-native Perfetto via Perfetto SDK

damaestro:
## Perfetto tracing via Perfetto SDK

Adds opt-in, zero-overhead Perfetto tracing to the server-side pipeline.

Screenshot From 2026-06-06 16-52-59

Optional server profiling built on the Perfetto SDK, independent of Monado's own tracing. Adds CPU spans (compositor, encoder, network), GPU encode/copy slices from Vulkan timestamp queries (Vulkan/NVENC/VAAPI), and headset feedback markers from `dump_time`. Compiled out unless built with `WIVRN_USE_PERFETTO`.

### Motivation

Diagnosing encoder latency, compositor stalls, and network jitter today means adding printf timing or parsing CSV dumps offline. This gives a timeline view of every pipeline stage with hardware-accurate GPU timestamps, optionally on a single timeline with Monado and full system detail.

### Configuration

- `WIVRN_USE_PERFETTO` (CMake, default OFF) — build gate; fully compiled out when off.
- `WIVRN_TRACE_MONADO` (CMake, default OFF, requires `WIVRN_USE_PERFETTO`) — also build Monado with percetto so its `u_trace` markers join the same session (system backend).
- `WIVRN_TRACING` (env) — single runtime switch that enables tracing and selects the backend: `system` (connect to an external `traced`), `inprocess` (run the service in-process and write the trace), or `system+inprocess`.

### What's added

**Infrastructure** (`server/utils/wivrn_trace.{h,cpp}`, `gpu_timestamp_pool.{h,cpp}`)
- `wivrn::trace::scope` — RAII CPU span (plus `cpu_begin`/`cpu_end` for spans crossing calls, `cpu_instant`); single relaxed atomic load per site when enabled, no-op when disabled or unbuilt.
- `gpu_slice()` — GPU-hardware-timestamped spans via Vulkan query pools, converted to host monotonic ns through Monado's calibration (Vulkan encode/copy, NVENC pre-copy, VAAPI pre-copy).
- `instant_feedback()` — feedback-track markers aligned with the CSV `dump_time` timestamps.

**Call sites** — all three encoder paths (Vulkan, NVENC, VAAPI) instrumented consistently: `present_image`, `encode`, encoder-specific GPU spans, `encoder_work_iter`, `SendData`, and the `dump_time` feedback markers, with span starts aligned to the CSV timestamps.

**Build/CI** — add new pipeline to ensure builds work and prevent bitrot

**Tooling** (`tools/perfetto/`)
- `wivrn_capture.py` — automatic capture setup
- `pftrace_summary.py` — CLI stats Δmean for two traces

**Docs** — `docs/profiling.md`

### Build

```cmake
cmake --preset server-tracing
cmake --build build-server-tracing
```

Defaults OFF.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -9,6 +9,7 @@
 		{
 			"name": "base",
 			"hidden": true,
+			"generator": "Ninja",
 			"cacheVariables": {
 				"CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
 				"CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON",
@@ -63,6 +64,24 @@
 				"WIVRN_BUILD_SERVER": "ON",
 				"WIVRN_BUILD_SERVER_LIBRARY": "ON",
 				"WIVRN_BUILD_WIVRNCTL": "ON"
+			}
+		},
+		{
+			"name": "server-tracing",
+			"displayName": "Server with Perfetto tracing",
+			"inherits": "server",
+			"binaryDir": "${sourceDir}/build-server-tracing",
+			"cacheVariables": {
+				"WIVRN_USE_PERFETTO": "ON"
+			}
+		},
+		{
+			"name": "server-tracing-monado",
+			"displayName": "Server with Perfetto tracing (incl. Monado)",
+			"inherits": "server-tracing",
+			"binaryDir": "${sourceDir}/build-server-tracing-monado",
+			"cacheVariables": {
+				"WIVRN_TRACE_MONADO": "ON"
 			}
 		},
 		{
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -1,0 +1,190 @@
+# Profiling WiVRn
+
+Performance analysis for WiVRn: per-frame timing CSV and [Perfetto](https://perfetto.dev/) tracing.
+For general debugging see [docs/debugging.md](debugging.md).
+
+## Quick start
+
+Build with tracing (`server-tracing` preset = `server` + `WIVRN_USE_PERFETTO=ON`):
+```bash
+cmake --preset server-tracing
+cmake --build build-server-tracing
+```
+
+Self-contained **in-process** trace (no daemon, no script):
+```bash
+WIVRN_TRACING=inprocess build-server-tracing/server/wivrn-server
+```
+Connect the HMD, stream, disconnect. Each session writes a `wivrn-<timestamp>-<pid>.pftrace` in cwd;
+open it in <https://ui.perfetto.dev>.
+
+**System-backend** capture via the daemon + wrapper:
+```bash
+WIVRN_TRACING=system wivrn-server     # one shell, with a client streaming
+tools/perfetto/wivrn_capture.py       # WiVRn-only; --full adds Monado + ftrace
+```
+
+`WIVRN_USE_PERFETTO` defaults `OFF`; even built `ON`, tracing is inert until `WIVRN_TRACING` is set.
+The Perfetto amalgamated SDK must be installed where CMake looks (`/usr/share/perfetto/sdk`, override
+with `-DPERFETTO_SDK_DIR=<dir>`); no automatic fetch.
+
+## Timing analysis (CSV)
+
+Per-frame timing without any Perfetto setup — same events as the `wivrn_feedback` track, one row each:
+```bash
+WIVRN_DUMP_TIMINGS=/tmp/wivrn-timings.csv wivrn-server
+```
+CSV and Perfetto are independent; set either, neither, or both. See
+[CSV ↔ Perfetto](#csv--perfetto-alignment).
+
+## HMD profiling
+
+```bash
+adb shell setprop persist.traced.enable 1
+```
+Then use the [Perfetto UI](https://ui.perfetto.dev/) → connect device. See the
+[Android tracing quickstart](https://perfetto.dev/docs/quickstart/android-tracing).
+
+## Backends — `WIVRN_TRACING`
+
+`WIVRN_TRACING` enables tracing and selects how it is collected. Unset/`off` = disabled; unknown
+values warn and stay off.
+
+| Value              | Behavior |
+| ------------------ | -------- |
+| `system`           | Connect to an external `traced` daemon; an external consumer owns the session. The only mode that gives a unified system-wide capture. Use `tools/perfetto/wivrn_capture.py`. |
+| `inprocess`        | wivrn-server runs the service itself and writes a self-contained `.pftrace`. No daemon, no sudo. WiVRn track events only — no ftrace/system data. |
+| `system+inprocess` | Both, independently → two separate traces (not merged). |
+
+In-process options:
+- `WIVRN_TRACING_FILE` — output path. A trace is written on **each client disconnect** and at server
+  shutdown, so one run can produce several files. Unset → timestamped `wivrn-<timestamp>-<pid>.pftrace`
+  in cwd. An explicit value expands `%t` (timestamp)/`%p` (pid) and gains a `-2`, `-3`, … suffix
+  rather than overwriting. Written from the per-session child, so use an absolute path if unsure;
+  parent dirs are created.
+- `WIVRN_TRACING_BUFFER_KB` — ring buffer size in KiB (default `65536`).
+
+A headset must be streaming during the capture window. Confirm setup by raising Monado's log level
+(default `warn` hides the lines):
+```bash
+XRT_LOG=info WIVRN_TRACING=system wivrn-server
+# → wivrn::trace: producer socket /run/user/<uid>/wivrn-perfetto-producer.sock
+# → wivrn::trace: init complete
+```
+
+> Perfetto's default socket is root-only, so WiVRn (and the wrapper) use
+> `$XDG_RUNTIME_DIR/wivrn-perfetto-{producer,consumer}.sock` (fallback `/tmp`). Override via
+> `PERFETTO_PRODUCER_SOCK_NAME`.
+
+
+## Capturing — `wivrn_capture.py`
+
+A wrapper around `tracebox` (downloaded on first run): starts the tracing services, waits for the
+wivrn-server producer, captures, and tears down. If no producer connects within 20s it **aborts**
+rather than write an empty trace — start wivrn-server with `WIVRN_TRACING=system`, stream, retry.
+
+```bash
+tools/perfetto/wivrn_capture.py                       # default: WiVRn-only, no ftrace, no sudo
+tools/perfetto/wivrn_capture.py --full                # + every producer + broad ftrace (needs sudo)
+tools/perfetto/wivrn_capture.py --full --cpu-only     # full producers, ftrace stripped (no sudo)
+tools/perfetto/wivrn_capture.py --duration-ms 5000 -o my.pftrace
+tools/perfetto/wivrn_capture.py -c custom.pbtx        # any Perfetto config
+```
+
+`--cpu-only` strips any `linux.ftrace` block from the config. The wrapper auto-detects ftrace in the
+config and only then starts `traced_probes` + uses `sudo`.
+
+### Default vs `--full`
+
+| Aspect | default (`wivrn_trace_cfg.pbtx`) | `--full` (`wivrn_full_trace_cfg.pbtx`) |
+| ------ | -------------------------------- | -------------------------------------- |
+| track_event | WiVRn categories only       | every category, every producer |
+| ftrace | none                             | sched + freq + idle + IRQs + vblank |
+| Buffers | 64 MiB (track_event)            | 64 MiB track_event **+** 256 MiB system |
+| sudo   | no                               | yes (ftrace) |
+| Use for | encoder-pipeline questions      | "something outside WiVRn is doing X" |
+
+`--full` isolates `track_event` in its own buffer so the high-rate ftrace stream can't evict the
+WiVRn/Monado slices, and defaults to a 5 s window (override with `--duration-ms`). If a trace is too
+big to load, drop the `irq/*` events from a copy of the config.
+
+### Manual capture
+
+Drive Perfetto yourself against WiVRn's default socket pair:
+```bash
+SOCK_DIR="${XDG_RUNTIME_DIR:-/tmp}"
+export PERFETTO_PRODUCER_SOCK_NAME="$SOCK_DIR/wivrn-perfetto-producer.sock"
+export PERFETTO_CONSUMER_SOCK_NAME="$SOCK_DIR/wivrn-perfetto-consumer.sock"
+
+traced --background                                    # add sudo + traced_probes for ftrace configs
+WIVRN_TRACING=system wivrn-server                      # in another shell, streaming
+perfetto -c tools/perfetto/wivrn_trace_cfg.pbtx --txt -o trace.pftrace
+```
+For ftrace configs, run `traced`/`traced_probes` under `sudo -E` and add
+`--set-socket-permissions "$USER:0660:$USER:0660"` to `traced`.
+
+### Summaries
+
+`tools/perfetto/pftrace_summary.py` prints per-slice stats, or a side-by-side diff for two traces:
+```bash
+tools/perfetto/pftrace_summary.py nvenc.pftrace                 # one trace
+tools/perfetto/pftrace_summary.py nvenc.pftrace vulkan.pftrace  # diff with Δmean
+```
+
+## Tracing Monado too
+
+Monado's own `u_trace` (built on percetto) can join the same session. Build with
+`WIVRN_TRACE_MONADO=ON` (requires `WIVRN_USE_PERFETTO=ON`):
+```bash
+cmake --preset server-tracing-monado
+cmake --build build-server-tracing-monado
+
+WIVRN_TRACING=system wivrn-server     # also sets Monado's XRT_TRACING for you
+tools/perfetto/wivrn_capture.py --full
+```
+
+- **System backend + `--full` only.** Monado's percetto can't write in-process, and the default
+  config allowlists only `wivrn_*` categories.
+- **percetto must be available** at configure time (pkg-config, or `-DPercetto_ROOT_DIR=<path>`);
+  Monado does not fetch it. Missing → `XRT_HAVE_PERCETTO specified but not available`.
+- Links a second Perfetto SDK copy (percetto's); both feed one `traced`. Leave off unless you need
+  Monado spans.
+
+## Adding instrumentation
+
+Include `utils/wivrn_trace.h`; no `#ifdef` at call sites.
+
+```cpp
+// CPU span (RAII; ends at scope exit)
+wivrn::trace::scope s(wivrn::trace::cpu_track::encoder, stream_idx, frame_index, "my_span");
+
+// CPU instant
+wivrn::trace::cpu_instant(wivrn::trace::cpu_track::compositor, "my_event", 0, 0);
+
+// CPU begin/end pair (lifetime spans multiple calls)
+wivrn::trace::cpu_begin(wivrn::trace::cpu_track::network, stream_idx, frame_idx, "Send");
+wivrn::trace::cpu_end  (wivrn::trace::cpu_track::network, stream_idx, frame_idx, "Send");
+
+// GPU span: bracket work with a gpu_timestamp_pool's cmd_begin/cmd_end, emit after the fence
+if (auto s = pool.collect(slot))
+        wivrn::trace::gpu_slice(wivrn::trace::gpu_track::nvenc_copy,
+                                "my_gpu_span", s->begin_ns, s->end_ns, s->frame_index, stream);
+```
+
+GPU begin/end come from the pool already in host-monotonic ns (via `VK_EXT_calibrated_timestamps`).
+A disabled pool (no timestamp bits, or `WIVRN_TRACING` unset) returns `nullopt`. `name` must outlive
+the emit (a literal is fine). Place the span at the operation's semantic start, not the function top.
+
+| `cpu_track` | category | track |
+| ----------- | -------- | ----- |
+| `encoder` | `wivrn_encoder` | WiVRn encoder |
+| `compositor` | `wivrn_compositor` | WiVRn compositor |
+| `network` | `wivrn_network` | WiVRn network |
+
+## CSV ↔ Perfetto alignment
+
+`WIVRN_DUMP_TIMINGS` (CSV) and the `wivrn_feedback` track read the **same** `dump_time` stream, clock
+(`CLOCK_MONOTONIC`), and event names (`wake_up`, `encode_begin`, `send_end`, `display`, …) — so
+`grep encode_begin` over the CSV and `name = "encode_begin"` in Perfetto select the same frames at
+the same timestamps. Perfetto additionally carries the CPU/GPU spans the CSV does not. Use the CSV
+for offline scripting, the trace for a visual timeline.
--- a/patches/monado/0009-ipc-join-client-threads-before-teardown.patch
+++ b/patches/monado/0009-ipc-join-client-threads-before-teardown.patch
@@ -1,0 +1,62 @@
+From a149552933bda15ffab81460f38dbe90abd0dfe0 Mon Sep 17 00:00:00 2001
+From: no name <git-am@invalid>
+Date: Sun, 28 Jun 2026 11:36:08 -0600
+Subject: [PATCH] ipc: join client threads before teardown
+
+On shutdown main_loop() returns as soon as s->running is cleared, but the
+per-client threads may still be in common_shutdown() destroying their
+compositor and dereferencing s->xsysd/s->xso. teardown_all() then frees
+those resources concurrently, so a client thread can lock a mutex the main
+thread has already destroyed (assert om->initialized in
+multi_compositor_destroy). Join the client threads first.
+---
+ src/xrt/ipc/server/ipc_server_process.c | 28 +++++++++++++++++++++++++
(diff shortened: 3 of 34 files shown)


server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,213 @@
...
+} // namespace
+
+void init()
+{
+	std::call_once(init_flag, [] {
+		if (!env_truthy(std::getenv("WIVRN_TRACING")))
+		{
+			U_LOG_I("wivrn::trace: WIVRN_TRACING not set, percetto tracing disabled");
+			return;
+		}
```
xytovl: You can use `u_debug.h` from monado, either through `DEBUG_GET_ONCE_BOOL_OPTION` or `debug_get_bool_option`

server/utils/gpu_timestamp_pool.h
```diff
@@ -0,0 +1,47 @@
...
+// A begin/end GPU timestamp pair per encoder slot, used to measure GPU-side
+// encode/copy durations for tracing. Every method is a no-op unless built with
+// WIVRN_USE_PERFETTO and the queue family supports timestamps.
+class gpu_timestamp_pool
+{
+	vk::raii::QueryPool pool = nullptr;
+	float period_ns = 0.f;
+
+public:
+	void init(vk_bundle & vk, uint32_t queue_family_index, uint32_t slot_count, const std::string & name);
```
xytovl: Any reason for an `init` instead of a constructor?

server/utils/gpu_timestamp_pool.h
```diff
@@ -0,0 +1,47 @@
...
+
+	void cmd_reset(const vk::raii::CommandBuffer & cmd, uint8_t slot) const;
+	void cmd_write(const vk::raii::CommandBuffer & cmd, uint8_t slot, uint32_t index, vk::PipelineStageFlagBits2 stage) const;
+
+	// GPU duration for slot, or nullopt when disabled / unsupported / not ready.
+	std::optional<int64_t> duration_ns(uint8_t slot);
```
xytovl: With [calibrated timestamps](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#calibrated-timestamps) it is possible to get absolute begin/end times.
See `vk_convert_timestamps_to_host_ns` in `auxiliary/vk/vk_helpers.h`

server/utils/wivrn_trace.h
```diff
@@ -0,0 +1,102 @@
...
+#define I_WIVRN_TRACE_UID(prefix) I_WIVRN_TRACE_CAT2(prefix, __LINE__)
+
+// domain: unqualified cpu_track enumerator (encoder / compositor / network).
+#define WIVRN_TRACE_SCOPE(domain, stream, frame, name)               \
+	::wivrn::trace::scope I_WIVRN_TRACE_UID(wivrn_trace_scope_)( \
+	        ::wivrn::trace::cpu_track::domain, (stream), (frame), (name))
+
+#define WIVRN_TRACE_INSTANT(domain, stream, frame, name) \
+	::wivrn::trace::cpu_instant(::wivrn::trace::cpu_track::domain, (name), (frame), (stream))
+
```
xytovl: Do we need the macros? I think we're fine calling the functions/constructors directly.

server/CMakeLists.txt
```diff
@@ -171,6 +173,26 @@ if (WIVRN_BUILD_SERVER)
...
+		elseif(PERCETTO_FOUND)
+			target_link_libraries(wivrn-server PRIVATE PkgConfig::PERCETTO)
+		else()
+			FetchContent_Declare(percetto
+				GIT_REPOSITORY https://github.com/olvaffe/percetto.git
+				GIT_TAG        8f4968309af74bd9d4f29b8859dad167ec37e6b8 # main as of 2026-05
+			)
+			FetchContent_MakeAvailable(percetto)
+			target_link_libraries(wivrn-server PRIVATE percetto)
+		endif()
```
xytovl: The fetchcontent fallback should not be automatic

server/utils/gpu_timestamp_pool.h
```diff
@@ -0,0 +1,47 @@
...
+class gpu_timestamp_pool
+{
+	vk::raii::QueryPool pool = nullptr;
+	float period_ns = 0.f;
+
+public:
+	void init(vk_bundle & vk, uint32_t queue_family_index, uint32_t slot_count, const std::string & name);
+
+	void cmd_reset(const vk::raii::CommandBuffer & cmd, uint8_t slot) const;
+	void cmd_write(const vk::raii::CommandBuffer & cmd, uint8_t slot, uint32_t index, vk::PipelineStageFlagBits2 stage) const;
```
xytovl: I would replace the public methods with `begin`/`end`, we will always do reset+write(0), then write(1).

server/encoder/video_encoder.cpp
```diff
@@ -296,6 +299,7 @@ void video_encoder::encode(wivrn_session & cnx,
 
 void video_encoder::SendData(std::span<uint8_t> data, bool end_of_frame, bool control)
 {
+	WIVRN_TRACE_SCOPE(network, stream_idx, shard.frame_idx, "SendData");
```
xytovl: On x264 this will create one scope per NAL, if we want it per frame, the scope would have to be opened/closed depending on the "end of frame" value.

server/encoder/ffmpeg/video_encoder_va.h
```diff
@@ -43,10 +44,14 @@ class video_encoder_va : public video_encoder_ffmpeg
 		vk::raii::Image luma = nullptr;
 		vk::raii::Image chroma = nullptr;
 		std::vector<vk::raii::DeviceMemory> mem;
+		uint64_t pending_frame_index = 0;
```
xytovl: This had to be added to each encoder just for the timestamp pool, we could store it in the gpu_timestamp_pool object, when resetting the query.

server/encoder/video_encoder_vulkan.cpp
xytovl, line 1: There's 3 additional optional vulkan operations:
* GPU -> GPU image copy (NVIDIA)
* GPU -> CPU buffer copy (dGPU)
* GPU -> CPU buffer copy (if the first copy was too small)

With calibrated timestamps this may be less important as the image copy will appear as a gap before encoding, and the other ones as a gap before network transmission.

server/encoder/video_encoder_vulkan.cpp
```diff
@@ -1043,6 +1053,17 @@ std::optional<wivrn::video_encoder::data> wivrn::video_encoder_vulkan::encode(ui
...
 
+	if (auto encode_ns = ts_pool.duration_ns(encode_slot))
+	{
+		const int64_t gpu_begin = slot_item.encode_submit_ns;
+		wivrn::trace::gpu_slice(wivrn::trace::gpu_track::vulkan_encode,
+		                        "encodeVideoKHR",
+		                        gpu_begin,
+		                        gpu_begin + *encode_ns,
+		                        slot_item.pending_frame_index,
+		                        stream_idx);
```
xytovl, line 1090: It would be nice to find a way to send this to the headset in timing_info.

damaestro: Very possible, but I'd like to defer this to another PR.

server/utils/gpu_timestamp_pool.h
damaestro: Nope. I'll change it.

server/utils/wivrn_trace.h
damaestro: Sounds good!

server/CMakeLists.txt
damaestro: K, agreed. Thanks for confirming the style you'd like.

server/utils/gpu_timestamp_pool.cpp
```diff
@@ -0,0 +1,104 @@
...
+	if (vk.monado_vk == nullptr)
+	{
+		U_LOG_W("gpu_timestamp_pool: monado vk_bundle unavailable, GPU spans disabled");
+		return;
+	}
+
+	pool = vk.device.createQueryPool(vk::QueryPoolCreateInfo{
+	        .queryType = vk::QueryType::eTimestamp,
+	        .queryCount = slot_count * 2,
+	});
```
xytovl, line 57: As we already handle null pool when tracing is disabled at compile time, we should also do that when it's disabled at run time

server/utils/gpu_timestamp_pool.cpp
```diff
@@ -0,0 +1,104 @@
...
+{
+#ifdef WIVRN_USE_PERFETTO
+	auto qprops = vk.physical_device.getQueueFamilyProperties();
+	if (queue_family_index >= qprops.size() or qprops[queue_family_index].timestampValidBits == 0)
+		return;
+	if (vk.monado_vk == nullptr)
+	{
+		U_LOG_W("gpu_timestamp_pool: monado vk_bundle unavailable, GPU spans disabled");
+		return;
+	}
```
xytovl: This could be an assertion

server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,238 @@
...
+PERCETTO_TRACK_DEFINE(wivrn_network_cpu, PERCETTO_TRACK_EVENTS);
+
+DEBUG_GET_ONCE_BOOL_OPTION(wivrn_tracing, "WIVRN_TRACING", false)
+
+namespace wivrn::trace
+{
+namespace
+{
+std::once_flag init_flag;
+std::atomic<bool> init_ok{false};
```
xytovl: I don't think we need something fancy here, maybe a static bool in debug mode to ensure that init is done only once and that it's called before using traces.

We know that it will be done before streaming actually starts so there's no concurrency issue.

server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,238 @@
...
+	percetto_event_with_args(cat, mask, nullptr, PERCETTO_EVENT_END, trk, 0, (uint64_t)end_ns);
+}
+} // namespace
+
+void init()
+{
+	std::call_once(init_flag, [] {
+		if (!debug_get_bool_option_wivrn_tracing())
+		{
+			U_LOG_I("wivrn::trace: WIVRN_TRACING not set, percetto tracing disabled");
```
xytovl: disabled tracing is the default, I think it would be better to log when tracing is enabled instead

xytovl: actually there's a log when trace init is complete.

The disabled log should be a debug one.

server/utils/wivrn_vk_bundle.h
```diff
@@ -105,6 +109,12 @@ struct vk_bundle
 	std::vector<const char *> instance_extensions;
 	std::vector<const char *> device_extensions;
 
+	// Set by the compositor after it initializes monado's vk_bundle from
+	// our Vulkan handles (vk_init_from_given). Null when running outside the
+	// compositor; consumers that need calibrated timestamps treat null as
+	// "calibration unavailable".
+	::vk_bundle * monado_vk = nullptr;
```
xytovl: I don't have a solution right now, but the pointer that is set after construction is not very nice.

server/encoder/video_encoder_vulkan.cpp
damaestro: Added, please make sure it's what you expected.

server/utils/wivrn_trace.cpp
damaestro: I recall this had something to do with some odd interaction I experienced between the wivrn and the monado implementation of percetto.

One of the cool things we will be able to do is trace the full stack, including Monado, Mesa, kernel, etc. However, it matters who actually inits things. It might have been a transient issue early in the development, so I've switched this for now.

server/utils/wivrn_vk_bundle.h
damaestro: I hope this new option is more reasonable.

server/encoder/video_encoder_vulkan.cpp
xytovl: I didn't think about the different queues, so you had to make different pools, the diff looks fine.

server/utils/wivrn_trace.cpp
xytovl: The atomic doesn't make a difference, it's independent from `u_trace_marker_init` in Monado.

I don't know much about Monado's tracing, if we want to use both it may be better to use the same compile option and same environment variable. In that case, the implementation may just be a wrapper around Monado's tracing.

damaestro:
Okay, looks like it's all working:
Screenshot From 2026-06-07 16-02-56

xytovl:
I'll have to test it on my machine but it should be good now.

I would like to make a release soon, so I'll keep this pull request on hold.

xytovl:
I was trying to package percetto for Gentoo so I could test, and I see that it depends on perfetto-sdk.

Can't we use perfetto-sdk directly? We don't need a C layer. The sdk is a pair of files, so a WIVRN_USE_SYSTEM_PERFETTO=OFF should not be hard to implement and could be used in offline builds too. percetto uses a meson wrap so there's 2 layers to fetch.

damaestro:
> Can't we use perfetto-sdk directly?

Seems to work fine:

Screenshot From 2026-06-19 18-04-20


> We don't need a C layer.

Monado uses percetto, so I thought using the same implementation might be helpful. However, it's not necessary.

server/utils/gpu_timestamp_pool.cpp
```diff
@@ -0,0 +1,107 @@
...
+	if (not *pool)
+		return std::nullopt;
+	::vk_bundle * cal = wivrn::trace::calibration_source();
+	assert(cal != nullptr);
+	auto [res, ticks] = pool.getResults<uint64_t>(
+	        slot * 2, 2, 2 * sizeof(uint64_t), sizeof(uint64_t), vk::QueryResultFlagBits::e64 | vk::QueryResultFlagBits::eWait);
+	if (res != vk::Result::eSuccess or ticks[1] < ticks[0])
+		return std::nullopt;
+	std::array<uint64_t, 2> host_ns = {ticks[0], ticks[1]};
+	if (vk_convert_timestamps_to_host_ns(cal, 2, host_ns.data()) != VK_SUCCESS)
```
xytovl: You can pass ticks.size(), ticks.data() to the convert function

xytovl: Or keep the array, but pass it to `pool.getResults` so we skip the vector allocation

```cpp
        std::array<uint64_t, 2> times;                                                                                                            
        auto res = pool.getResults(slot * 2, 2, sizeof(uint64_t), times.data(), sizeof(uint64_t), vk::QueryResultFlagBits::e64 | vk::QueryResultFlagBits::eWait);
        if (res != vk::Result::eSuccess or times[1] < times[0])  
                return std::nullopt;                                                                                                                                                       
        if (vk_convert_timestamps_to_host_ns(cal, times.size(), times.data()) != VK_SUCCESS)
                return std::nullopt;
```

server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,278 @@
...
+		U_LOG_D("wivrn::trace: WIVRN_TRACING not set, perfetto tracing disabled");
+		return;
+	}
+
+	const auto sock = default_producer_sock_path();
+	if (setenv("PERFETTO_PRODUCER_SOCK_NAME", sock.c_str(), 0) == 0)
+		U_LOG_I("wivrn::trace: producer socket %s", std::getenv("PERFETTO_PRODUCER_SOCK_NAME"));
+
+	perfetto::TracingInitArgs args;
+	args.backends |= perfetto::kSystemBackend;
```
xytovl: Would the InProcess backend be easier to deploy? Tracing would be self-contained.

damaestro: I started with the specific goal of being able to capture wivrn down to the kernel. This would include wivrn, monado, mesa, kernel, etc. This is possible and actually works pretty well.

```
tools/perfetto/wivrn_capture.py --full --duration-ms 5000
```
Screenshot From 2026-06-20 10-24-57
Screenshot From 2026-06-20 10-23-52
Screenshot From 2026-06-20 10-27-21

tools/perfetto/wivrn_capture.py
```diff
@@ -0,0 +1,344 @@
...
+                elif isinstance(e, str):
+                    names.append(e)
+            unique = list(dict.fromkeys(n for n in names if n))
+            if len(unique) == 1:
+                return unique[0]
+            if unique:
+                return ",".join(unique)
+        return "auto"
+    except Exception:
+        return "auto"
```
xytovl: I don't want to maintain one more location where the configuration is read

damaestro: This is only used for UX on file naming for the single command capture setup. It's really nice to have files named according to the configuration. I could drop it, but I found trying to remember which capture was with which encoder was a pain.

server/utils/wivrn_trace.cpp
xytovl: I see, but that adds a script that is way too large just to setup tracing. I thought we could enable tracing with a single environment variable.

This feature will be rarely used, so everything that is not checked at compile time will bitrot and will require fixing every time someone will want to use it.

damaestro: I've added the ability to also use in process tracing.

damaestro:
@xytovl re: making sure this code is regularly built, does this mean I should add a variant build to the action pipeline? Or how would you like that handled?

Maybe we just always build with it enabled? There is robust runtime disabling that should result in full noop execution if it's disabled.

xytovl:
We could enable it in CI, I don't think it's good to have by default, as the perfetto sdk is too large to be copied in external and it's not something that will be useful for most people.

My concern for bitrot is also about the surrounding scripts/config/docs, it's quite a lot to maintain

server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,410 @@
...
+}
+
+// Per-user runtime directory, used to anchor the producer socket. Falls back to /tmp
+// when XDG_RUNTIME_DIR is unset.
+std::string runtime_dir()
+{
+	if (const char * xdg = std::getenv("XDG_RUNTIME_DIR"); xdg and *xdg)
+		return xdg;
+	return "/tmp";
+}
```
xytovl: This should be in `common/utils/xdg_base_directory.cpp`

server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,410 @@
...
+	const std::string s = v;
+	if (s == "system")
+		return backend_mode::system;
+	if (s == "inprocess")
+		return backend_mode::inprocess;
+	if (s == "both")
+		return backend_mode::both;
+	U_LOG_W("wivrn::trace: unknown WIVRN_TRACING_BACKEND '%s', falling back to system", v);
+	return backend_mode::system;
+}
```
xytovl, line 99: It feels like this could be merged with `WIVRN_TRACING`:
system/inprocess/system+inprocess to enable it and select the mode. Invalid values emit a warning and don't enable tracing.

server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,410 @@
...
+// cwd is not the launcher's. An absolute path is used verbatim.
+std::string resolve_inproc_file_path()
+{
+	const char * file = std::getenv("WIVRN_TRACING_FILE");
+	if (!file or !*file)
+		return home_dir() + "/wivrn.pftrace";
+	if (file[0] == '/')
+		return file;
+	return home_dir() + "/" + file;
+}
```
xytovl: I don't think we change cwd, relative path should be safe to use.
I would expect the default to be in the current directory rather than home

server/utils/wivrn_trace.cpp
```diff
@@ -0,0 +1,410 @@
...
+void start_inprocess_session()
+{
+	g_inproc_file = resolve_inproc_file_path();
+
+	uint32_t buffer_kb = 65536;
+	if (const char * b = std::getenv("WIVRN_TRACING_BUFFER_KB"); b and *b)
+	{
+		if (const long parsed = std::strtol(b, nullptr, 10); parsed > 0)
+			buffer_kb = static_cast<uint32_t>(parsed);
+	}
```
xytovl, line 201: Why does this need to be configurable?

server/utils/wivrn_trace.cpp
damaestro: When I ran from envision, this caused it to be in `$HOME`. It took me a while to find it at first, so I figured explicit would be good. This should also work in a flatpak.

server/utils/wivrn_trace.cpp
damaestro: As discussed in Slack, this is due to the in-process recording design being:

* Constantly record data and stuff it into this buffer
* Write out the current buffer at shutdown time (producing the trace file)

When using the system profiling, you are able to tap into the tracing at any point and record based on trace settings (time, buffer, filters, etc.)

Making this configurable with a sane default allows the needed flexibility for capture window size when using the in-process model.

damaestro:
I'm currently debugging an issue with the `WIVRN_TRACING=inprocess` path but I don't think I'll finish the fix tonight.

docs/profiling.md
```diff
@@ -0,0 +1,207 @@
...
+# → wivrn::trace: init complete
+```
+
+> Perfetto's default socket is root-only, so WiVRn uses
+> `$XDG_RUNTIME_DIR/wivrn-perfetto-{producer,consumer}.sock` (fallback `/tmp`); the wrapper uses the
+> same pair. Override via `PERFETTO_PRODUCER_SOCK_NAME`.
+
+## What WiVRn records
+
+| Category (`wivrn_*`) | Spans / instants | Track |
```
damaestro: Should I make this more generic so we can add/remove things into the category and not have to update docs? Maybe describe what events should go into each category vs listing them out?

xytovl: I don't think it's even needed to be able to list/describe them. There's already a description where they are defined.

damaestro:
@xytovl Everything appears to be working well. The addition of the standalone inprocess is actually rather useful for day to day stuff. One trace file per connection lifecycle worked out pretty well. The system tracing is also showing some interesting things such as the pacing logic and shows what kind of headroom is present, etc. Cool stuff.

Screenshot From 2026-06-27 21-06-29
Screenshot From 2026-06-27 20-11-11
Screenshot From 2026-06-27 20-10-39

server/encoder/ffmpeg/video_encoder_va.cpp
```diff
@@ -430,9 +432,12 @@ video_encoder_va::video_encoder_va(wivrn::vk_bundle & vk,
 			vk.device.bindImageMemory2(bind_info);
 		}
 	}
+
+	// GPU timestamps for the Vulkan→VAAPI luma/chroma copy.
+	ts_pool = gpu_timestamp_pool(vk, vk.queue.family_index, num_slots, std::format("va encoder {} timestamp pool", stream_idx));
```
xytovl: If the comment is useful, you can use that as the timestamp pool name

In general, comment on simple code don't really help, "GPU timestamps" is already in the function and name, what is adds is mainly that we measure the pixel copy operation.

server/main.cpp
```diff
@@ -953,7 +954,9 @@ int inner_main(int argc, char * argv[], bool show_instructions)
 		active_runtime::cleanup_openxr();
 	listen_socket = create_listen_socket();
 
-	u_trace_marker_init();
+	// wivrn::trace::init runs in the per-session child via xrt_instance_create
+	// (target_instance_wivrn.cpp): the Perfetto producer is not fork-safe, so it
+	// must connect from the forked child rather than the daemon parent.
```
xytovl: Once the code is not there, nobody will wonder why we are not initializing traces here. Drop the comments, it would only confuse readers

server/target_instance_wivrn.cpp
```diff
@@ -99,7 +100,9 @@ void instance::set_ipc_server(ipc_server * server)
 xrt_result_t
 xrt_instance_create(struct xrt_instance_info * ii, struct xrt_instance ** out_xinst)
 {
+	wivrn::trace::propagate_to_monado();
 	u_trace_marker_init();
+	wivrn::trace::init();
```
xytovl: It looks safer to have `propagate_to_monado` and `u_trace_marker_init` merged into `wivrn::trace::init`

CMakePresets.json
```diff
@@ -25,7 +26,7 @@
 				"WIVRN_USE_NVENC": "ON",
 				"WIVRN_USE_SYSTEMD": "ON",
 				"WIVRN_USE_SYSTEM_BOOST": "ON",
-				"WIVRN_USE_PULSEAUDIO": "ON",
+				"WIVRN_USE_PULSEAUDIO": "OFF",
```
xytovl: Until pulseaudio backend is in the code, there should be at least one CI path that builds it

patches/monado/0009-ipc-join-client-threads-before-teardown.patch
xytovl, line 1: This looks like you could make an upstream merge request for this. Don't remove it from here, we'll do that when it's upstreamed and we update Monado.

damaestro: Agreed this is an existing shutdown race bug. I just happened to start hitting it because we are flushing the ring buffer on shutdown and that takes a moment. Agreed this needs to upstream.

CMakePresets.json
damaestro: Good call. I'll work on removing this in another PR.

xytovl:
It looks fine, I'll experiment a little and then it's good for merge.

PR #985, opened 2026-06-19
README: use friendlier log locations

hypevhs:
I talked to a few people who

1. Had no idea what the XDG Directory Spec vars were, and were plugging them into the file browser outright
2. Didn't see the dedicated dashboard button for logs

So, use the most likely location first and let the fallback be a fallback. And move "Or," to the beginning which should hint that they've skipped over the easiest option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,9 +169,10 @@ If the server list is empty in the headset app:
 
 <details><summary>How do I see server logs when using the dashboard?</summary>
 
-- Click Troubleshoot > Open server logs, or
-- Navigate to `${XDG_STATE_HOME}/wivrn/wivrn-dashboard` (with fallback to `${HOME}/.local/state` for `${XDG_STATE_HOME}`, or
-- For flatpak, navigate to `${HOME}/.var/app/io.github.wivrn.wivrn/.local/state/wivrn/wivrn-dashboard`.</details>
+- Click **Troubleshoot > Open server logs**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -165,9 +165,9 @@ If the server list is empty in the headset app:
 
 <details><summary>How do I see server logs when using the dashboard?</summary>
 
-- Click Troubleshoot > Open server logs, or
-- Navigate to `${XDG_STATE_HOME}/wivrn/wivrn-dashboard` (with fallback to `${HOME}/.local/state` for `${XDG_STATE_HOME}`, or
-- For flatpak, navigate to `${HOME}/.var/app/io.github.wivrn.wivrn/.local/state/wivrn/wivrn-dashboard`.</details>
+- Click **Troubleshoot > Open server logs**
+- Or, navigate to `~/.local/state/wivrn/wivrn-dashboard` (the `~/.local/state` part may be different if `${XDG_STATE_HOME}` is set)
```
xytovl: `${XDG_STATE_HOME}/wivrn/wivrn-dashboard` is the primary location where the dashboard writes logs. It should remain written like this in docs.
There is no reason to rephrase it to be vague ("may be different"), you can instead add something like "On common setups, the dashboard writes logs in `~/.local/state/wivrn/wivrn-dashboard`" but not replace the proper location.

hypevhs:
How's this?

PR #986, opened 2026-06-22
Add gamepad forwarding from the headset to the server

dniminenn:
## Summary

Adds optional forwarding of a gamepad paired to the headset through to the server. Useful when the controller is in range of the headset but not the server, such as a DualSense paired to a Quest over Bluetooth while streaming to a PC. Off by default.

## Commits

- **Add gamepad forwarding through OpenXR**: the interaction profile data model, no extra config or UI.
- **Mirror forwarded input devices to uinput**: the uinput add-on and its configuration.

## How it works

The headset reads the gamepad through the OpenXR `/interaction_profiles/microsoft/xbox_controller` profile and forwards it on the existing input rails: gamepad `device_id`s in the inputs packet, plus a gamepad entry in the interaction profile array sent from the tracking thread. The switch to `none` signals disconnection.

The server exposes a native OpenXR gamepad xdev at `/user/gamepad`, connected while the headset reports a gamepad interaction profile. This needs no permission and is always available.

It can also mirror the gamepad to a `uinput` device with the standard Xbox layout (`BTN_GAMEPAD` plus a Microsoft VID/PID), so SDL and Steam Input pick it up. The uinput mirror is gated by the `hid-forwarding` server option, which also covers forwarded keyboard and mouse.

Rumble is forwarded back to the headset and applied through OpenXR haptics, best effort.

Only the digested controller state is forwarded, so device specific features such as gyro, adaptive triggers and the touchpad are not available.

## Configuration

The headset chooses which devices to forward (keyboard, mouse, gamepad). The `hid-forwarding` server option gates the uinput mirror only; the OpenXR gamepad is always exposed. With `hid-forwarding` off, the headset reflects that forwarded keyboard and mouse will not reach the server.

The uinput mirror requires the `uinput` kernel module and write access to `/dev/uinput`.

## Testing

Tested on Arch Linux, Quest 3, DualSense over Bluetooth. Buttons, both sticks, analog triggers and d-pad map correctly. Played Project Wingman with no noticeable added latency. Confirmed the same input reaches both the `uinput` device and `/user/gamepad`.

### DIFF
--- a/client/android/hid.cpp
+++ b/client/android/hid.cpp
@@ -249,10 +249,13 @@ bool android_hid::input_handler::handle_input(scene * current_scene, AInputEvent
 {
 	const int32_t type = AInputEvent_getType(event);
 	const int32_t src = AInputEvent_getSource(event);
-	const int32_t dev = AInputEvent_getDeviceId(event);
 
 	if (type == AINPUT_EVENT_TYPE_KEY)
 	{
+		// Gamepad input is read through the OpenXR microsoft/xbox_controller profile, not here.
+		if ((src & AINPUT_SOURCE_GAMEPAD) == AINPUT_SOURCE_GAMEPAD)
+			return false;
+
 		int32_t flags = AKeyEvent_getFlags(event);
 		if ((flags & AKEY_EVENT_FLAG_SOFT_KEYBOARD) != 0)
 			return false; // ignore soft keyboards
@@ -288,6 +291,10 @@ bool android_hid::input_handler::handle_input(scene * current_scene, AInputEvent
 
 	if (type == AINPUT_EVENT_TYPE_MOTION)
 	{
+		// Gamepad axes are read through the OpenXR microsoft/xbox_controller profile, not here.
+		if ((src & AINPUT_SOURCE_JOYSTICK) == AINPUT_SOURCE_JOYSTICK)
+			return false;
+
 		// don't care about absolute mouse due to mapping difficulties
 		if ((src & AINPUT_SOURCE_MOUSE_RELATIVE) == 0)
 			return false;
--- a/client/application.cpp
+++ b/client/application.cpp
@@ -39,6 +39,7 @@
 #include <boost/locale.hpp>
 #include <boost/url/parse.hpp>
 #include <chrono>
+#include <cstring>
 #include <ctype.h>
 #include <exception>
 #include <magic_enum.hpp>
@@ -602,6 +603,35 @@ static std::vector<interaction_profile> interaction_profiles{
                         "/user/eyes_ext/input/gaze_ext/pose",
                 },
         },
+        interaction_profile{
+                .profile_name = "/interaction_profiles/microsoft/xbox_controller",
+                .input_sources = {
+                        "/user/gamepad/input/menu/click",
+                        "/user/gamepad/input/view/click",
+                        "/user/gamepad/input/a/click",
+                        "/user/gamepad/input/b/click",
+                        "/user/gamepad/input/x/click",
+                        "/user/gamepad/input/y/click",
+                        "/user/gamepad/input/dpad_up/click",
+                        "/user/gamepad/input/dpad_down/click",
+                        "/user/gamepad/input/dpad_left/click",
+                        "/user/gamepad/input/dpad_right/click",
+                        "/user/gamepad/input/shoulder_left/click",
+                        "/user/gamepad/input/shoulder_right/click",
+                        "/user/gamepad/input/thumbstick_left/click",
+                        "/user/gamepad/input/thumbstick_right/click",
+                        "/user/gamepad/input/trigger_left/value",
+                        "/user/gamepad/input/trigger_right/value",
+                        "/user/gamepad/input/thumbstick_left/x",
+                        "/user/gamepad/input/thumbstick_left/y",
+                        "/user/gamepad/input/thumbstick_right/x",
+                        "/user/gamepad/input/thumbstick_right/y",
+                        "/user/gamepad/output/haptic_left",
+                        "/user/gamepad/output/haptic_right",
+                        "/user/gamepad/output/haptic_left_trigger",
+                        "/user/gamepad/output/haptic_right_trigger",
+                },
+        },
 };
 
 static const std::pair<std::string_view, XrActionType> action_suffixes[] =
@@ -634,11 +664,15 @@ static const std::pair<std::string_view, XrActionType> action_suffixes[] =
 		{"/ready_ext", XR_ACTION_TYPE_BOOLEAN_INPUT},
 
 		// Output paths
-		{"/haptic",           XR_ACTION_TYPE_VIBRATION_OUTPUT},
-		{"/haptic_trigger",   XR_ACTION_TYPE_VIBRATION_OUTPUT},
-		{"/haptic_trigger_fb",XR_ACTION_TYPE_VIBRATION_OUTPUT},
-		{"/haptic_thumb",     XR_ACTION_TYPE_VIBRATION_OUTPUT},
-		{"/haptic_thumb_fb",  XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic",              XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_trigger",      XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_trigger_fb",   XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_thumb",        XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_thumb_fb",     XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_left",         XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_right",        XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_left_trigger", XR_ACTION_TYPE_VIBRATION_OUTPUT},
+		{"/haptic_right_trigger",XR_ACTION_TYPE_VIBRATION_OUTPUT},
                 // clang-format on
 };
 
@@ -1717,8 +1751,10 @@ void application::run()
 		while (ALooper_pollOnce(100, nullptr, &events, (void **)&source) >= 0)
 		{
 			// Process this event.
-			if (source != nullptr)
-				source->process(app_info.native_app, source);
+			if (source == nullptr)
+				continue;
+
+			source->process(app_info.native_app, source);
 		}
 
 		if (app_info.native_app->destroyRequested)
--- a/client/scenes/lobby.h
+++ b/client/scenes/lobby.h
@@ -63,6 +63,10 @@ class lobby : public scene_impl<lobby>
 	std::string server_name;
 	bool autoconnect_enabled = true;
 
+	// Last known uinput forwarding state of the connected server, kept after disconnection so the
+	// settings can warn that forwarded keyboard and mouse will not reach the server.
+	std::optional<bool> server_hid_forwarding;
+
 	std::optional<input_profile> input;
 	entt::entity lobby_entity;
 
--- a/client/scenes/lobby_gui.cpp
+++ b/client/scenes/lobby_gui.cpp
@@ -784,6 +784,32 @@ void scenes::lobby::gui_settings()
 		ImGui::Unindent();
 		ImGui::EndDisabled();
 	}
+	{
+		// Virtual devices: a keyboard, mouse or gamepad connected to the headset is
+		// replicated on the server through uinput. The gamepad is also exposed through
+		// OpenXR, whether or not it is mirrored to a virtual device.
+		ImGui::Text("%s", _S("Virtual devices"));
+		if (ImGui::IsItemHovered())
+			imgui_ctx->tooltip(_("Input devices connected to the headset are replicated as virtual devices on the PC.\nThe server must allow forwarded input devices; this is not available when the server runs as Flatpak.\nA gamepad is also usable without a virtual device by applications that read it through OpenXR."));
+		ImGui::Indent();
+		if (ImGui::Checkbox(_S("Keyboard"), &config.forward_keyboard))
+			config.save();
+		imgui_ctx->vibrate_on_hover();
+		if (ImGui::Checkbox(_S("Mouse"), &config.forward_mouse))
+			config.save();
+		imgui_ctx->vibrate_on_hover();
+		if (ImGui::Checkbox(_S("Gamepad"), &config.forward_gamepad))
+			config.save();
+		imgui_ctx->vibrate_on_hover();
+		ImGui::Unindent();
+
+		if (server_hid_forwarding == false and (config.forward_keyboard or config.forward_mouse or config.forward_gamepad))
+		{
+			ImGui::TextColored(constants::style::warn, ICON_FA_TRIANGLE_EXCLAMATION);
+			ImGui::SameLine();
+			ImGui::Text("%s", _S("The server does not allow forwarded input devices"));
+		}
+	}
 	if (system.hand_tracking_supported())
 	{
 		bool enabled = config.check_feature(feature::hand_tracking);
--- a/client/scenes/stream.cpp
+++ b/client/scenes/stream.cpp
@@ -138,6 +138,27 @@ static const std::unordered_map<std::string, device_id> device_ids = {
 	{"/user/hand/right/input/aim_activate_ext/ready_ext",device_id::RIGHT_AIM_ACTIVATE_READY},
 	{"/user/hand/right/input/grasp_ext/value",      device_id::RIGHT_GRASP_VALUE},
 	{"/user/hand/right/input/grasp_ext/ready_ext",  device_id::RIGHT_GRASP_READY},
+
+	{"/user/gamepad/input/menu/click",             device_id::GAMEPAD_MENU_CLICK},
+	{"/user/gamepad/input/view/click",             device_id::GAMEPAD_VIEW_CLICK},
+	{"/user/gamepad/input/a/click",                device_id::GAMEPAD_A_CLICK},
+	{"/user/gamepad/input/b/click",                device_id::GAMEPAD_B_CLICK},
+	{"/user/gamepad/input/x/click",                device_id::GAMEPAD_X_CLICK},
+	{"/user/gamepad/input/y/click",                device_id::GAMEPAD_Y_CLICK},
+	{"/user/gamepad/input/dpad_down/click",        device_id::GAMEPAD_DPAD_DOWN_CLICK},
+	{"/user/gamepad/input/dpad_right/click",       device_id::GAMEPAD_DPAD_RIGHT_CLICK},
+	{"/user/gamepad/input/dpad_up/click",          device_id::GAMEPAD_DPAD_UP_CLICK},
+	{"/user/gamepad/input/dpad_left/click",        device_id::GAMEPAD_DPAD_LEFT_CLICK},
+	{"/user/gamepad/input/shoulder_left/click",    device_id::GAMEPAD_SHOULDER_LEFT_CLICK},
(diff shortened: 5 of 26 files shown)


xytovl:
I am trying to move as much configuration as possible on the headset side.

The HID forwarding on server is present because it requires some permissions and has security implications.
It may be better to rephrase the server side to something like "allow forwarded input devices" and on headset have toggles for mouse, keyboard and gamepad.

You don't have to do the changes, but I'd like to discuss the proper user interface to configure features.

common/wivrn_packets.h
```diff
@@ -39,7 +39,7 @@
 namespace wivrn
 {
 
-static constexpr int protocol_revision = 2;
+static constexpr int protocol_revision = 3;
```
xytovl: Technically not needed, I'll add a comment about this field. It is only required when data structures are not changed but their range or meaning are: added flags, enum values.
Adding a field, changing a data type, adding a variant type will automatically break protocol compatibility.

xytovl:
Have you checked if [/interaction_profiles/microsoft/xbox_controller](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#_microsoft_xbox_controller_profile) is implemented on Quest? This would be more portable, and we might be able to expose it through OpenXR as well

dniminenn:
> Have you checked if /interaction_profiles/microsoft/xbox_controller is implemented on Quest? This would be more portable, and we might be able to expose it through OpenXR as well

I have not. I don't have such a controller and the DS I was using registered as hid-playstation. Ultimately I wanted to forward the raw HID but Quest will not allow it. This is probably worth an exploration.

> but I'd like to discuss the proper user interface to configure features.

Of course. I just wanted to get shooting down bogeys ASAP 😂

In any case, I am happy to implement the proposed changes and roll them into the PR once we/the community have aligned on how to ship the feature.

Cheers,

JR

dniminenn:
Thanks for the review. Reworked along the lines you suggested:

- Capture now goes through the OpenXR `microsoft/xbox_controller` profile. It binds on Quest, so the `android_native_app_glue` pre-dispatch workaround is gone. Rumble uses OpenXR haptics now as well.
- The server side is a single `forward-input-devices` permission. The per-device choice (keyboard, mouse, gamepad) lives on the headset and can be set before connecting. The old `hid-forwarding` and `gamepad-forwarding` keys still enable it.
- Added the native OpenXR exposure you mentioned: a Monado `xbox_controller` device at `/user/gamepad`, enabled from the headset interaction profile. Off by default via `gamepad-openxr`.
- Dropped the `protocol_revision` bump.

Tested end to end with a DualSense and Project Wingman, with input confirmed on both the uinput device and `/user/gamepad`.

client/application.cpp
```diff
@@ -1770,13 +1806,104 @@ void application::push_scene(std::shared_ptr<scene> s)
 	instance().scene_stack.push_back(std::move(s));
 }
 
+namespace
+{
+// Read the OpenXR microsoft/xbox_controller gamepad. nullopt if none connected or not focused.
+std::optional<wivrn::from_headset::gamepad::state> read_gamepad()
```
xytovl: With the interaction profile method, you don't need a read_gamepad method or separate packets.
`scenes::stream::read_actions` can be used, you would have to allocate device ids for the gamepad and reuse the same mechanisms.

dashboard/settings.cpp
```diff
@@ -367,20 +367,27 @@ void Settings::set_application(const QString & value)
...
 }
 
-bool Settings::hidForwarding() const
+bool Settings::forwardInputDevices() const
 {
-	auto it = m_jsonSettings.find("hid-forwarding");
-	if (it != m_jsonSettings.end() and it->is_boolean())
-		return *it;
+	// Legacy keys also enable the unified permission.
+	for (const char * key: {"forward-input-devices", "hid-forwarding", "gamepad-forwarding"})
```
xytovl: If we go the OpenXR route both on client and server, this option should only control if we use uinput to expose the controller. The `hid-forwarding` name is still valid.

client/application.cpp
xytovl: The interaction profile changes are sent for left/right hand in the tracking thread, for consistency we should also send the gamepad interaction profile (so array of 3). The switch to "none" signals that the gamepad is disconnected and should be reflected on the server

dniminenn: Noted. Is this too wide a surface for a single PR, I can break it up in easier to review chunks. If not I will continue refining according to your review.

xytovl: It's fine as a single PR, the changes excluding configuration should be:

protocol change: add ids for gamepad inputs, add one entry in the interaction profile array for gamepad

client changes: add suggested bindings, mapping to ids, send in the inputs packet. Send the gamepad interaction profile when it changes

server changes: add the xdev for gamepad (OpenXR), for the uinput part I don't know if it's better as a separate object or an optional component in the xdev

dniminenn:
Reworked onto the existing input rails as suggested.

Protocol: gamepad device ids in the inputs packet, one extra entry in the interaction profile array for the gamepad. No separate gamepad packets.

Client: suggested bindings on the xbox_controller profile, mapped to the gamepad ids, sent in the inputs packet. The gamepad interaction profile is sent from the tracking thread alongside the hands, switching to `none` on disconnect.

Server: a gamepad xdev exposes `/user/gamepad`, connected from the headset interaction profile. The uinput mirror is a separate object, gated by `hid-forwarding`, which now only controls the uinput exposure. I went with a separate object but am happy to fold it into the xdev as an optional component if you prefer, easy to switch either way.

Dropped the `protocol_revision` bump.

The headset has per device toggles for keyboard, mouse and gamepad. The OpenXR gamepad needs no permission so it is always exposed. With `hid-forwarding` off the headset shows that forwarded keyboard and mouse will not reach the server.

Tested end to end with a DualSense and Project Wingman, input confirmed on both the uinput device and `/user/gamepad`.

client/wivrn_client.h
```diff
@@ -49,6 +50,9 @@ class wivrn_session
 	std::atomic<uint64_t> bytes_sent_ = 0;
 	std::atomic<uint64_t> bytes_received_ = 0;
 
+	// sends may come from the render thread and the input thread concurrently
+	std::mutex send_mutex;
```
xytovl: the sockets are thread safe, and we use atomics for the counters, no need to add synchronization

common/wivrn_packets.h
```diff
@@ -320,7 +342,7 @@ struct tracking
 	};
 
 	// /user/hand/left and /user/hand/right
-	std::array<interaction_profile, 2> interaction_profiles;
+	std::array<interaction_profile, 3> interaction_profiles; // left hand, right hand, gamepad
```
xytovl: The comment was right above the member, keep only one

dniminenn:
Fixed, and removed the mutex

server/driver/wivrn_session.cpp
```diff
@@ -669,6 +678,26 @@ void wivrn_session::operator()(from_headset::inputs && inputs)
...
 		right_controller.set_inputs(inputs, offset);
+
+	if (gamepad_device)
+	{
+		gamepad_device->set_inputs(inputs);
+		try
+		{
+			// Mirror to a uinput gamepad for non-OpenXR consumers, when permitted.
+			if (uinput_handler)
+				uinput_handler->handle_gamepad(gamepad_device->get_state());
```
xytovl: I think we should be able to use the `inputs` object directly here, without doing the translation through `get_state`. This would remove the `gamepad` struct in wivrn_packets and get_state

common/wivrn_packets.h
```diff
@@ -770,6 +825,14 @@ struct refresh_rate_change
 	float hz;
 };
 
+// Rumble forwarded back to the controller. Best-effort; magnitudes normalized.
+struct gamepad_rumble
```
xytovl: This should just be a haptics packet, with 4 new device ids.

I don't know if there's specs on low and high frequencies, in OpenXR you would use that instead of specifying the motor.

dniminenn: ~~Switching rumble to a haptics packet; evdev force feedback only gives the two rumble motors, so I'll add just the left and right haptic ids unless you'd rather reserve trigger ids too.~~

Went with all four haptic ids (left/right plus both triggers) to match the convention of defining ids for every output even when not all are driven. evdev force feedback only fills the two motors, so the trigger ids stay reserved.

On frequency: the two outputs map 1:1 to evdev's strong/weak magnitudes, so I kept one output per motor rather than encoding low/high via the frequency field, which a fixed gamepad motor can't act on anyway.

Also fixed the other review items and refactored to be more idiomatic after studying the rest of the driver: gamepad inputs feed the uinput mirror straight from the inputs packet (dropped `gamepad::state` and `get_state`), and the `device_id` mapping is now a single shared `map_gamepad` table consumed by both the OpenXR xdev and the uinput mirror, instead of a duplicated switch.

server/driver/wivrn_gamepad.cpp
```diff
@@ -0,0 +1,139 @@
...
+static_assert(gp_count == 18);
+
+wivrn_gamepad::wivrn_gamepad(wivrn::wivrn_session & cnx) :
+        xrt_device{
+                .name = XRT_DEVICE_XBOX_CONTROLLER,
+                .device_type = XRT_DEVICE_TYPE_GAMEPAD,
+                .str = "WiVRn Gamepad",
+                .serial = "WiVRn Gamepad",
+                .tracking_origin = &origin,
+                .update_inputs = method_pointer<&wivrn_gamepad::update_inputs>,
```
xytovl, line 43: You should also handle `set_output` for OpenXR haptics

client/scenes/lobby.h
```diff
@@ -63,6 +63,10 @@ class lobby : public scene_impl<lobby>
 	std::string server_name;
 	bool autoconnect_enabled = true;
 
+	// Last known uinput forwarding state of the connected server, kept after disconnection so the
+	// settings can warn that forwarded keyboard and mouse will not reach the server.
+	std::optional<bool> server_hid_forwarding;
```
xytovl, line 68: We could use the `gui_toast` object in scene.h to notify users when they connect

server/driver/wivrn_gamepad.cpp
dniminenn: Done both.

`set_output`: wired it on the gamepad xdev the same way as `wivrn_controller`. The four `XRT_OUTPUT_NAME_XBOX_HAPTIC_*` outputs map to the `GAMEPAD_HAPTIC_*` ids and send a `haptics` packet. The client already had those four ids registered, so OpenXR `xrApplyHapticFeedback` on `/user/gamepad` now reaches the headset.

Toast: `gui_toast` lives on the stream scene, not `scene.h`, so I put it in the `feature_control` handler that runs on connect. When the server has hid-forwarding disabled and the user has keyboard or mouse forwarding on, it raises a toast on connect, reusing the same string as the existing settings-page warning.

client/android/hid.cpp
```diff
@@ -1,6 +1,8 @@
 #include "hid.h"
 
 #include "scene.h"
+#include <algorithm>
+#include <cmath>
```
xytovl: No longer needed

client/android/hid.h
```diff
@@ -2,6 +2,7 @@
 
 #include "scene.h"
 #include <android_native_app_glue.h>
+#include <atomic>
```
xytovl: No longer needed

client/scenes/lobby_gui.cpp
```diff
@@ -784,6 +784,29 @@ void scenes::lobby::gui_settings()
...
+		if (ImGui::Checkbox(_S("Forward keyboard"), &config.forward_keyboard))
+			config.save();
+		imgui_ctx->vibrate_on_hover();
+		if (ImGui::Checkbox(_S("Forward mouse"), &config.forward_mouse))
+			config.save();
+		imgui_ctx->vibrate_on_hover();
+		if (ImGui::Checkbox(_S("Forward gamepad"), &config.forward_gamepad))
+			config.save();
+		imgui_ctx->vibrate_on_hover();
+		if (ImGui::IsItemHovered())
```
xytovl: It may be nice to split the HID related changes into a separate commit (same pull request), in case we missed a side effect in mouse/keyboard code.

client/scenes/stream.cpp
```diff
@@ -21,6 +21,7 @@
 #include "xr/face_tracker.h"
 #include "xr/fb_face_tracker2.h"
 #include "xr/space.h"
+#include <cstring>
```
xytovl: If that's for `size_t`, `<cstddef>` is a more appropriate header

server/driver/wivrn_uinput.cpp
```diff
@@ -111,6 +120,133 @@ void wivrn_uinput::handle_input(wivrn::from_headset::hid::input & e)
...
+		};
+	};
+
+	input_event ev;
+	while (::read(fd, &ev, sizeof(ev)) == (ssize_t)sizeof(ev))
+	{
+		if (ev.type == EV_UINPUT and ev.code == UI_FF_UPLOAD)
+		{
+			uinput_ff_upload upload{};
+			upload.request_id = ev.value;
```
xytovl: I prefer the pattern
```cpp
uinput_ff_upload upload{
  .request_id = ev.value,
};
```
Readers directly see that we set one value

server/driver/wivrn_uinput.cpp
```diff
@@ -111,6 +120,133 @@ void wivrn_uinput::handle_input(wivrn::from_headset::hid::input & e)
...
+		{
+			// ev.code is the effect id, ev.value != 0 starts it, 0 stops it.
+			if (ev.value == 0)
+			{
+				motors(0.f, 0.f, 0);
+			}
+			else if (auto it = ff_effects.find(ev.code); it != ff_effects.end() and it->second.type == FF_RUMBLE)
+			{
+				const auto & rumble = it->second.u.rumble;
+				motors(rumble.strong_magnitude / 65535.f, rumble.weak_magnitude / 65535.f, (uint16_t)it->second.replay.length);
```
xytovl, line 260: Do you have any source for mapping strong to left and weak to right?

uinput.h says
>  Some rumble pads have two motors of different weight. Strong_magnitude represents the magnitude of the vibration generated by the heavy one.
 
OpenXR does not specify what "haptic_left" and "haptic_right" mean, but it's safe to assume it's the physical location of the haptic motor. The `amplitude` field would be a better match.

dniminenn:
Split the work in two commits as suggested: gamepad over OpenXR, then the uinput mirror on top.

- removed the unused includes and the stray `<cstring>`
- uinput force-feedback requests use designated initializers
- rumble: mapping strong to the left motor and weak to the right followed SDL's convention (low-frequency = left motor), but that is an SDL abstraction, not the evdev contract, where strong/weak are motor weights. Switched to `max(strong, weak)` on both outputs.

xytovl:
I finally had time to do some quick testing

x and y buttons are inverted in Steam controller testing (when going through the uinput layer)
rumble didn't seem to work through Steam controller testing, I have not tested through OpenXR

the uinput controller gets created even when not selected on headset

The dashboard tooltip may be confusing, I would change it to
"Replicate mouse, keyboard and gamepad connected to the headset as virtual devices on PC.
Replicated devices will appear as if they were plugged to the PC, some keys may be reserved by the headset OS and not be available. Gamepad is also available without virtual devices for applications that access it through OpenXR."

On the headset, it may be better to have a "virtual devices" section and "mouse", "keyboard", "headset" as checkboxes (indented).
"virtual devices" tooltip to explain this is the uinput stuff, say that it needs to be enabled on server and is not available under flatpak.

dniminenn:
> x and y buttons are inverted in Steam controller testing (when going through the uinput layer) rumble didn't seem to work through Steam controller testing, I have not tested through OpenXR

Noted, I did not notice this on PW with DS controller, square cross circle and triangle mapped correctly. For the rumble I was not able to get it to fire either. I'm not sure if we should pull the feature, no-op rumble, or leave it wired in. Likewise for the OpenXR side, I have genuinely not tested.

> the uinput controller gets created even when not selected on headset

Noted, this should be an easy bugfix.

> The dashboard tooltip may be confusing, I would change it to "Replicate mouse, keyboard and gamepad connected to the headset as virtual devices on PC. Replicated devices will appear as if they were plugged to the PC, some keys may be reserved by the headset OS and not be available. Gamepad is also available without virtual devices for applications that access it through OpenXR."

Thank you for the revised wording.

> On the headset, it may be better to have a "virtual devices" section and "mouse", "keyboard", "headset" as checkboxes (indented). "virtual devices" tooltip to explain this is the uinput stuff, say that it needs to be enabled on server and is not available under flatpak.

Good feedback.

xytovl:
For rumble, we first need to check if it works on the headset directly, then if openxr haptic thought the server works and finally through uinput.

dniminenn:
Rebased on master. OpenXR commit untouched, all fixes in the uinput commit.

**X/Y**: my bug. xpad/SDL expect the kernel aliases (`BTN_X == BTN_NORTH`, `BTN_Y == BTN_WEST`), I used the position codes. Fixed, verified in Steam's tester.

**uinput when not selected**: fixed. The checkbox now gates only the mirror, OpenXR always gets the inputs. Sent as a `mirror_gamepad` bool in the inputs packet. Device created on first gamepad values, destroyed on toggle off or profile none.

**Tooltip**: took your wording. Headset UI is now a "Virtual devices" section with indented checkboxes.

**Rumble**: your first check passes, the DS5 rumbles fine when driven directly on the headset through the Android input vibrator API. I still need to test it through the stream. I aligned `read_rumble` to `send_stream` like `wivrn_controller`. Also noticed a pre-existing self-assignment in `stream::operator()(haptics &&)` (`it->second.amplitude = old;`) that makes stop requests always skip, not fixed in this PR.

common/wivrn_packets.h
```diff
@@ -435,6 +461,9 @@ struct inputs
 		XrTime last_change_time;
 	};
 	std::vector<input_value> values;
+	// Whether the server should mirror the gamepad to a virtual uinput device;
+	// gamepad inputs are always forwarded for the OpenXR path
+	bool mirror_gamepad = false;
```
xytovl: We may want to move that into the `settings_changed` structure (runtime configuable settings).

The code to handle settings update is not yet very nice, but I think it's the right place for this.

dniminenn: Done. Set at connect via the headset info settings, and on settings updates.

xytovl:
I have tested quickly and I think it's fine for merging.

Rumble still doesn't work from the test app but I don't know if it works at all through the headset.
One little thing that could be improved is to only have the uinput device visible if the interaction profile is enabled.

dniminenn:
uinput device is now gated on the gamepad interaction profile.

To check rumble on the headset without WiVRn: evdev from adb shell is SELinux-blocked, but the input vibrator API works through app_process. This buzzed my DS5 over BT:

```java
// Rumble.java
import android.os.Looper;
import android.os.VibrationEffect;
import android.os.Vibrator;
import android.view.InputDevice;

public class Rumble {
	public static void main(String[] args) throws Exception {
		Looper.prepareMainLooper();
		int target = args.length > 0 ? Integer.parseInt(args[0]) : -1;
		for (int i : InputDevice.getDeviceIds()) {
			InputDevice d = InputDevice.getDevice(i);
			if (d != null)
				System.out.println(i + ": " + d.getName() + " hasVibrator=" + d.getVibrator().hasVibrator());
		}
		if (target >= 0) {
			InputDevice d = InputDevice.getDevice(target);
			Vibrator v = d.getVibrator();
			v.vibrate(VibrationEffect.createOneShot(1500, 255));
			Thread.sleep(2000);
			v.cancel();
		}
	}
}
```

```sh
javac --release 11 -classpath $ANDROID_HOME/platforms/android-34/android.jar -d classes Rumble.java
$ANDROID_HOME/build-tools/34.0.0/d8 --output . classes/Rumble.class   # d8 needs java 11+
adb push classes.dex /data/local/tmp/rumble.dex
adb shell CLASSPATH=/data/local/tmp/rumble.dex app_process /system/bin Rumble        # lists input devices
adb shell CLASSPATH=/data/local/tmp/rumble.dex app_process /system/bin Rumble <id>   # 1.5s full rumble
```

PR #989, opened 2026-06-25
Add keywords to the desktop file and add 'Game' category

dawkagaming:
Hello,

I tried to improve the desktop file.


xytovl:
I don't think it should be in the Game category, per <https://specifications.freedesktop.org/menu/latest/category-registry.html> it's for games themselves.

For keywords, I would drop "SteamVR" and plain "Virtual", keywords should accept spaces, so "Virtual reality" should be fine.

dawkagaming:
You are right, thank you.

dawkagaming:
Thanks!

PR #993, opened 2026-06-27
Auto-detect audio backends instead of requiring PulseAudio

damaestro:
Probe libpipewire/libpulse without REQUIRED and default each WIVRN_USE_{PIPEWIRE,PULSEAUDIO} option to whether its library was found, mirroring the existing encoder auto-detection. An explicit -DWIVRN_USE_*=ON still force-requires the library , and a guard requires at least one backend. Drop the forced WIVRN_USE_PULSEAUDIO=ON from the base preset; CI still builds both since the dev libs are installed.


damaestro:
I got sick of having to pass `-DWIVRN_USE_PULSEAUDIO=OFF` after switching my dev workflow to not use envision. :)

xytovl:
I prefer not having automatic dependencies. I don't really use the presets, pulseaudio backend should be off, like the default in CMakeLists.txt.
I should actually drop pulseaudio entirely...

damaestro:
> I prefer not having automatic dependencies. I don't really use the presets, pulseaudio backend should be off, like the default in CMakeLists.txt. I should actually drop pulseaudio entirely...

Ah, then let's close this in favor of dropping pulse. I thought you were keeping it around for compat reasons.

PR #997, opened 2026-06-30
Remove PulseAudio audio backend

damaestro:
PipeWire is the modern, preferred backend. Remove the now-redundant PulseAudio backend entirely.


server/wivrn_ipc.cpp
```diff
@@ -78,7 +78,7 @@ void run_cleanup_functions()
 	if (std::ranges::all_of(*cleanup_functions, [](auto & i) { return i.callback == nullptr; }))
 		return;
 
-	// Fork because pulseaudio doesn't like being initialized in the parent and the child
+	// Fork because the audio backend doesn't like being initialized in the parent and the child
 	// of a fork, and ipc_server_main() is called in a child process
 	pid_t child = fork();
```
xytovl, line 84: The whole cleanup_functions system is now unused. The only purpose was to unload the pulseaudio module in case of unclean child exit.

We can do a second commit in the same pull request to remove cleanup functions entirely.

PR #1003, opened 2026-07-03
Headless per-encoder Profiling of the WiVRn Pipeline

damaestro:
`tools/perfetto/encoder_profile.py` runs one session per encoder (headless Monado as the client runtime + wivrn-server auto-launching hello_xr + the wivrn client), writes `out/<enc>.pftrace`, and prints a summary + baseline diff.

- Encoders default to what the GPUs support; failures are reported and skipped.
- Adds the `profiling` CMake preset and `docs/headless.md`.
- `pftrace_summary.py` gains `--exclude`; the harness drops the network slice.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -84,6 +84,15 @@
 			}
 		},
 		{
+			"name": "profiling",
+			"displayName": "Server (tracing, all encoders) + client, for encoder profiling",
+			"inherits": "server-tracing",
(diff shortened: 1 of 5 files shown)


xytovl:
I am not sure what the point is. Encoders other than vulkan video are more or less deprecated and I don't intend on doing much comparisons.

damaestro:
I'm going to redo this and make it for profiling specifically vulkan in a git bisect workflow.

PR #1033, opened 2026-07-19
server: add supersampling support to the compositor

caszuu:
Adds a multisampling step to the layer squasher compute shader. (partially based on Reyka's monado implementation)

It primarily helps with applications which output layers at higher resolutions than they are being sampled at (e.g. WayVR) which currently causes a lot of aliasing or "shimmering" on high detail portions of the layer.


caszuu:
I'm split if there should be a client option for controlling supersampling or if it's okay to leave it hard-coded as there's close to no benefit to disabling it. The performance impact is close to negligible and while in theory it blurs the sampled output, in reality details which would be blurred currently shimmer anyway.

patches/monado/0010-c-shaders-target-spirv-1.3-Vulkan-1.1.patch
xytovl, line 1: This is not needed, we don't use Monado code to build the shaders: https://github.com/WiVRn/WiVRn/blob/master/server/CMakeLists.txt#L278

caszuu: ah sorry i only included it because the build was failing. Should we disable Monado-side compilation in that case?

It seems the shader library (`comp_shaders`) is linked to the renderer library (`comp_render`) so that change would have to be patched in.

xytovl: I'll have to check the dependency chain if we can disable the parts we don't actually need

xytovl:
Is there an upstream merge request for this ? I don't like carrying Monado-only patches

caszuu:
No, not yet. I planned on upstreaming the patch later, but I can close this for now and try to submit it to monado first if that's cleaner.

xytovl:
I just ask for a plan for upstreaming, we've done it in the past to carry patches for some time before getting them into Monado.

Do you have something to change before removing the draft status? Reviewing patches is not very comfortable so I'll look at the actual diff onces it's ready for review.

patches/monado/0010-c-shaders-target-spirv-1.3-Vulkan-1.1.patch
xytovl: We use `comp_base.c`, ` comp_vulkan.c` from `comp_util` target, which links to `aux_shaders`.
I don't think it's useful to further split `comp_util`, the patch in the build is the lesser evil.

caszuu:
Sorry for the abrupt delay.

The only reason i put it up as a draft was the lack of user control over the feature. After some more testing i decided to just add it as an option to the client since i feel this might be a feature every user reacts to differently. (due to it being a tradeoff between shimmering and slight blur)

The only thing i'm worried about is having the term "Supersampling" in the GUI twice. I tried to differentiate the postprocessing version from the server-side one using tooltips, but if it's still too unclear it can be changed.

damaestro:
Huh.. those `ruff` errors must be new checks. I can push updates if you want or just run the tool in fix mode to make the needed updates.

xytovl:
> Huh.. those `ruff` errors must be new checks. I can push updates if you want or just run the tool in fix mode to make the needed updates.

Ruff now uses 0.16 instead of 0.15, which is why we get new errors.
If you can make a fix that would be nice.

xytovl:
> Sorry for the abrupt delay.
> 
> The only reason i put it up as a draft was the lack of user control over the feature. After some more testing i decided to just add it as an option to the client since i feel this might be a feature every user reacts to differently. (due to it being a tradeoff between shimmering and slight blur)
> 
> The only thing i'm worried about is having the term "Supersampling" in the GUI twice. I tried to differentiate the postprocessing version from the server-side one using tooltips, but if it's still too unclear it can be changed.

For GUI changes, you should wait some days, we're rebasing ahsanswr's branch as https://github.com/WiVRn/WiVRn/tree/ui-overhaul and that rewrites almost all of the client GUI

xytovl:
The GUI changes have been pushed to master, I believe it's easier to rewrite the GUI part than resolving merge/rebase conflicts.

caszuu:
ah newer monado, will fix shortly

ImSapphire (approved):

ImSapphire:
Thank you for working on this! Works very well for me

client/scenes/gui_settings.cpp
```diff
@@ -281,6 +281,24 @@ void settings_video(const settings_context & ctx)
 		});
 	}
 
+	{
+		list.push_back({
+		        .id = "##supersampling",
+		        .label = _("Supersampling"),
```
xytovl: We should add context for new localisable elements with `_C(context, text)` macro, for full sentences this is less important as the meaning will be clear for translators.

client/scenes/gui_settings.cpp
```diff
@@ -281,6 +281,24 @@ void settings_video(const settings_context & ctx)
...
+		        .ui = ui_kind::segmented,
+		        .get_int = [&config] { return std::log2(config.supersampling_mode); },
+		        .set_int = [&ctx, &config](int v) { config.supersampling_mode = 1u << v; config.save(); if (ctx.on_streaming_changed) ctx.on_streaming_changed(); },
+		        .options = [] {
+					std::vector<std::string> opts;
+					opts.push_back(_("None"));
+					for (uint32_t i = 2; i <= 16; i <<= 1)
+						opts.push_back(fmt::format("{}x", i));
+					return opts; },
+		        .default_int = 2, // 4x
```
xytovl: This would be better if it's the same source as the default value in the configuration object.

client/scenes/gui_settings.cpp
```diff
@@ -487,7 +505,7 @@ void settings_post_processing(const settings_context & ctx)
 			});
 		};
 
-		flag_combo("##supersampling", _("Supersampling"), _("Reduce flicker for high contrast edges. Useful when the input resolution is high compared to the headset display."), {0, XR_COMPOSITION_LAYER_SETTINGS_NORMAL_SUPER_SAMPLING_BIT_FB, XR_COMPOSITION_LAYER_SETTINGS_QUALITY_SUPER_SAMPLING_BIT_FB}, &configuration::openxr_post_processing_settings::super_sampling);
+		flag_combo("##supersampling", _("Supersampling"), _("Reduce flicker for high contrast edges during decoding. Useful when the input resolution is high compared to the headset display."), {0, XR_COMPOSITION_LAYER_SETTINGS_NORMAL_SUPER_SAMPLING_BIT_FB, XR_COMPOSITION_LAYER_SETTINGS_QUALITY_SUPER_SAMPLING_BIT_FB}, &configuration::openxr_post_processing_settings::super_sampling);
```
xytovl: maybe "after decoding", but I agree it's not easy to explain the difference for composition supersampling and reprojection supersampling

common/wivrn_packets.h
```diff
@@ -291,6 +291,7 @@ struct settings_changed
 
 	uint32_t fps_divider = 1;
 	uint32_t bitrate_bps;
+	uint32_t supersampling_mode;
```
xytovl: maybe "composition_sample_count"

xytovl:
This is going to be very hard to describe to users, compositor supersampling is only applied when multiple OpenXR layers are used (or we'll go through fast path). In real life, this more or less means that it will only affect WayVR.

The similar post-processing option will confuse users too as it does the same but on headset side, meaning that it will affect the fast path too.

ImSapphire:
I'm honestly not sure keeping the OpenXR supersampling option really makes sense? I never noticed any visual improvement from using it, compositor supersampling looks way better.

xytovl:
OpenXR supersampling is supposed to be useful when the rendered resolution is higher than the display.

I think I already discussed that in Monado's discord, compositor supersampling should be controlled through that extension, so WayVR could request higher quality

patches/monado/0011-a-shaders-add-per-layer-supersampling.patch
xytovl, line 1: It could make sense to add support for `XR_FB_composition_layer_settings` on Monado side, drop all the configuration here and let WayVR provide the flag.

client/scenes/gui_settings.cpp
caszuu: fixed

client/scenes/gui_settings.cpp
caszuu: fixed

common/wivrn_packets.h
caszuu: fixed

client/scenes/gui_settings.cpp
caszuu: counterintuitively, the default option matches the configuration here because the ui element operates with log2 of the actual sample count. (so default_int = 2 implies 4 samples as the default)

If there is a cleaner way to do this, i'll fix it right away.

caszuu:
> This is going to be very hard to describe to users, compositor supersampling is only applied when multiple OpenXR layers are used (or we'll go through fast path). In real life, this more or less means that it will only affect WayVR.

It is always only applied to non-projection layers. I excluded projection layers from performing supersampling since a vast majority of applications render at `recommendedImageRectWidth/Height` (and any extra samples would not affect the image in any way), so fast path should look identical.

> The similar post-processing option will confuse users too as it does the same but on headset side, meaning that it will affect the fast path too.

 I agree with the confusion part though, tooltips can get us only so far. I also have the same experience as sapphire where the client-side supersampling seems to do close to nothing (and i have a hunch the quest compositor does not even implement the supersampling setting bits), but i'm not sure if it should be removed.

> It could make sense to add support for `XR_FB_composition_layer_settings` on Monado side, drop all the configuration here and let WayVR provide the flag.

That is the plan for the Monado implementation and probably the correct thing to do. The reason i added manual control for this PR was that no applications (maybe with the exception of WayVR later on) would implement the extension and lose out on the quality gain.

Moreover, i added the GUI setting itself because some users might still prefer having all applications rendered with minimal blur. (or on the other hand might want to select the `16x` option if someone is trying to display a 8K desktop view on a poor quest 2)

Reading the specification it seems that the runtime is free to interpret to the setting bits (or lack thereof) in any way it wants. It might be possible to have a default supersampling setting when no layer settings are provided and allow the application to override it when applicable.

client/scenes/gui_settings.cpp
xytovl: You can use `std::countr_zero(configuration{}.compositor_sample_count)`

If it's easier, we can also store in config log2_compositor_sample_count

xytovl:
I think the good solution would be to use the extension to give applications control, have sane defaults (the debug gui shows compositor timings, perfetto traces could be added easily).

If we can avoid user-visible settings, that will be better, I don't think multisampling here is a performance/quality tradeoff that is worth exposing.

xytovl:
I have quickly tested the change, the performance hit is quite low and the improvement in WayVR is visible (2x is sufficient, additional levels change close to nothing).

However it would be better to have a Monado merge request, and there may be relevant discussions on configuration.

Beyley:
> That is the plan for the Monado implementation and probably the correct thing to do. The reason i added manual control for this PR was that no applications (maybe with the exception of WayVR later on) would implement the extension and lose out on the quality gain.

This is the way it should work by default, yeah. We probably shouldn't be supersampling things unconditionally, since it _is_ a minor perf hit. Adding this extension to WayVR would be very easy, likely only a couple lines of change.

> Moreover, i added the GUI setting itself because some users might still prefer having all applications rendered with minimal blur. (or on the other hand might want to select the `16x` option if someone is trying to display a 8K desktop view on a poor quest 2)

This should just be an environment variable (`OXR_FORCE_SUPERSAMPLING=?` for settings it) to force it for a particular app, rather than a global option/environment variable. Putting it into libmonado is also a good idea, to force particular supersampling settings of a particular app at runtime.

caszuu:
Okay, I implemented the changes as requested. The configuration is now dropped in favor of listening to `XR_FB_composition_layer_settings` and `OXR_FORCE_SUPER_SAMPLING=<sample count>`.

I'm only pushing this here for the sake of completeness, but the PR can stay open or be closed until I get to upstreaming the patches if that's preferable.

caszuu:
Also a quick dirty patch for `xrgears` for testing layer settings. 

[0001-Request-supersampling-for-quad-layers-when-available.patch](https://github.com/user-attachments/files/30947616/0001-Request-supersampling-for-quad-layers-when-available.patch)

xytovl:
Thanks again for the changes, I think upstreaming will be quick, if there's issues we could merge as a patch, the interface change is simple so it would not be a big issue to update to whatever ends up in Monado.

PR #1037, opened 2026-07-22
Add client-side chroma key passthrough

anosatsuk124:
Implements client-side chroma keying for the entire decoded video stream, similar to the workflow supported by Virtual Desktop and ALVR.

This allows the headset’s passthrough environment—FB, HTC, or another alpha-blend environment—to be used with OpenXR applications that do not output alpha themselves.

Because the keying is applied to the final decoded stream on the WiVRn client, it can also be used with compositor overlays included in that stream. For example, an overlay application can display a chroma-key-coloured region to create an arbitrarily positioned passthrough window, without requiring support from the underlying application, VRChat avatar, or world.

This differs from WayVR’s existing chroma key feature, which applies to content presented through WayVR and does not provide the same whole-stream workflow.

## Motivation

Some applications do not provide alpha output, and modifying the application itself is not always practical or possible.

Applying chroma keying at the WiVRn client level makes the feature application-independent and enables workflows such as:

* Creating passthrough regions from an overlay application
* Using passthrough in applications that do not support alpha composition
* Avoiding application-, avatar-, or world-specific chroma key implementations
* Reproducing Virtual Desktop-style passthrough workflows on Linux

## Implementation

* The decoded video is sampled in the reprojection fragment shader.
* Pixels inside a configurable HSV range are made transparent.
* A configurable soft falloff, controlled by `curve`, is applied around the selected range.
* A despill pass attenuates residual key colour on pixels that remain visible.
* When chroma keying is enabled, stream composition is routed through the existing alpha-blend path, enabling passthrough automatically.
* Configuration is stored in `client/configuration.{h,cpp}` and persisted through JSON.
* The following settings are exposed in the in-headset stream settings GUI:

  * `enabled`
  * `hsv_min`
  * `hsv_max`
  * `curve`
  * `despill`
* The GUI includes:

  * Sliders for all configurable parameters
  * A hue-bar preview showing the selected key range
  * Support for hue-range wrap-around
  * A colour swatch representing the centre of the selected range
* Parameters are passed to the shader through a packed push-constant block in the defoveator.
* The existing “stream provides its own alpha” path is preserved.
* Native stream alpha and chroma keying are mutually exclusive to prevent double masking.

## Compatibility

* Client-side only
* No protocol changes
* No server-side support required
* Applications do not need to output alpha or implement chroma keying themselves

## Testing

Verified on-device with a headset supporting an alpha-blend passthrough environment.

Confirmed that:

* Keyed regions reveal the real-world passthrough view
* Chroma keying works with content included in the decoded stream
* Despill reduces green fringing around surviving pixels
* Hue-range wrap-around behaves correctly
* Settings persist across client restarts
* The existing native-alpha stream path continues to work when chroma keying is disabled

## Additional build fix

Also includes a small unrelated build fix:

* Adds the missing `<unistd.h>` include to `server/encoder/ffmpeg/video_encoder_va.cpp`
* Fixes `dup` and `close` being undeclared with recent toolchains

https://github.com/user-attachments/assets/a8dd02ea-95fd-4367-9dca-135fd1a89757

### DIFF
--- a/client/configuration.cpp
+++ b/client/configuration.cpp
@@ -177,6 +177,33 @@ configuration::configuration(xr::system & system, xr::session & session)
 		if (auto val = root["passthrough_enabled"]; val.is_bool())
 			passthrough_enabled = val.get_bool();
 
+		if (auto ck = root["chroma_key"]; ck.is_object())
+		{
+			if (auto val = ck["enabled"]; val.is_bool())
+				chroma_key.enabled = val.get_bool();
+			auto read_triplet = [](simdjson::simdjson_result<simdjson::dom::element> el, std::array<float, 3> & out) {
+				if (!el.is_array())
+					return;
+				size_t i = 0;
+				for (auto v: simdjson::dom::array(el))
+				{
+					if (i >= out.size())
+						break;
+					if (v.is_double())
+						out[i] = v.get_double();
+					else if (v.is_int64())
+						out[i] = v.get_int64();
+					++i;
+				}
+			};
+			read_triplet(ck["hsv_min"], chroma_key.hsv_min);
+			read_triplet(ck["hsv_max"], chroma_key.hsv_max);
+			if (auto val = ck["curve"]; val.is_double())
+				chroma_key.curve = val.get_double();
+			if (auto val = ck["despill"]; val.is_double())
+				chroma_key.despill = val.get_double();
+		}
+
 		if (auto val = root["mic_unprocessed_audio"]; val.is_bool())
 			mic_unprocessed_audio = val.get_bool();
 
@@ -318,6 +345,13 @@ void configuration::save()
 	json << ",\"openxr_post_processing\":";
 	write_openxr_post_processing(json, openxr_post_processing);
 	json << ",\"passthrough_enabled\":" << std::boolalpha << passthrough_enabled;
+	json << ",\"chroma_key\":{"
+	     << "\"enabled\":" << std::boolalpha << chroma_key.enabled
+	     << ",\"hsv_min\":[" << chroma_key.hsv_min[0] << "," << chroma_key.hsv_min[1] << "," << chroma_key.hsv_min[2] << "]"
+	     << ",\"hsv_max\":[" << chroma_key.hsv_max[0] << "," << chroma_key.hsv_max[1] << "," << chroma_key.hsv_max[2] << "]"
+	     << ",\"curve\":" << chroma_key.curve
+	     << ",\"despill\":" << chroma_key.despill
+	     << "}";
 	json << ",\"mic_unprocessed_audio\":" << std::boolalpha << mic_unprocessed_audio;
 	json << ",\"forward_keyboard\":" << std::boolalpha << forward_keyboard;
 	json << ",\"forward_mouse\":" << std::boolalpha << forward_mouse;
--- a/client/configuration.h
+++ b/client/configuration.h
@@ -22,6 +22,7 @@
 #include "wivrn_discover.h"
 #include "wivrn_packets.h"
 
+#include <array>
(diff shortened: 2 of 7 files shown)


galister:
This sounds like an XY problem to me. You're trying to implement rectangles that act as pass-through windows. There's no need to involve chroma keying here at all.

There are extensions like `XR_FB_passthrough` or `XR_ANDROID_composition_layer_passthrough_mesh` which already address the same need and can be implemented instead of "asking the user to paint their screens purple and then use chroma key".

anosatsuk124:
@galister

I see your point, but my use case is not limited to implementing rectangular passthrough windows.

The goal is to apply the mask to the final decoded stream so that existing overlay applications can define arbitrary and dynamic passthrough regions without being modified or explicitly submitting passthrough layers or geometry.

My understanding is that implementing "XR_FB_passthrough" or "XR_ANDROID_composition_layer_passthrough_mesh" in the WiVRn compositor would allow applications that explicitly use those extensions to submit passthrough regions, but would not by itself make existing overlay applications work as passthrough masks.

Am I misunderstanding how you expect those extensions to support this workflow?

xytovl:
I would like to avoid client-side features like this.
Extending the server side passthrough should and a headset side GUI to control it should be sufficient.

There may be solutions using "XR_FB_passthrough" or "XR_ANDROID_composition_layer_passthrough_mesh" on the server side to cover the use case through a different workflow.

We should not have features that overlap that much with Monado's chroma keying.

XenoPL:
@xytovl  Would server-side passtrough support custom masks, ie areas painted with some distinctive color? It's quite common way for MR experience in flight/racing sims, where people want to see their physical instruments/cockpits over rendered game world.
Also would it be possible to implement hands passtrough? A nifty VD feature, where with hand tracking enabled, hand models are replaced with orb shaped passtrough mask. Quite handy to be able to type, grab your mouse, game controller or your drink while being in VR.

xytovl:
server-side passthrough replaces specific colours with transparency, with the same tuning as this pull request. It only operates on projection layers by design (quads/cylinders will be unaffected).

We don't have hands passthrough, and in this case it would make sense to implement on headset side in order to use more up to date hand position. This would be a separate feature from chroma keying.

schenlap:
Is there currently a way to use chroma key passthrough? I need it for X-Plane on linux, but I haven't found any information about it apart from this PR. Perhaps there's a workaround I haven't discovered yet.

galister:
@schenlap if you use this branch of wivrn + latest wayvr, you'll get an option on the keyboard hamburger menu to a "add a passthrough cutout"

https://github.com/WiVRn/WiVRn/tree/poc/layer-alpha-blend
Downloads on bottom of page here:
https://github.com/WiVRn/WiVRn/actions/runs/31069781294

schenlap:
Thank you. In sideloaded the apk and installed flatpak version from https://github.com/WiVRn/WiVRn/actions/runs/31069781294.

It run's but I don't see any setting for chroma key in wivrn client app  (with or without X-Plane running in VR mode).

isn't this a setting in WiVRn? I am not using wayvr (until now).

galister:
if you want actual chroma key:
wayvr → dashboard → monado → chroma key

if you want rectangular cutouts:
wayvr → keyboard → hamburger menu

none of this is in wivrn settings, you'll need wayvr for this.

schenlap:
Ok, I did not know that wayvr is needed. I tested it with wayvr and it worked. Very nice :-)


(ok the a320 instruments don't fit in the B58.

ImSapphire:
I'm closing this since we already have chroma key on the server side, plus galister's alpha blend layer work should soon be available (which fulfils the use-case of having a passthrough cutout quad).

PR #1038, opened 2026-07-22
Add sunglasses mode: dim the stream with a configurable colour tint

anosatsuk124:
Adds a "sunglasses" mode to the client: a solid, HSV-selected colour is blended over the decoded stream with a configurable opacity, dimming the whole image (default colour is black, i.e. pure dimming). Useful to reduce eye strain in bright VR content.

Stacked on #1037 (chroma key passthrough) — the implementation mirrors its data path exactly and the code depends on it. Until #1037 merges, this PR's diff also shows its commits; only the last two commits are new here.

## Implementation

- `configuration.{h,cpp}`: new `sunglasses_settings` (enabled / HSV / alpha), persisted in `client.json` under `"sunglasses"`.
- `scenes/stream.cpp`: converts HSV to RGB on the CPU each frame and passes `sunglasses_params` to the defoveator.
- `scenes/stream_defoveator.{h,cpp}`: one extra `vec4 sunglasses_tint` push constant (xyz = RGB, w = opacity, 0 when disabled). The push constant block is now exactly 128 bytes — at the Vulkan guaranteed minimum.
- `shaders/reprojection.glsl`: applied at the end of `main()` as `mix(outColor.rgb, tint * outColor.a, opacity)`. Multiplying by `outColor.a` keeps premultiplied alpha intact, so chroma-keyed passthrough regions keep their real brightness. On sRGB swapchains the tint is converted to linear first.
- `scenes/stream_gui.cpp`: new "Sunglasses" section under the chroma key settings — enable checkbox, Hue/Saturation/Value/Opacity sliders, colour preview swatch, reset button; changes save immediately.

Also includes a standalone build fix for GCC 16 / recent OpenXR headers (missing `<mutex>` includes; `XR_META_body_tracking_fidelity` redefinition guard) needed to build this branch on an up-to-date toolchain.

## Design notes

- Client-side only; no protocol change (`wivrn_packets.h` untouched), works with any server.
- Passthrough regions are intentionally not tinted — only virtual content is dimmed.

## Testing

- Builds clean on Linux (native client + server, GCC 16.1.1) and Android (`assembleRelease`).
- Verified on device (Quest 3, server and client built from this tree): full-screen dimming with the opacity slider, HSV colour response matching the preview swatch, settings persisted across restarts, no regression with chroma key on or off (passthrough regions keep their real brightness).

https://github.com/user-attachments/assets/a4ac52f0-540c-4f32-84a9-2b22c10fd42e

### DIFF
--- a/client/application.cpp
+++ b/client/application.cpp
@@ -43,6 +43,7 @@
 #include <ctype.h>
 #include <exception>
 #include <magic_enum.hpp>
+#include <mutex>
 #include <string>
 #include <thread>
 #include <vector>
--- a/client/configuration.cpp
+++ b/client/configuration.cpp
@@ -25,6 +25,7 @@
 
 #include <fstream>
 #include <magic_enum.hpp>
+#include <mutex>
 
 #ifdef __ANDROID__
 #include "android/permissions.h"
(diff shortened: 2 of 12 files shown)


galister:
Does the brightness slider of the headset not cover this use case?

Wouldn't it be disturbing to the user if the HMD's system popups etc are still at full brightness, while they're at 10% brightness in stream?

ImSapphire:
> Does the brightness slider of the headset not cover this use case?

For a lot of people the lowest brightness on e.g. Quest 3 is still pretty bright. I use a screen dimmer on the PC side at night. It's also very useful for sleeping.

> Wouldn't it be disturbing to the user if the HMD's system popups etc are still at full brightness, while they're at 10% brightness in stream?

Not really, at least in my case I just enable do-not-disturb so I don't get notifications from the HMD.

I don't think we really need a sunglasses mode though, there's already screen dimmers that exist which run on the PC side (e.g. eepyxr, OyasumiVR fork).

ImSapphire:
I'm closing this since screen dimmer applications already exist, we don't need extra code inside WiVRn to fulfil this purpose.

PR #1043, opened 2026-07-25
fix(ci): bump pinned Vulkan SDK to 1.4.328.1

damaestro:
The Linux matrix moved to ubuntu-26.04, which has GCC >= 14 and CMake >= 3.28. Vulkan-Headers at vulkan-sdk-1.3.296.0 defaults VULKAN_HEADERS_ENABLE_MODULE to ${COMPILER_SUPPORTS_CXX_MODULES}, so that combination creates a CXX_MODULES target and the from-source SDK build fails at generate time:

> The target named "Vulkan-Module" has C++ sources that may use modules, but modules are not supported by this generator: Unix Makefiles

From vulkan-sdk-1.4.304.1 on, the option is opt-in (defaults to OFF), so the module target is never created.

Pin 1.4.328.1, the tag already used for SPIRV-Reflect in CMakeLists.txt.

### DIFF
--- a/.github/workflows/APK-Release.yml
+++ b/.github/workflows/APK-Release.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Prepare Vulkan SDK
       uses: humbletim/setup-vulkan-sdk@v1.2.1
       with:
-        vulkan-query-version: 1.3.296.0
+        vulkan-query-version: 1.4.328.1
         vulkan-components: Glslang, SPIRV-Tools
         vulkan-use-cache: true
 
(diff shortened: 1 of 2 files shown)


damaestro:
I noticed this build failure when I updated my fork to catch up to `master`.

xytovl:
We may have to increase the `Vulkan_VERSION` requirement in `CMakeLists.txt` if we test with much more recent ones.

We could upgrade to 1.4.341 from Ubuntu 26.04 and skip the setup-vulkan-sdk action

damaestro:
> We could upgrade to 1.4.341 from Ubuntu 26.04 and skip the setup-vulkan-sdk action

If you'd prefer to use the ubuntu headers, I'll do that update instead. It should speed up builds and if we don't need to target very specific versions, this seems like a net good thing?

xytovl:
If we can get rid of the action, that's better.

damaestro:
On it.

PR #1045, opened 2026-07-25
feat(vulkan): use distro Vulkan packages in CI instead of setup-vulkan-sdk 

damaestro:
Fixes CI on ubuntu-26.04 and stops three jobs refetching their dependencies every run.

 - Vulkan: use Ubuntu's libvulkan-dev / glslang-tools / spirv-tools instead of building the SDK from source with setup-vulkan-sdk, which fails on ubuntu-26.04 (Vulkan-Headers 1.3.296 defines a C++20 module target, unsupported by the Unix Makefiles generator). Server now requires Vulkan 1.4.304 — what every CI platform ships, Debian 13 / Ubuntu 25.04+ — which subsumes the 1.3.283 video-encode check and the server-side VK_HEADER_VERSION >= 304 branch.

 - flatpak: cache .flatpak-builder with restore-keys and save it on failure, drop CMakeLists.txt from the key, pre-fetch sources with retries, add mirror-urls to the toolchain tarballs, and use the container image matching the manifest's runtime version. A dead ftpmirror.gnu.org mirror was failing the job outright; build step 3008s → 595s on a cache hit.
 
- Android: cache ~/.gradle, so a Maven Central 403 no longer fails the build before it configures.

### DIFF
--- a/.github/workflows/APK-Release.yml
+++ b/.github/workflows/APK-Release.yml
@@ -25,6 +25,15 @@ jobs:
         ref: ${{ github.sha }}
         fetch-depth: 0
 
+    # Gradle distribution and the Android plugin's dependencies.
+    - uses: actions/cache@v5
+      with:
+        key: gradle-android-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties', 'build.gradle') }}
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        restore-keys: gradle-android-
+
     - name: Install dependencies
       run: |
         sudo apt-get update && sudo apt-get install glslang-tools spirv-tools librsvg2-bin gettext npm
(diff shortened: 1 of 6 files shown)


xytovl:
Do we really need complex stuff for flatpak cache? We have cache hits with the current system, what is it that needs fixing?

damaestro:
> Do we really need complex stuff for flatpak cache? We have cache hits with the current system, what is it that needs fixing?

I've been regularly running into flatpak failing because the `ftpmirror.gnu.org` times out, likely due to some rate limiting. This more aggressively caches so we call them a lot less, adds more mirrors, and updates the base container image so we are not having to do an upgrade immediately.

Leaving the caching as-is with only the mirror list fallback would work too. I figured we could just stop calling them under normal circumstances and save that bandwidth for someone else.

xytovl:
In past CI runs I can see lines like "Cache hit for: flatpak-builder-eb0da0fbbe868df899fc62d500fbccde479dc7f7b3d4b817bb5ebcf284141e21-x86_64"

I don't see why separating the download from the build would improve things.

PR #1050, opened 2026-07-29
Update Simplified Chinese localization

ZokuTe:


xytovl:
it looks like there's still 4 fields missing (or marked as fuzzy):
'Change'
'Enter PIN'
'Download community-made spaces.'
'STREAM'

PR #1054, opened 2026-07-29
Translated lots of new strings to pt_BR

Patola:
Fixes #446

### DIFF
--- a/locale/pt_BR/wivrn.po
+++ b/locale/pt_BR/wivrn.po
@@ -11,8 +11,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: WiVRn\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-07-25 18:50+0200\n"
-"PO-Revision-Date: 2026-07-11 13:28+0200\n"
+"POT-Creation-Date: 2026-07-28 00:23+0200\n"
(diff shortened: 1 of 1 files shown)


xytovl:
Make sure all the "fuzzy" annotations are gone, they are added when the text is unchanged but the context did

Patola:
> Make sure all the "fuzzy" annotations are gone, they are added when the text is unchanged but the context did

Oh. I didn't realize that. I used lokalize to translate the strings, do you know any easy way to spot that? Are you talking about the `msgctxt` fields?

xytovl:
Graphical tools will tell you translations are fuzzy, in the file that means there is a "#, fuzzy" annotation.
If the translation is correct, just remove the annotation.

Those were generated by translations where the text did not change but the context did, we tried to add context to most labels, especially when they're single word.

Patola:
Do I still need to change something for it to be accepted? Looking at the diff, the only #fuzzy tags came from what was already there, my commit didn't change them.

xytovl:
They were marked as fuzzy when we added the context strings, as we had no means to verify if they were accurate, they're fuzzy.

If they're fine, remove the annotation and we can merge.

Patola:
Done. There should be no more fuzzy strings, I reviewed them all (and grepped the file before committing).

PR #1060, opened 2026-08-07
assign xrt_device::id's

galister:
Necessary for Monado changes:

https://gitlab.freedesktop.org/monado/monado/-/commit/468a06d73ed9e00a7195d99c32424522e96dd5a7

https://gitlab.freedesktop.org/monado/monado/-/commit/4362f92ef6e831088b306fe8ef4c5775b358ab1d

> A limited unique id, it is only unique for the process it is in, so must not be
> used or synchronized across process boundaries. A value of zero is invalid
> and means it has not be properly initialised.

### DIFF
--- a/server/driver/wivrn_session.cpp
+++ b/server/driver/wivrn_session.cpp
@@ -47,6 +47,7 @@
 #include "b_hand_tracker.h"
 #include "b_system.h"
 #include "target_builder_helpers.h"
+#include "util/u_device_id.h"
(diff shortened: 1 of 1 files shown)


xytovl:
I was fixing it through https://github.com/WiVRn/WiVRn/tree/fix/device-id, I think it's safer in the constructor

galister:
absolutely! let me test it real quick.

PR #1061, opened 2026-08-07
add slider for LH_STICK_DEADZONE

galister:
With systemd becoming the norm, it's getting awkward to set arbitrary environment variables.

This adds a slider to the UI and setenv's the deadzone before steamvr_lh is initialized.

image

Tested and works!

### DIFF
--- a/dashboard/qml/SettingsPage.qml
+++ b/dashboard/qml/SettingsPage.qml
@@ -132,6 +132,26 @@ Kirigami.ScrollablePage {
                     toolTipText: i18n("Allows the use of lighthouse-based controllers and trackers.\nRequires SteamVR to be installed.\nDevices must be be powered on before connecting to WiVRn.\nAn external tool such as motoc is needed for calibration.")
                 }
             }
+            RowLayout {
+                visible: Settings.steamvr_lh_supported && steamvr_lh.checked
+                Kirigami.FormData.label: i18n("SteamVR joystick deadzone")
+                Controls.Slider {
+                    id: lh_stick_deadzone
+                    Layout.fillWidth: true
+                    from: 0.0
+                    to: 0.9
+                    stepSize: 0.05
+                    value: Settings.lhStickDeadzone
+                }
+                Controls.Label {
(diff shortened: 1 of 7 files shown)


server/driver/wivrn_session.cpp
```diff
@@ -141,6 +141,12 @@ wivrn::wivrn_session::wivrn_session(std::unique_ptr<wivrn_connection> connection
 
 	if (use_steamvr_lh)
 	{
+		if (configuration().lh_stick_deadzone.has_value() && configuration().lh_stick_deadzone.value() > 0.01f)
```
xytovl: It would be better to parse configuration only once, before the `use_steamvr_lh`.

`std::optional<float>` can be directly compared to a value, it's true if it has a value and compares as requested.

```cpp
configuration conf;
if (conf.lh_stick_deadzone > 0.01f)
  setenv("LH_STICK_DEADZONE", std::format("{:.2}", *conf.lh_stick_deadzone).c_str(), true);
```

server/driver/configuration.h
```diff
@@ -57,6 +57,7 @@ struct configuration
 	std::vector<std::string> application;
 	bool debug_gui = false;
 	bool use_steamvr_lh = false;
+	std::optional<float> lh_stick_deadzone;
```
xytovl, line 60: We would need a corresponding entry in the configuration.md doc

dashboard/qml/SettingsPage.qml
```diff
@@ -132,6 +132,26 @@ Kirigami.ScrollablePage {
                     toolTipText: i18n("Allows the use of lighthouse-based controllers and trackers.\nRequires SteamVR to be installed.\nDevices must be be powered on before connecting to WiVRn.\nAn external tool such as motoc is needed for calibration.")
                 }
             }
+            RowLayout {
+                visible: Settings.steamvr_lh_supported
```
xytovl: should be disabled or invisible when steamvr_lh is disabled

PR #1066, opened 2026-08-10
update monado - no longer need to set LH_DISCOVER_WAIT_MS

galister:
This is to pull in https://gitlab.freedesktop.org/monado/monado/-/commit/f037264d23e2472a444a157370647fcd601ed81b

which eliminates the need for users to set `LH_DISCOVER_WAIT_MS`

edit: probably squash-merge this so the PR message is visible and it'll hopefully make it into the changelog


xytovl:
Can you amend the commit with the message you want?

galister:
done

PR #1070, opened 2026-08-16
update pt-BR locale

NotBlankyu:
Fix typos, consistency and flow of the translation


xytovl:
@Patola do you agree with the changes?

Patola:
I agree. There's some tiny controversy on translating "supersampling" to "superamostragem" which would be the correct one in Portuguese but I never saw someone actually use "superamostragem", so the original "supersampling" word might be more appropriate indeed. I'll adopt this standard from now on too. Thanks for the corrections, @NotBlankyu.

NotBlankyu:
Ty i mostly changed to that one because it was the one used already in other strings.

PR #1071, opened 2026-08-20
Gracefully stop applications on server quit

ImSapphire:


xytovl:
It should work with Ctrl+C too, applications are spawned through systemd user service.

server/driver/wivrn_session.cpp
```diff
@@ -373,13 +373,27 @@ void wivrn_session::stop()
...
 }
 
-bool wivrn_session::request_stop()
+void wivrn_session::request_stop()
 {
-	assert(mnd_ipc_server);
-	bool b = net_thread.request_stop();
+	stop_application(std::nullopt, 2l * U_TIME_1S_IN_NS);
+
+	int64_t end = os_monotonic_get_ns() + 3l * U_TIME_1S_IN_NS;
```
xytovl, line 380: 3s looks like a long time, do we have some long to inform the user that the request has been received?

galister:
for signal handling, I would recommend something like this:
- add a signal handler for SIGINT & SIGTERM to the main process. having the same logic for both is fine IMO. just having a handler registered will stop the default abrupt termination behavior.
- upon receiving the signal, print something to stderr in order to let the user know that we are working on closing everything and it might take a while
- the main process shall gracefully shut down the forks and then shut itself down gracefully as well
- normally extra threads would be cleaned up by the OS in a non-graceful way, if there's graceful exit needed for a specific thread then the main thread needs to wait for that thread to finish before the main thread itself exits

xytovl:
We already handle signals through https://github.com/WiVRn/WiVRn/blob/master/server/main.cpp#L419-L428

Forks started through https://github.com/WiVRn/WiVRn/blob/master/server/start_application.h will receive signals too, and may not handle thel gracefully.

ImSapphire:
> It should work with Ctrl+C too, applications are spawned through systemd user service.

To elaborate, it does work, but the intended exit flow doesn't work since the foreground process group is signaled by tty, and Monado itself has signal handlers which immediately stop the IPC server. I was considering patching it out, but I could also just add this functionality to Monado when stopping the IPC server... perhaps I should do that instead?

server/driver/wivrn_session.cpp
ImSapphire: In stop_application there is a log when it calls xrt_session_request_exit

xytovl:
> > It should work with Ctrl+C too, applications are spawned through systemd user service.
> 
> To elaborate, it does work, but the intended exit flow doesn't work since the foreground process group is signaled by tty, and Monado itself has signal handlers which immediately stop the IPC server. I was considering patching it out, but I could also just add this functionality to Monado when stopping the IPC server... perhaps I should do that instead?

Monado is right to handle the signal, maybe we could upstream the logic to gracefully terminate clients.
In case there is anything to take from WiVRn's code, ping me in the merge request so that I can clarify the relicensing.

xytovl:
We can add a `setsid()` right after the fork and Ctrl-C works just fine.

ImSapphire:
> Monado is right to handle the signal, maybe we could upstream the logic to gracefully terminate clients. In case there is anything to take from WiVRn's code, ping me in the merge request so that I can clarify the relicensing.

I created [!2987](https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2987), I don't think relicensing is required as I wrote the graceful session exit code

xytovl:
I feel like we should upstream more than just the graceful close.
We have in WiVRn a timeout mechanism after `xrt_session_request_exit`, it may be possible to upstream that. The same code could be used for application stop and server stop.

ImSapphire:
I think the timeout pretty much functions the same in that MR, at least in the view of client applications. When the IPC server shuts down and disconnects clients, they just lose the instance

xytovl:
The Monado MR is fine, but I think the WiVRn/Monado interaction is a bit clumsy and we may have to rethink how it works.

Currently, stop application calls `xrt_session_request_exit`, and then starts a timer for the hard kill. Maybe this timer should be part of the `ipc_client_state` and managed in Monado directly.

Let's just add the `setsid` in the current pull request, and maybe revisit if we can move more logic upstream.

xytovl (approved):

ImSapphire:
oops, I forgot to amend the commit message... oh well

PR #1073, opened 2026-08-26
Add patches and dashboard setting for maximum steamvr_lh pose prediction

ImSapphire:


dashboard/qml/SettingsPage.qml
```diff
@@ -132,6 +132,31 @@ Kirigami.ScrollablePage {
...
+                Controls.Slider {
+                    id: lh_max_extrapolation
+                    Layout.fillWidth: true
+                    from: 0.0
+                    to: 100.0
+                    stepSize: 1.0
+                    value: Settings.lhMaxExtrapolationEnabled
+                }
+                Controls.Label {
+                    text: lh_max_extrapolation.value.toString() + "ns"
```
xytovl: It's "ms" I suppose.

The field has to be localizable, I don't know if any language has plurals for units, you might have to use `i18np("%1ms", "%1ms", lh_max_extrapolation.value)`

server/driver/configuration.h
```diff
@@ -57,6 +57,7 @@ struct configuration
 	std::vector<std::string> application;
 	bool debug_gui = false;
 	bool use_steamvr_lh = false;
+	std::optional<int64_t> lh_max_extrapolation;
```
xytovl, line 60: should be in docs/configuration.md

dashboard/settings.cpp
```diff
@@ -417,6 +419,36 @@ void Settings::set_steamVrLh(const bool & value)
...
+	{
+		m_jsonSettings["lh-max-extrapolation"] = value;
+	}
+	else
+	{
+		m_jsonSettings.erase("lh-max-extrapolation");
+	}
+
+	if (old != value)
+		lhStickDeadzoneChanged();
```
xytovl: bad copy-paste

patches/monado/0010-a-math-Add-m_relation_history_get_clamped.patch
```diff
@@ -0,0 +1,136 @@
...
+ 
+diff --git a/src/xrt/auxiliary/math/m_relation_history.h b/src/xrt/auxiliary/math/m_relation_history.h
+index 86e08ef5e..5196abd63 100644
+--- a/src/xrt/auxiliary/math/m_relation_history.h
++++ b/src/xrt/auxiliary/math/m_relation_history.h
+@@ -83,12 +83,33 @@ m_relation_history_push_with_motion_estimation(struct m_relation_history *rh,
+  * Read-only operation - doesn't remove anything from the buffer or anything like that - you can call this as often
+  * as you want.
+  *
++ * @param max_extrapolation_ns Maximum time the pose may be extrapolated to. Negative means extrapolation is unlimited.
```
xytovl: Why not use INT64_MAX instead of a magic value?

server/driver/configuration.h
xytovl: I actually wonder if this has to be user-visible configuration.
A good default would be sufficient, and further tuning would still be possible through the environment variable.

I believe 20ms is low, but I don't have hardware to test.

dashboard/qml/SettingsPage.qml
ImSapphire: oops, I keep making that mistake!

server/driver/configuration.h
ImSapphire: A user I had test previously said 20ms felt pretty good, but some people may want more or less prediction, especially dancers, so I left it tunable here (especially since I don't have hardware to fine-tune it myself, but will soon)

ImSapphire: An environment variable is also pretty inaccessible for those running through dashboard

patches/monado/0010-a-math-Add-m_relation_history_get_clamped.patch
ImSapphire: I think I briefly thought about that and forgot but that's a good idea, I'll change it

server/driver/configuration.h
xytovl: For other extrapolation settings, once we introduced the polynomial interpolator we didn't have requests for processing on controller data.
Lighthouse devices should be the same, such a low limit on extrapolation shows there's an issue and we're not fixing the actual issue. Providing GUI space for a workaround just looks wrong.

ImSapphire: steamvr_lh devices go through Monado's `m_relation_history` which does pretty simple interpolation/extrapolation rather than our polynomial interpolator, so movements end up heavily exaggerated, especially with the predicted display time being far (i.e. 30 or more milliseconds) in the future

xytovl: It would be interesting to see how lighthouse would behave if we upstream the polynomial interpolator. It is still unable to predict far into the future, we only use it on top of headset-provided predictions.

Back to the original question, we can add the setting, as only lighthouse users will see it, but I strongly dislike adding settings if the reason is that we can't fix a bug or don't know what the correct value should be.

dashboard/qml/SettingsPage.qml
```diff
@@ -132,6 +132,31 @@ Kirigami.ScrollablePage {
...
+                Controls.Slider {
+                    id: lh_max_extrapolation
+                    Layout.fillWidth: true
+                    from: 0.0
+                    to: 100.0
+                    stepSize: 1.0
+                    value: Settings.lhMaxExtrapolationEnabled
+                }
+                Controls.Label {
+                    text: i18np("%1ms", "%1ms", lh_max_extrapolation.value)
```
xytovl: actually that's still insufficient, [i18ncp](https://api.kde.org/klocalizedstring.html#i18ncp) may be needed, so context can explain we're talking about milliseconds

patches/monado/0011-d-steamvr_lh-Add-LH_MAX_EXTRAPOLATION_MS.patch
```diff
@@ -0,0 +1,45 @@
...
+
+diff --git a/src/xrt/drivers/steamvr_lh/device.cpp b/src/xrt/drivers/steamvr_lh/device.cpp
+index fc3da6066..1f62f7dff 100644
+--- a/src/xrt/drivers/steamvr_lh/device.cpp
++++ b/src/xrt/drivers/steamvr_lh/device.cpp
+@@ -53,6 +53,7 @@
+ DEBUG_GET_ONCE_BOOL_OPTION(lh_emulate_hand, "LH_EMULATE_HAND", true)
+ DEBUG_GET_ONCE_FLOAT_OPTION(lh_override_ipd, "LH_OVERRIDE_IPD_MM", -1.0f)
+ DEBUG_GET_ONCE_BOOL_OPTION(lh_standby_on_exit, "LH_STANDBY_ON_EXIT", false)
++DEBUG_GET_ONCE_NUM_OPTION(lh_max_extrapolation_ms, "LH_MAX_EXTRAPOLATION_MS", 20)
```
galister: i think go for 50ms by default.

we know that 80~100ms is too high, but even a regular wired index can easily go over 20ms if FPS is low.

and the sample size of users that prefer 20ms is 1 person

server/driver/configuration.h
galister: i think i might be worth a shot just as a patch in the meantime. even if latency is low (60~70ms) and unclamped, it's very apparent that wivrn and lh devices move differently

patches/monado/0010-a-math-Add-m_relation_history_get_clamped.patch
```diff
@@ -0,0 +1,137 @@
...
+ 			*out_relation = {};
+ 			return M_RELATION_HISTORY_RESULT_INVALID;
+ 		}
+@@ -112,17 +117,27 @@ m_relation_history_get(const struct m_relation_history *rh,
+ 			// (pose-prediction)
+ 			// Output flags match the most recent buffer entry.
+ 			int64_t diff_prediction_ns = static_cast<int64_t>(at_timestamp_ns) - rh->impl.back().timestamp;
++			diff_prediction_ns = max_extrapolation_ns < INT64_MAX
++			                         ? MIN(diff_prediction_ns, max_extrapolation_ns)
++			                         : max_extrapolation_ns;
```
xytovl: You don't need the ternary, and the case `max_extrapolation_ns == INT64_MAX` is wrong. It's always `MIN(diff_prediction_ns, max_extrapolation_ns)`

patches/monado/0010-a-math-Add-m_relation_history_get_clamped.patch
```diff
@@ -0,0 +1,137 @@
...
+ 			*out_relation = it->relation;
+ 			return M_RELATION_HISTORY_RESULT_EXACT;
+ 		}
+@@ -132,9 +147,16 @@ m_relation_history_get(const struct m_relation_history *rh,
+ 			// (an edge case where somebody asks for a really old pose and we do our best)
+ 			// Output flags are the same as the input flags for the history entry we use
+ 			int64_t diff_prediction_ns = static_cast<int64_t>(at_timestamp_ns) - rh->impl.front().timestamp;
++			diff_prediction_ns = max_extrapolation_ns < INT64_MAX
++			                         ? -MIN(abs(diff_prediction_ns), max_extrapolation_ns)
++			                         : max_extrapolation_ns;
```
xytovl: Same comment for ternary operator

patches/monado/0011-d-steamvr_lh-Add-LH_MAX_EXTRAPOLATION_MS.patch
```diff
@@ -0,0 +1,45 @@
...
+ void
+ Device::get_pose(uint64_t at_timestamp_ns, xrt_space_relation *out_relation)
+ {
+-	m_relation_history_get(this->relation_hist, at_timestamp_ns, out_relation);
++	static const int64_t max_extrapolation_ns = []() -> int64_t {
++		int64_t raw = debug_get_num_option_lh_max_extrapolation_ms();
++		if (raw < 0) {
++			U_LOG_W("LH_MAX_EXTRAPOLATION_MS value below zero; defaulting to unlimited extrapolation");
++			return 0;
++		}
```
xytovl: 0 is no extrapolation with the INT64_MAX change.
negative (or just -1) for unlimited extrapolation sounds reasonable and doesn't need to emit a warning.

xytovl:
The patch is OK for merging, but I still believe it's a workaround and a proper fix is possible.

ImSapphire:
I'm not sure what a proper fix would look like without improving m_relation_history upstream to handle extrapolation far into the future better, which I'm not really equipped to do

galister:
where does the polynomial interpolator live right now? i want to look into this

ImSapphire:
> where does the polynomial interpolator live right now? i want to look into this

https://github.com/WiVRn/WiVRn/blob/master/server/driver/polynomial_interpolator.h

galister:
thanks, gonna see what I can do about this. would be awesome if the devices would just move together. (and maybe put a 80ms hard cap on everything on top of that)

xytovl:
When using streamed data, extrapolation is done in 2 steps: on the headset with the estimated required extrapolation then on server using polynomial interpolator to match the precise timestamp (both interpolation and extrapolation).

We tried to use polynomial interpolator only initially and on headset request +0 samples, this didn't produce satisfying results. The headset runtime may have more accurate models or use different inputs, it is unlikely that 80ms polynomial extrapolation on lighthouse devices would work well.

galister:
I'm still curious about how it does if at least we do weighted prediction using multiple samples. currently the monado one only uses the latest sample, which I think can easily cause the jumpiness that many people complain about.

galister:
it seems to be worse at both high and low latency. let's just merge this for now. :adhesive_bandage:

PR #1076, opened 2026-08-29
Add support for extended refresh rates on Quest 3

Localcrap:
# Add support for extended refresh rates on Quest 3

## Problem

On Quest 3 with HorizonOS 2.7 and later, the display accepts any integer refresh rate from 72 through 207 Hz through `xrRequestDisplayRefreshRateFB` (240 Hz when display scaling is enabled in developer mode), but `xrEnumerateDisplayRefreshRatesFB` still only reports rates up to 120 Hz, so applications have no way to discover the extended rates.

Reference: [Set Display Refresh Rates — Meta Horizon OS docs](https://developers.meta.com/horizon/documentation/native/android/mobile-display-refresh-rate/)

## This change

**Commit 1 — client:**

- **`hmd_traits`**: new per-headset `extra_refresh_rates` field. Quest 3 (`eureka`) gets `144, 165, 185, 200, 207` when the HorizonOS version is at least 2.7, reusing the existing `ro.hzos.build.display_name` detection, plus `240` when `debug.oculus.forceDisplayScaling` is `1` (240 Hz additionally requires the user to set `debug.oculus.refreshRate`). The field participates in the existing `debug.wivrn.*` / `WIVRN_*` override mechanism, so any headset can get extra rates with e.g. `adb shell setprop debug.wivrn.extra_refresh_rates "144,165"`.
- **`xr::session::get_refresh_rates()`**: inserts the extra rates at the sorted position in the enumerated list. This is the single choke point: the in-headset picker, saved-config validation, and `headset_info_packet` all pick the extended rates up automatically, so the server-side auto adjuster and D-Bus see them too. No protocol changes.
- Requesting a rate outside of the enumerated list may fail with `XR_ERROR_DISPLAY_REFRESH_RATE_UNSUPPORTED_FB`, which `set_refresh_rate` already logs.

**Commit 2 — server:**

- In automatic mode, start at the highest available rate not exceeding 120 Hz instead of the highest available rate; `refresh_rate_adjuster` still increases it if the application sustains it. Explicit rate selection is unaffected.

## Testing

- Quest 3 (HorizonOS 2.7, RTX 5070 Ti server, NVENC AV1):
  - 165 Hz: works, noticeably smoother than 120 Hz
  - 207 Hz: runs but shows white-noise artifacts at the top of the image — looks like an encoder/decoder issue at this fps, still investigating (happy to hear ideas)
  - 240 Hz: reachable with the props set (not fully validated yet)
- Non-extended paths (enumerated rates, other headsets) are unchanged: the merge is a no-op when `extra_refresh_rates` is empty.

<details>
<summary>Build verification</summary>

- Server + Linux client build clean with `-Werror` (GCC 16 / Fedora 44)
- Android APK (`assembleRelease`) builds and was tested on device
- `clang-format` clean
</details>

---

<sup>Acknowledgement: this pull request was developed with AI assistance (opencode, GLM). All code was reviewed, built and tested on hardware by the submitter.</sup>

### DIFF
--- a/client/hmd_traits.cpp
+++ b/client/hmd_traits.cpp
@@ -26,6 +26,7 @@
 #include <boost/pfr.hpp>
 #include <cassert>
 #include <cctype>
+#include <concepts>
 #include <cstdlib>
 #include <cstring>
 #include <format>
@@ -140,6 +141,42 @@ std::optional<std::unordered_set<std::string>> env(std::string_view name)
 	return std::unordered_set<std::string>{std::from_range, utils::split(*values, ",")};
 }
 
+template <>
+std::optional<std::vector<float>> env(std::string_view name)
+{
+	const auto value = env<std::string>(name);
+	if (not value)
+		return std::nullopt;
+
+	std::vector<float> result;
+	for (const auto & i: utils::split(*value, ","))
+	{
+		try
+		{
+			result.push_back(std::stof(i));
+		}
+		catch (std::exception & e)
+		{
+			spdlog::warn("failed to parse {} (value: {}): {}", name, *value, e.what());
+		}
+	}
+
+	return result;
+}
+
+// A trait field that is a list of values, overridden as a whole
+template <typename T>
+concept trait_container = requires(T & t, const typename T::value_type & v) {
+	typename T::value_type;
+	t.begin();
+	t.end();
+	t.insert(t.end(), v);
+} and (std::same_as<typename T::value_type, std::string> or std::same_as<typename T::value_type, float>);
+
+// Set-like containers support removing values prefixed with '-'
+template <typename T>
+concept trait_set = trait_container<T> and requires(T & t, const typename T::value_type & v) { t.erase(v); };
+
 void hmd_traits::init()
 {
 #ifndef NDEBUG
@@ -170,6 +207,28 @@ void hmd_traits::init()
 		// Quest breaks spec and does not support grip_surface for ext/hand_interaction_ext
 		hand_interaction_grip_surface = false;
 
+		// HorizonOS version, e.g. "2.7.0"
+		std::optional<std::pair<int, int>> hzos_version;
+		if (auto v = get_property("ro.hzos.build.display_name"))
+		{
+			auto digits = utils::split(*v, ".");
+			try
+			{
+				hzos_version = std::pair{std::stoi(digits.at(0)), std::stoi(digits.at(1))};
+			}
+			catch (...)
+			{
+				spdlog::warn("Failed to parse Horizon OS version {}", *v);
+			}
+		}
+
+		auto hzos_at_least = [&hzos_version](int major, int minor) {
+			if (not hzos_version)
+				return false;
+			auto [actual_major, actual_minor] = *hzos_version;
+			return actual_major > major or (actual_major == major and actual_minor >= minor);
+		};
+
 		if (device == "monterey") // Quest 1
 		{
 			panel_width_override = 1440;
@@ -191,6 +250,18 @@ void hmd_traits::init()
 			panel_width_override = 2064; // Quest 3
 			controller_profile = "meta-quest-touch-plus";
 			usb_net = true;
+
+			// The runtime may not report extended refresh rates even though
+			// it accepts them: Quest 3 supports any integer rate from 72 to
+			// 207 Hz with HorizonOS 2.7 and later, up to 240 Hz with display
+			// scaling enabled
+			// https://developers.meta.com/horizon/documentation/native/android/mobile-display-refresh-rate/
+			if (hzos_at_least(2, 7))
+			{
+				extra_refresh_rates = {144, 165, 185, 200, 207};
+				if (get_property("debug.oculus.forceDisplayScaling").value_or("") == "1")
+					extra_refresh_rates.push_back(240);
+			}
 		}
 		else if (device == "panther") // Quest 3S
 		{
@@ -198,20 +269,8 @@ void hmd_traits::init()
 			controller_profile = "meta-quest-touch-plus";
 		}
 
-		if (auto v = get_property("ro.hzos.build.display_name"))
-		{
-			auto digits = utils::split(*v, ".");
-			try
-			{
-				auto major = std::stoi(digits.at(0));
-				auto minor = std::stoi(digits.at(1));
-				usb_net = major > 2 or (major == 2 and minor >= 6);
-			}
-			catch (...)
-			{
-				spdlog::warn("Failed to parse Horizon OS version {}", *v);
-			}
-		}
+		if (hzos_version)
+			usb_net = hzos_at_least(2, 6);
 	}
 
 	else if (model == "Lynx-R1")
@@ -357,16 +416,26 @@ void hmd_traits::init()
 		                }
 	                },
 
-	                [](std::string_view name, std::unordered_set<std::string> & field) {
-		                if (auto val = env<std::unordered_set<std::string>>(name))
+	                [](std::string_view name, trait_container auto & field) {
+		                using field_type = std::remove_cvref_t<decltype(field)>;
+
+		                if (auto val = env<field_type>(name))
 		                {
-			                for (const auto & i: *val)
+			                spdlog::info("\t{} override", name);
+
+			                if constexpr (trait_set<field_type>)
 			                {
-				                if (i.starts_with("-"))
-					                field.erase(i.substr(1));
-				                else
-					                field.insert(i);
+				                // values prefixed with '-' are removed from the set
+				                for (const auto & i: *val)
+				                {
+					                if (i.starts_with("-"))
+						                field.erase(i.substr(1));
+					                else
+						                field.insert(field.end(), i);
+				                }
 			                }
+			                else
+				                field = std::move(*val);
 		                }
 
 		                for (const auto & i: field)
--- a/client/xr/session.cpp
+++ b/client/xr/session.cpp
@@ -19,6 +19,7 @@
 
 #include "session.h"
 
+#include "application.h"
 #include "details/enumerate.h"
 #include "openxr/openxr.h"
 #include "utils/contains.h"
@@ -285,11 +286,13 @@ float xr::session::get_current_refresh_rate()
 
 std::vector<float> xr::session::get_refresh_rates()
 {
+	std::vector<float> rates;
+
 	if (xrEnumerateDisplayRefreshRatesFB)
 	{
 		try
 		{
-			return xr::details::enumerate<float>(xrEnumerateDisplayRefreshRatesFB, id);
+			rates = xr::details::enumerate<float>(xrEnumerateDisplayRefreshRatesFB, id);
 		}
 		catch (...)
 		{
@@ -297,7 +300,17 @@ std::vector<float> xr::session::get_refresh_rates()
 		}
 	}
 
-	return {};
+	// The runtime may not enumerate all supported rates, add the rates from
+	// hmd_traits, requesting unenumerated values is allowed on supported devices
+	for (float rate: application::get_hmd_traits().extra_refresh_rates)
+	{
+		// the enumerated list is sorted, insert at the sorted position
+		auto pos = std::ranges::lower_bound(rates, rate);
+		if (pos == rates.end() or *pos != rate)
+			rates.insert(pos, rate);
+	}
+
+	return rates;
 }
 
 void xr::session::set_refresh_rate(float refresh_rate)
--- a/server/driver/wivrn_session.cpp
+++ b/server/driver/wivrn_session.cpp
@@ -476,12 +476,25 @@ bool wivrn_session::connected()
 	return connection->is_active();
 }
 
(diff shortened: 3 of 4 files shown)


client/hmd_traits.cpp
```diff
@@ -190,28 +213,20 @@ void hmd_traits::init()
...
 			controller_profile = "meta-quest-touch-plus";
-			usb_net = true;
+
+			// The runtime may not report extended refresh rates even though it
+			// accepts them: Quest 3 supports any integer rate from 72 to 207 Hz
+			// with HorizonOS v2.7+, up to 240 Hz with display scaling enabled
+			// https://developers.meta.com/horizon/documentation/native/android/mobile-display-refresh-rate/
+			extra_refresh_rates = {144, 165, 185, 200, 207};
+			if (get_property("debug.oculus.forceDisplayScaling").value_or("") == "1")
+				extra_refresh_rates.push_back(240);
```
galister: you're automatically assuming that anyone with a quest3 will automatically be on horizon os 2.7+

which is not going to be the case, as many people prefer to use their devices offline and not receive updates.

galister: display scaling means you'll get lower image resolution as you increase frame rate. you definitely don't want to include it as something that can be automatically scaled to.

xytovl:
The OpenXR specification clearly says
> The runtime must return XR_ERROR_DISPLAY_REFRESH_RATE_UNSUPPORTED_FB if displayRefreshRate is not either 0.0f or one of the values enumerated by [xrEnumerateDisplayRefreshRatesFB](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#xrEnumerateDisplayRefreshRatesFB)

Whatever Meta describes in https://developers.meta.com/horizon/documentation/native/android/mobile-display-refresh-rate/ is in violation of the specification.

client/hmd_traits.cpp
Localcrap: good point, should be easy enough to implement

xytovl: It's already present on master

README.md
```diff
@@ -138,6 +138,22 @@ See [building](docs/building.md) for building the [dashboard](docs/building.md#d
...
+adb shell setprop debug.oculus.forceDisplayScaling 1
+adb shell setprop debug.oculus.refreshRate 240
+```
+Above 207 Hz the display bandwidth is insufficient for a full-resolution frame, the image is upscaled by the headset at the cost of image quality. Higher refresh rates also increase the required bitrate and encoding load.
+
+On other headsets, additional rates can be made available in the refresh rate list with (replace with the rates supported by your headset):
+```bash
+adb shell setprop debug.wivrn.extra_refresh_rates "144,165"
+```
+
```
xytovl: We don't need additional docs for this, especially on the main readme.

client/hmd_traits.cpp
```diff
@@ -374,6 +389,19 @@ void hmd_traits::init()
 			                spdlog::info("\t{}: {}", name, i);
 		                }
 	                },
+
+	                [](std::string_view name, std::vector<float> & field) {
```
xytovl: This would be cleaner if we could use a C++20 concept and unify with the `std::unordered_set`.

It's the same thing: iterable with one element.
The env function would be trickier, we can live with the duplication.

server/driver/wivrn_session.cpp
```diff
@@ -476,12 +476,25 @@ bool wivrn_session::connected()
...
+
+	float rate = 0;
+	for (float i: headset_info.available_refresh_rates)
+	{
+		if (i > max_auto_rate)
+			break;
+		rate = i;
+	}
+
+	return rate == 0 ? headset_info.available_refresh_rates.front() : rate;
```
xytovl, line 497: This should be on its own commit

client/xr/session.cpp
```diff
@@ -300,16 +301,67 @@ std::vector<float> xr::session::get_refresh_rates()
 	return {};
 }
 
+std::vector<float> xr::session::get_refresh_rates()
+{
+	std::vector<float> rates = enumerated_refresh_rates();
+
+	// xrEnumerateDisplayRefreshRatesFB is deprecated and may not report rates
+	// above 120 Hz even though the runtime accepts them, add the rates from
+	// hmd_traits (requesting unenumerated values is allowed on supported devices)
```
xytovl: The extension is in no way deprecated, Meta decided to ignore the standard but I assume other vendors will not.

client/xr/session.h
```diff
@@ -46,6 +48,12 @@ class session : public utils::handle<XrSession, xrDestroySession>
 	PFN_xrEnumerateDisplayRefreshRatesFB xrEnumerateDisplayRefreshRatesFB = nullptr;
 	PFN_xrRequestDisplayRefreshRateFB xrRequestDisplayRefreshRateFB = nullptr;
 
+	// Deadline of a pending request for a rate outside of the enumerated list
+	std::optional<std::chrono::steady_clock::time_point> refresh_rate_request_deadline;
+	float requested_refresh_rate = 0;
```
xytovl: I don't see why we should track requests, the runtime would log itself relevant stuff if it ignored the request

client/xr/session.cpp
```diff
@@ -300,16 +301,67 @@ std::vector<float> xr::session::get_refresh_rates()
 	return {};
 }
 
+std::vector<float> xr::session::get_refresh_rates()
+{
+	std::vector<float> rates = enumerated_refresh_rates();
```
xytovl: It doesn't have to be a separate method

client/xr/session.cpp
```diff
@@ -300,16 +301,67 @@ std::vector<float> xr::session::get_refresh_rates()
...
+	// xrEnumerateDisplayRefreshRatesFB is deprecated and may not report rates
+	// above 120 Hz even though the runtime accepts them, add the rates from
+	// hmd_traits (requesting unenumerated values is allowed on supported devices)
+	for (float rate: application::get_hmd_traits().extra_refresh_rates)
+	{
+		if (not utils::contains(rates, rate))
+			rates.push_back(rate);
+	}
+
+	std::ranges::sort(rates);
```
xytovl: With https://en.cppreference.com/cpp/algorithm/lower_bound you can replace both contains and sort (insertion in the middle is ok-ish as the number of elements will remain low)

client/hmd_traits.cpp
Localcrap: modified it slightly so the logic can be reused, hope it's ok :)

xytovl:
As I said in my first comment, the feature relies on a spec violation by Meta.
I will discuss with people from Monado who know better about OpenXR/Khronos, I think it is very wrong to implement features like this as it defeats the purpose of a specification.

My position would be to close the merge request and wait for Meta to expose this in a different manner.

Localcrap:
i guess it's not optimal, but i can see why the team at meta decided to go against returning a list looking like [72,73,74...240], would probably break a lot of existing applications (like the gui in wivrn feks)

zoeleu:
> As I said in my first comment, the feature relies on a spec violation by Meta. I will discuss with people from Monado who know better about OpenXR/Khronos, I think it is very wrong to implement features like this as it defeats the purpose of a specification.
> 
> My position would be to close the merge request and wait for Meta to expose this in a different manner.

Regardless of Meta's ability to actually implement features properly, should users really be restricted by that? This is something rather trivial to implement, and has a huge benefit to perceived latency by the user. It seems rather silly to not implement this, given that it's a single API call.

client/hmd_traits.cpp
```diff
@@ -170,6 +207,28 @@ void hmd_traits::init()
 		// Quest breaks spec and does not support grip_surface for ext/hand_interaction_ext
 		hand_interaction_grip_surface = false;
 
+		// HorizonOS version, e.g. "2.7.0"
+		std::optional<std::pair<int, int>> hzos_version;
+		if (auto v = get_property("ro.hzos.build.display_name"))
```
galister, line 212: perhaps use the incremental build number instead? so that it's an unmistakable integer comparison. relying on something called "display_name" is asking for trouble

xytovl:
There's many reasons to not merge it:
* Deviating from the spec sets a precedent, which will make testing harder as more stuff becomes device-specific
* The implementation requires hardcoding of values for the headset, which might change in the future
* The decoder already struggles with 120FPS, I doubt we can sustain decode under 7ms (for 144Hz)

galister:
the higher refresh rates would only really make sense when used with space warp. except space warp doesn't work. so you'd get the same lower frame rate but with more battery usage and heat.

client/hmd_traits.cpp
Localcrap: ro.build.version.incremental? might be out of scope for this pr as it's used for the usb_hmd feature as well

xytovl: I can't find any source from Meta about build versions, all it says is "Extended refresh rates require HorizonOS v2.7 or later.".

USB networking is even worse, every reference was removed from Meta docs, https://web.archive.org/web/20260709221756/https://developers.meta.com/horizon/documentation/android-apps/usb-networking/ does not specify a version.

Localcrap:
streaming surprisingly works ok up to 200hz. 
there is a small amount of flickering happening on my hmd at 185hz and above, not enough to bother but it's  somewhat visible.
the wivrn gui  also struggles above 144hz, think it has to do with passthrough, as when im in a game the performance recovers.

this is with a 5800x and 5070ti  so YMMV

zoeleu:
> There's many reasons to not merge it:
> 
> * Deviating from the spec sets a precedent, which will make testing harder as more stuff becomes device-specific
> * The implementation requires hardcoding of values for the headset, which might change in the future
> * The decoder already struggles with 120FPS, I doubt we can sustain decode under 7ms (for 144Hz)

Idk about you, but it definitely worked for me. It requires hard coding for now. Also, there's nothing wrong with quirks. Do you really think manufacturers think of WiVRn when making changes to their headsets? And this is the Quest 3, one of the most common headsets out there, not some hyper specific rare device.

galister:
> streaming surprisingly works ok up to 200hz. there is a small amount of flickering happening on my hmd at 185hz and above, not enough to bother but it's somewhat visible. the wivrn gui also struggles above 144hz, think it has to do with passthrough, as when im in a game the performance recovers.

i would be interested in seeing some perf charts. someone posted one in #alvr channel of LVRA discord yesterday where with a 205hz selection the hmd is actually at /5 at 41fps. i'm curious how WiVRn compares there

Localcrap:
made some really quick and dirty mesurements, ignore 72 and 144hz as i only ran those for a couple of seconds.
also had some bad wifi, so try to ignore the 200ms network spikes as well

## Stage latency (median / p95, ms)

| Stage | 165 Hz | 200 Hz |
|---|---|---|
| encode | 3.28 / 3.41 | 3.33 / 3.85 |
| network | −0.24 / 1.56 | −0.22 / 5.22 |
| decode | 5.91 / 7.58 | 5.41 / 7.88 |
| to_blit | 15.50 / 20.07 | 14.56 / 17.21 |
| blit→display | 13.46 / 17.04 | 13.20 / 21.12 |
| **glass-to-glass** | 34.5 / 42.1 | 33.9 / 42.0 |


pacing
stage_cdfs
stages_over_time

uotlaf:
> There's many reasons to not merge it:
[...]
>     * The implementation requires hardcoding of values for the headset, which might change in the future

Sorry, but why not have an option to select a custom target/maximum framerate?

galister:
> Sorry, but why not have an option to select a custom target/maximum framerate?

The spec actually doesn't have anything against this.

> The runtime must return XR_ERROR_DISPLAY_REFRESH_RATE_UNSUPPORTED_FB if displayRefreshRate is not either 0.0f or one of the values enumerated by [xrEnumerateDisplayRefreshRatesFB](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#xrEnumerateDisplayRefreshRatesFB). A display refresh rate of 0.0f indicates the application has no preference.
> 
> Note that this is only a request and does not guarantee the system will switch to the requested display refresh rate.

It could be locked behind the extended values checkbox.

TJC47:
> > There's many reasons to not merge it:
> > [...]
> > * The implementation requires hardcoding of values for the headset, which might change in the future
> 
> Sorry, but why not have an option to select a custom target/maximum framerate?

In my opinion that's actually an even better solution since you can then select any refresh rate between 72 and 207 so you can really tune it in if you want to.

Localcrap:
> > > There's many reasons to not merge it:
> > > [...]
> > > 
> > > * The implementation requires hardcoding of values for the headset, which might change in the future
> > 
> > 
> > Sorry, but why not have an option to select a custom target/maximum framerate?
> 
> In my opinion that's actually an even better solution since you can then select any refresh rate between 72 and 207 so you can really tune it in if you want to.

would be really cool, thought about implementing it in this pr, but would require a bigger refactor of the codebase.

if anybody wants to test it out, i have a this commit https://github.com/WiVRn/WiVRn/commit/2c14e96c111ab85ceeaaf2b4ef52d192c7a87321 that replaces the buttons with a slider,  then you just have to find somewhere to hard code the values 72-207

PR #1083, opened 2026-08-30
Update es locale

BlackCherryCat:
Solves #442

### DIFF
--- a/locale/es/wivrn-dashboard.po
+++ b/locale/es/wivrn-dashboard.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: WiVRn-dashboard\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-08-07 21:01+0200\n"
+"POT-Creation-Date: 2026-08-29 22:24+0200\n"
 "PO-Revision-Date: 2025-11-19 20:20+0100\n"
 "Last-Translator: Álvaro De Quinta <blackcherry@danwin1210.de>\n"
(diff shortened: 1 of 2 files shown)


locale/es/wivrn-dashboard.po
```diff
@@ -599,12 +599,41 @@ msgstr ""
...
+msgid "SteamVR max pose extrapolation"
+msgstr "Extrapolación máxima de pose de SteamVR"
+
+#: dashboard/qml/SettingsPage.qml:152
+#, kde-format
+msgctxt "value display for SteamVR max pose extrapolation"
+msgid "%1ms"
+msgid_plural "%1ms"
+msgstr[0] "%1ms"
+msgstr[1] "%1ms%n"
```
xytovl: that's weird

BlackCherryCat:
In this case, I would recommend to remove the plural string, it's not necessary

xytovl:
> In this case, I would recommend to remove the plural string, it's not necessary

I mean, the "%n" part is weird.
We can't remove the plural, some languages might have different forms for units.

BlackCherryCat:
oh, in that case, I added "%n" because Qt Linguist told me that its necessary:
`Translation does not contain the necessary %n/%Ln place marker: %1ms`

xytovl:
Qt Linguist must be wrong, other tools using ki18ncp don't seem to have %n markers:
https://github.com/KDE/kunitconversion/blob/master/po/es/kunitconversion6.po
https://github.com/KDE/kunitconversion/blob/master/src/binary_data.cpp

Lokalize should be more accurate for KDE wrappers.

Issue #921, opened 2026-05-17
[Feature Request]: Vulkan AV1 Support


EmersonBiggons:
### Feature description

Add AV1 support for vulkan encode

### Reason for the feature

VAAPI seems less performant (at least compared to Vulkan in WiVRn), H264 looks awful and H265 10Bit while looking perfect is a bit slow to encode and decode compared to AV1. I don't have the best number to show this, but I can achieve significantly better output with AV1 over H265 with less bit rate. Especially at higher framerates

My use case is rather niche, I use 2 VR headsets at once and every bit counts to whether the game is playable or spinning christmas lights. Especially since we both play at 120hz and reasonable resolutions for visibilty and performance. AV1 would let me drop the bitrate considerably which would help in every regard.

### Additional context

#857 This closed issue seems to be semi related as it was a direct result of AV1 not being available.

xytovl:
There was an attempt in https://github.com/ZokuTe/WiVRn/tree/vulkan/av1 but is incomplete.

Contributions would be very welcome.

PR #1087, opened 2026-09-01
Add Vulkan AV1 video encoder

schlegp:
Implements AV1 encoding through VK_KHR_video_encode_av1, so the Vulkan encoder can now use AV1 instead of falling back to VAAPI.

The extension is requested at device creation and the videoEncodeAV1 device feature is enabled alongside it. The encoder asks the driver for its AV1 capabilities and configures itself from them, rather than assuming fixed values.

Addresses #921.

Tested successfully on RX 7900XTX. Not yet tested on Intel/Nvidia.

### DIFF
--- a/server/encoder/video_encoder_vulkan_av1.cpp
+++ b/server/encoder/video_encoder_vulkan_av1.cpp
@@ -1,0 +1,596 @@
+/*
+ * WiVRn VR streaming
+ * Copyright (C) 2024  Patrick Nicolas <patricknicolas@laposte.net>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#include "video_encoder_vulkan_av1.h"
+
+#include "encoder/encoder_settings.h"
+#include "util/u_logging.h"
+#include "utils/wivrn_vk_bundle.h"
+
+#include <algorithm>
+#include <cstring>
+#include <limits>
+
+namespace
+{
+uint32_t align_div(uint32_t value, uint32_t divisor)
+{
+	return (value + divisor - 1) / divisor;
+}
+
+uint8_t bits_for(uint32_t value)
+{
+	uint8_t bits = 0;
+	while (value > 0)
+	{
+		value >>= 1;
+		++bits;
+	}
+	return std::max<uint8_t>(bits, 1);
+}
+} // namespace
+
+wivrn::video_encoder_vulkan_av1::video_encoder_vulkan_av1(
+        wivrn::vk_bundle & vk,
+        const vk::VideoCapabilitiesKHR & video_caps,
+        const vk::VideoEncodeCapabilitiesKHR & encode_caps,
+        const vk::VideoEncodeAV1CapabilitiesKHR & encode_av1_caps,
+        uint8_t stream_idx,
+        const encoder_settings & settings) :
+        video_encoder_vulkan(vk, video_caps, encode_caps, stream_idx, settings)
+{
+	if (not std::ranges::any_of(vk.device_extensions, [](std::string_view ext) { return ext == VK_KHR_VIDEO_ENCODE_AV1_EXTENSION_NAME; }))
+	{
+		throw std::runtime_error("Vulkan video encode AV1 extension not available");
+	}
+
+	if (settings.bit_depth != 8 && settings.bit_depth != 10)
+		throw std::runtime_error("av1 encoder supports 8-bit or 10-bit only");
+
+	configure_from_caps(encode_av1_caps);
+
+	color_config = {
+	        .flags = {
+	                .mono_chrome = 0,
+	                .color_range = 1,
+	                .separate_uv_delta_q = 0,
+	                .color_description_present_flag = 1,
+	        },
+	        .BitDepth = static_cast<uint8_t>(settings.bit_depth),
+	        .subsampling_x = 1,
+	        .subsampling_y = 1,
+	        .reserved1 = 0,
+	        .color_primaries = STD_VIDEO_AV1_COLOR_PRIMARIES_BT_709,
+	        .transfer_characteristics = STD_VIDEO_AV1_TRANSFER_CHARACTERISTICS_BT_709,
+	        .matrix_coefficients = STD_VIDEO_AV1_MATRIX_COEFFICIENTS_BT_709,
+	        .chroma_sample_position = STD_VIDEO_AV1_CHROMA_SAMPLE_POSITION_COLOCATED,
+	};
+
+	seq_header = {
+	        .flags = {
+	                .still_picture = 0,
+	                .reduced_still_picture_header = 0,
+	                .use_128x128_superblock = superblock_size == 128 ? 1u : 0u,
+	                .enable_filter_intra = 0,
+	                .enable_intra_edge_filter = 0,
+	                .enable_interintra_compound = 0,
+	                .enable_masked_compound = 0,
+	                .enable_warped_motion = 0,
+	                .enable_dual_filter = 0,
+	                .enable_order_hint = 1,
+	                .enable_jnt_comp = 0,
+	                .enable_ref_frame_mvs = 0,
+	                .frame_id_numbers_present_flag = 0,
+	                .enable_superres = 0,
+	                .enable_cdef = 0,
+	                .enable_restoration = 0,
+	                .film_grain_params_present = 0,
+	                .timing_info_present_flag = 0,
+	                .initial_display_delay_present_flag = 0,
+	        },
+	        .seq_profile = STD_VIDEO_AV1_PROFILE_MAIN,
+	        .frame_width_bits_minus_1 = static_cast<uint8_t>(bits_for(aligned_extent.width - 1) - 1),
+	        .frame_height_bits_minus_1 = static_cast<uint8_t>(bits_for(aligned_extent.height - 1) - 1),
+	        .max_frame_width_minus_1 = static_cast<uint16_t>(aligned_extent.width - 1),
+	        .max_frame_height_minus_1 = static_cast<uint16_t>(aligned_extent.height - 1),
+	        .delta_frame_id_length_minus_2 = 0,
+	        .additional_frame_id_length_minus_1 = 0,
+	        .order_hint_bits_minus_1 = static_cast<uint8_t>(order_hint_bits - 1),
+	        .seq_force_integer_mv = STD_VIDEO_AV1_SELECT_INTEGER_MV,
+	        .seq_force_screen_content_tools = STD_VIDEO_AV1_SELECT_SCREEN_CONTENT_TOOLS,
+	        .reserved1 = {},
+	        .pColorConfig = &color_config,
+	        .pTimingInfo = nullptr,
+	};
+
+	operating_point = {};
+	operating_point.flags.decoder_model_present_for_this_op = 0;
+	operating_point.flags.low_delay_mode_flag = 1;
+	operating_point.flags.initial_display_delay_present_for_this_op = 0;
+	operating_point.operating_point_idc = 0;
+	operating_point.seq_level_idx = STD_VIDEO_AV1_LEVEL_2_0;
+	operating_point.seq_tier = 0;
+	operating_point.decoder_buffer_delay = 0;
+	operating_point.encoder_buffer_delay = 0;
+	operating_point.initial_display_delay_minus_1 = 0;
+
+	const uint32_t sb_cols = align_div(aligned_extent.width, superblock_size);
+	const uint32_t sb_rows = align_div(aligned_extent.height, superblock_size);
+	const uint32_t mi_cols = align_div(aligned_extent.width, 4);
+	const uint32_t mi_rows = align_div(aligned_extent.height, 4);
+	const bool uniform_tile_spacing = bool(std_flags & vk::VideoEncodeAV1StdFlagBitsKHR::eUniformTileSpacingFlagSet);
+
+	mi_col_starts = {0u, static_cast<uint16_t>(mi_cols)};
+	mi_row_starts = {0u, static_cast<uint16_t>(mi_rows)};
+	if (!tile_widths_sb.empty())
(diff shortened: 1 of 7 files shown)


xytovl:
This looks very nice and the code looks genuine.

Have you tested the behaviour if you intentionally drop some random packets? This will cause the ref_slot in encode_info_next to not be the last transmitted frame, on H.265 the metadata to specify this is much more complex than setting an index. If AV1 is that's simple that's cool.

server/encoder/video_encoder_vulkan_av1.cpp
```diff
@@ -0,0 +1,596 @@
+/*
+ * WiVRn VR streaming
+ * Copyright (C) 2024  Patrick Nicolas <patricknicolas@laposte.net>
```
xytovl, line 3: I might own some copyright if the code is copy-pasted from H.264, but you should be in the list as well

server/encoder/video_encoder_vulkan_av1.cpp
```diff
@@ -0,0 +1,596 @@
...
+        const vk::VideoEncodeCapabilitiesKHR & encode_caps,
+        const vk::VideoEncodeAV1CapabilitiesKHR & encode_av1_caps,
+        uint8_t stream_idx,
+        const encoder_settings & settings) :
+        video_encoder_vulkan(vk, video_caps, encode_caps, stream_idx, settings)
+{
+	if (not std::ranges::any_of(vk.device_extensions, [](std::string_view ext) { return ext == VK_KHR_VIDEO_ENCODE_AV1_EXTENSION_NAME; }))
+	{
+		throw std::runtime_error("Vulkan video encode AV1 extension not available");
+	}
```
xytovl, line 59: I see it's the same on other encoders, but `vk.has_device_ext(VK_KHR_VIDEO_ENCODE_AV1_EXTENSION_NAME)` looks better

xytovl:
If you know enough about AV1, is it possible to encode more efficiently the alpha channel? It is currently a separate half-resolution colour stream, as decoders don't support monochrome.

A first step would be to encode as monochrome if it's mandatory for decoders to support, a much more complex solution (if AV1 allows it) is to be able to encode as a 4th channel, but this may need changes to the rest of the pipeline.

schlegp:
> Have you tested the behaviour if you intentionally drop some random packets? This will cause the ref_slot in encode_info_next to not be the last transmitted frame, on H.265 the metadata to specify this is much more complex than setting an index. If AV1 is that's simple that's cool.

Thanks, didn't even think of testing that. 
I set it up now to drop 2% of frames and despite the performance obviously degrading, the image was still correct. 
The work heavily draws from https://github.com/hgaiser/pixelforge, I sadly am not yet well versed in the whole topic.

> is it possible to encode more efficiently the alpha channel? 

This is my first venture into this topic, so I'll take this as a reason to do research into it. Sounds like a worthwhile improvement if it can be done.

server/encoder/video_encoder_vulkan_av1.cpp
xytovl: > The work heavily draws from https://github.com/hgaiser/pixelforge, I sadly am not yet well versed in the whole topic.

If you copied portions of the code, we would have to include the BSD license alongside the GPL one.
If there is no copy of copyrightable content, adding your name is sufficient.

xytovl:
I have briefly tested on a Radeon 8060S and a client running on Linux/Monado simulated driver.
After a few seconds I get an `ERROR [encode] Timeout on stream 0` followed by a `VK_ERROR_DEVICE_LOST`.

Kernel messages are not very helpful:
```
[294432.080536] amdgpu 0000:c4:00.0: Dumping IP State
[294432.082195] amdgpu 0000:c4:00.0: Dumping IP State Completed
[294432.082208] amdgpu 0000:c4:00.0: [drm] AMDGPU device coredump file has been created
[294432.082210] amdgpu 0000:c4:00.0: [drm] Check your /sys/class/drm/card0/device/devcoredump/data
[294432.082212] amdgpu 0000:c4:00.0: ring vcn_unified_1 timeout, signaled seq=68343, emitted seq=68345
[294432.082213] amdgpu 0000:c4:00.0:  Process wivrn-server pid 1157532 thread wivrn-server pid 1157538
[294432.082215] amdgpu 0000:c4:00.0: Starting vcn_unified_1 ring reset
[294432.335878] amdgpu 0000:c4:00.0: Register(1) [regUVD_POWER_STATUS] failed to reach value 0x00000001 != 0x00000002n
[294432.588354] amdgpu 0000:c4:00.0: Register(1) [regUVD_RB_RPTR] failed to reach value 0x00000080 != 0x00000000n
[294432.848328] amdgpu 0000:c4:00.0: Register(1) [regUVD_POWER_STATUS] failed to reach value 0x00000001 != 0x00000002n
[294432.850316] amdgpu 0000:c4:00.0: Ring vcn_unified_1 reset succeeded
[294432.850318] amdgpu 0000:c4:00.0: [drm] device wedged, but no recovery needed
```

This could be a driver bug, vulkan validation layers fail to find errors.

ergo3d:
> Implements AV1 encoding through VK_KHR_video_encode_av1, so the Vulkan encoder can now use AV1 instead of falling back to VAAPI.

YOU are HERO! thanks a lot!
just tested on 9070xt arch linux
works perfectly fine, smooth and stable, no any broken frame compared to vaapi

CherrieTheShifter:
Yes This Please 🙏

myghi63:
I've tested it here and it works really well on my RX 7800xt. It looks more crisp than h265 overall.

I might try to use it soon on a RTX 4090 to see if it works or not.

myghi63:
If anyone is having crashes with AMD, make sure the PC is not running the version 7.2.6 of the linux kernel

xytovl:
On kernel 7.2.6, mesa 26.2.2 I still get crashes.

the failed test this time: connect, stream something, works for at least 10s. stop the game, start it again, the server crashes on resume (encoder timeout)

Edit: I misread the previous comment, I had issues before and after kernel 7.2.6

server/encoder/video_encoder_vulkan_av1.cpp
schlegp: No code copied, took the H265 implementation and cross referenced how to do this in AV1, will add my Name to the header.

schlegp:
> I have briefly tested on a Radeon 8060S and a client running on Linux/Monado simulated driver. After a few seconds I get an `ERROR [encode] Timeout on stream 0` followed by a `VK_ERROR_DEVICE_LOST`.
> 
> Kernel messages are not very helpful:
> 
> ```
> [294432.080536] amdgpu 0000:c4:00.0: Dumping IP State
> [294432.082195] amdgpu 0000:c4:00.0: Dumping IP State Completed
> [294432.082208] amdgpu 0000:c4:00.0: [drm] AMDGPU device coredump file has been created
> [294432.082210] amdgpu 0000:c4:00.0: [drm] Check your /sys/class/drm/card0/device/devcoredump/data
> [294432.082212] amdgpu 0000:c4:00.0: ring vcn_unified_1 timeout, signaled seq=68343, emitted seq=68345
> [294432.082213] amdgpu 0000:c4:00.0:  Process wivrn-server pid 1157532 thread wivrn-server pid 1157538
> [294432.082215] amdgpu 0000:c4:00.0: Starting vcn_unified_1 ring reset
> [294432.335878] amdgpu 0000:c4:00.0: Register(1) [regUVD_POWER_STATUS] failed to reach value 0x00000001 != 0x00000002n
> [294432.588354] amdgpu 0000:c4:00.0: Register(1) [regUVD_RB_RPTR] failed to reach value 0x00000080 != 0x00000000n
> [294432.848328] amdgpu 0000:c4:00.0: Register(1) [regUVD_POWER_STATUS] failed to reach value 0x00000001 != 0x00000002n
> [294432.850316] amdgpu 0000:c4:00.0: Ring vcn_unified_1 reset succeeded
> [294432.850318] amdgpu 0000:c4:00.0: [drm] device wedged, but no recovery needed
> ```
> 
> This could be a driver bug, vulkan validation layers fail to find errors.

I have tried adhering closer to the Vulkan spec, can you try if this fixed anything?
Otherwise we'll have to dig deeper what the root cause is.

server/encoder/video_encoder_vulkan_av1.cpp
schlegp: Done.

xytovl:
I have the same error on 60a73af8574747f06b974fcdb7750fb7377826bf

This may be a GPU bug, I think we could merge the pull request if we disable automatic selection of vulkan/AV1 and gather some feedback.

The logic could be in `encoder_settings.cpp` to skip the `has_vk` check if `config.name.empty() and codec == av1`

schlegp:
Done, only selecting av1 always uses VAAPI now. Vulkan must be specifically configured.

server/encoder/video_encoder_vulkan_av1.cpp
```diff
@@ -0,0 +1,637 @@
...
+uint8_t bits_for(uint32_t value)
+{
+	uint8_t bits = 0;
+	while (value > 0)
+	{
+		value >>= 1;
+		++bits;
+	}
+	return std::max<uint8_t>(bits, 1);
+}
```
xytovl, line 41: That's [bit_width](https://en.cppreference.com/cpp/numeric/bit_width)

```cpp
std::max(std::bit_width(bits), 1);
```

server/encoder/video_encoder_vulkan_av1.cpp
```diff
@@ -0,0 +1,637 @@
...
+	operating_point = {};
+	operating_point.flags.decoder_model_present_for_this_op = 0;
+	operating_point.flags.low_delay_mode_flag = 1;
+	operating_point.flags.initial_display_delay_present_for_this_op = 0;
+	operating_point.operating_point_idc = 0;
+	operating_point.seq_level_idx = static_cast<uint8_t>(level);
+	operating_point.seq_tier = 0;
+	operating_point.decoder_buffer_delay = 0;
+	operating_point.encoder_buffer_delay = 0;
+	operating_point.initial_display_delay_minus_1 = 0;
```
xytovl: You can use the same style as seq_header:
```cpp
operating_point = {
  .flags = {
    .decoder_model_present_for_this_op = 0,
    .low_delay_mode_flag = 1,
    .initial_display_delay_present_for_this_op = 0,
  },
  .seq_level_idx = static_cast<uint8_t>(level),
...
};
```

server/encoder/video_encoder_vulkan_av1.cpp
```diff
@@ -0,0 +1,637 @@
...
+	const uint32_t sb_rows = align_div(aligned_extent.height, superblock_size);
+	const uint32_t mi_cols = align_div(aligned_extent.width, 4);
+	const uint32_t mi_rows = align_div(aligned_extent.height, 4);
+	const bool uniform_tile_spacing = bool(std_flags & vk::VideoEncodeAV1StdFlagBitsKHR::eUniformTileSpacingFlagSet);
+
+	// A single tile is enough for the resolutions WiVRn streams, but AV1 caps a tile at 4096
+	// samples wide (annex A.3), so warn rather than silently emit a non conforming stream.
+	if (sb_cols * superblock_size > 4096)
+		U_LOG_W("AV1: %u samples wide exceeds the maximum width of a single tile, stream may not be decodable",
+		        aligned_extent.width);
```
xytovl: If we know this is invalid, we may as well throw an error and fail here.

There's `check_video_size` in `encoder_settings.cpp` if we really want to gracefully downgrade but I don't think it is necessary

server/encoder/video_encoder_vulkan_av1.h
```diff
@@ -0,0 +1,102 @@
...
+	uint32_t single_reference_name_mask = 0;
+	uint32_t max_single_reference_count = 0;
+	uint32_t max_q_index = 0;
+	uint32_t min_q_index = 0;
+	uint32_t superblock_size = 64;
+	uint8_t order_hint_bits = 7;
+	// Reference name (LAST_FRAME + index) used for the single reference of inter frames
+	int32_t ref_name_index = 0;
+	bool rate_control_enabled = false;
+	bool use_gop_info = false;
```
xytovl: Could be combined with `gop_info` using an `std::optional`

server/encoder/video_encoder_vulkan_av1.cpp
```diff
@@ -0,0 +1,637 @@
...
+	}
+}
+
+void * wivrn::video_encoder_vulkan_av1::encode_info_next(uint32_t frame_num, size_t slot, std::optional<int32_t> ref_slot)
+{
+	const bool has_ref = ref_slot.has_value();
+	const bool is_keyframe = not has_ref;
+
+	// Initialize reference name slot indices to -1 (no reference)
+	std::fill(reference_name_slot_indices.begin(), reference_name_slot_indices.end(), -1);
```
xytovl:
```cpp
std::ranges::fill(reference_name_slot_indice, -1);
```

ergo3d:
> If anyone is having crashes with AMD, make sure the PC is not running the version 7.2.6 of the linux kernel

arch 7.2.6-273-linux-tkg there is no problem

xytovl:
I have tried to narrow down what causes the crash:

running the client on Monado simulated make it much more likely to happen when exiting/restarting XR apps

running on a Quest 3, disconnecting while the app (WayVR in this case) is running, then reconnecting almost always triggers the encoder timeout.

In both cases it is at or shortly after stream re-initialization.

server/encoder/video_encoder_vulkan_av1.cpp
schlegp: Done.

server/encoder/video_encoder_vulkan_av1.h
schlegp: Done.

server/encoder/video_encoder_vulkan_av1.cpp
schlegp: Done.

server/encoder/video_encoder_vulkan_av1.cpp
schlegp: Done.

schlegp:
I tried recreating your failure case, but on my machine it wouldn't break.

7900 XTX / mesa 26.2.3 / kernel 7.2.6-arch2-1 / AV1 10-bit

xytovl:
Let's assume the failure is an issue on RDNA3.5 driver side. The code looks good enough for merging.

All that's left to do is resolve the conflicts. Then either I'll squash on merge or if you think there's useful steps, make meaningful commits.

Once it is merged we can ask users to test on different hardware/software and see if we can enable by default for some.

grzesjam:
I have tested it with Nvidia 5070Ti / `Nvidia (Open) 615.71.09` / `7.2.7 #1-NixOS` and this does **not** work

Vulkan H265 works flawlessly (actually image quality is much better compared to NVENC H265)
Vulkan AV1 game starts, but image never appears, just like if video stream never started (NVENC AV1 works, but image quality is a little lower than Vulkan H265, and I'm getting random blocky corruption of image)

PR #1094, opened 2026-09-06
Add Headless Benchmarking/Profiling

damaestro:
Add the ability to run headless benchmarking for regression and performance analysis. Specifically paying attention to usefulness for a git bisect workflow.


server/encoder/video_encoder_vulkan.cpp
```diff
@@ -1053,25 +1054,39 @@ std::optional<wivrn::video_encoder::data> wivrn::video_encoder_vulkan::encode(ui
...
 		std::cerr << "device.getQueryPoolResults: " << vk::to_string(res) << std::endl;
 	}
 
-	if (auto s = ts_pool.collect(encode_slot))
+	// present_image() submits the encode, so the fence wait above spans it. Report only the part
+	// before the GPU started, matching wait_gpu elsewhere. Untrimmed without GPU timestamps.
+	const auto encode_ts = ts_pool.collect(encode_slot);
+	const int64_t wait_gpu_end_ns =
+	        encode_ts ? std::clamp(encode_ts->begin_ns, wait_begin_ns, wait_end_ns) : wait_end_ns;
+	wivrn::trace::cpu_slice(wivrn::trace::cpu_track::encoder, "wait_gpu", wait_begin_ns, wait_gpu_end_ns, frame_index, stream_idx);
```
xytovl: Do we need this trace? vulkan encode will do everything on GPU and we already measure those steps.
The CPU time spent waiting here actually measures the CPU time to submit the commands on the other stream.

patches/monado/0012-c-null-make-the-frame-rate-configurable.patch
xytovl, line 1: This is a candidate for upstreaming

tools/perfetto/wivrn_session.py
```diff
@@ -0,0 +1,621 @@
...
+        subprocess.run(["pkill", "-x", "hello_xr"], capture_output=True, check=False)
+        stop(server)
+        stop(monado)
+        # Undo the systemd user-manager env injection and clear failed app units.
+        systemctl_user("unset-environment", "XR_RUNTIME_JSON")
+        systemctl_user("reset-failed", "wivrn-application-*.service")
+        shutil.rmtree(scratch, ignore_errors=True)
+
+    try:
+        say("starting headless Monado (client runtime)")
```
xytovl: I think it's better to build the Monado runtime from clean (unpatched) sources.
The inproc build should work well enough so we don't have to manage separate processes.

patches/monado/0012-c-null-make-the-frame-rate-configurable.patch
damaestro: Agreed. I'll upstream.

tools/perfetto/wivrn_session.py
damaestro: Yeah... I went back and forth on that. I'll make that change after getting an upstream MR opened for the null compositor patch.

xytovl: Until 2dd07da12d2a2b62403f8d4186d5af9b5257dd5b we had a Monado build in CI to perform some testing.
I think the same approach would be fine: a possibly different Monado version to build the simulated OpenXR runtime. With `XRT_FEATURE_SERVICE_SYSTEMD=OFF` it simplifies execution a little, removing some process management and race conditions.

tools/perfetto/wivrn_session.py
```diff
@@ -0,0 +1,621 @@
...
+    Returns (Toolchain, []) or (None, [human-readable missing pieces])."""
+    build_dir = args.build_dir
+    server = find_binary(args.server, build_dir / "server/wivrn-server", "wivrn-server")
+    client = find_binary(args.client, build_dir / "bin/wivrn", "wivrn")
+    monado_service = find_binary(
+        args.monado_service,
+        build_dir / "bench-runtime/src/xrt/targets/service/monado-service",
+        "monado-service",
+    )
+    hello_xr = find_binary(args.hello_xr, None, "hello_xr")
```
xytovl: I would have the whole OpenXR app be an argument, so you provide `["hello_xr", "-g", "Vulkan2"]` to the script, defaults can ensure that it remains the same.
Of course there will be quoting issues, python should have something smart to process them so you can pass `--xr-app "hello_xr -g Vulkan2"`

This allows easily testing with a different binary

damaestro: I'll update to make the test runtime configurable with some default search for `hello_xr`, `xrgears`, etc.

server/encoder/video_encoder_vulkan.cpp
damaestro: Yeah, I think I got too apples-to-apples on the results parsing and this ended up as just fluff. Removing.

patches/monado/0012-c-null-make-the-frame-rate-configurable.patch
damaestro: https://gitlab.freedesktop.org/monado/monado/-/merge_requests/3006

damaestro:
When slicing out the Monado upstream patch I also added the ability to specify screen resolution to get us closer to simulating an actual HMD.

RTX 3080 Ti, driver 610.57.04, 90 Hz target, 15s/run

#### 1920x1472 per eye

| Codec | Metric | Vulkan | NVENC |
|---|---|--:|--:|
| H264 | encode call mean | 3.98 ms | 4.23 ms |
| H264 | encode call p99 | 5.57 ms | 6.03 ms |
| H264 | sustained rate | 168.1/s | 167.1/s |
| H265 | encode call mean | 13.61 ms | 4.85 ms |
| H265 | encode call p99 | 15.65 ms | 6.42 ms |
| H265 | sustained rate | 68.7/s | 167.6/s |

#### 2064x2208 per eye

| Codec | Metric | Vulkan | NVENC |
|---|---|--:|--:|
| H264 | encode call mean | 6.30 ms | 6.85 ms |
| H264 | encode call p99 | 9.27 ms | 9.51 ms |
| H264 | sustained rate | 147.9/s | 123.7/s |
| H265 | encode call mean | 22.78 ms | 7.56 ms |
| H265 | encode call p99 | 24.47 ms | 10.90 ms |
| H265 | sustained rate | 40.9/s | 55.2/s |

`rate` = combined stereo slices/sec (2× per-eye fps).

tools/perfetto/wivrn_session.py
xytovl: I am still not convinced in having cmake changes to build a clean Monado runtime.

Building the dummy runtime is completely independent from the WiVRn build, I believe the commit itself can be fixed in the script.

What do you see as advantages of building the dummy runtime as part of WiVRn ?
I would instead have a Monado checkout + build managed by the script.

damaestro: My preference has been to manage all of the build options in the same place so there is one authoritative place to go look at the flavors. But I agree this is tangential to the server side enough we could push it even further away.

I'll see about ripping any cmake changes this weekend.

PR #1095, opened 2026-09-06
Raise the vulkan h265 encoder's minimum coding block to 16x16

damaestro:
## Raise the vulkan h265 encoder's minimum coding block to 16x16

When working on a headless profiling solution, I did a test comparing nvenc and Vulkan for h265 and h264. A big performance difference popped out. This PR gets h265 encoding via Vulkan on par with nvenc.

The SPS allowed coding blocks down to 8x8. On NVIDIA that costs 4x the encode time, enough to miss the frame deadline at headset resolution.

Measured on an RTX 3080 Ti, 1920x1472 per eye at 90 Hz, 5 runs x 20 s:

| | `encodeVideoKHR` p50 | sustained fps |
| --- | --- | --- |
| before | 13.423 ms | 36.5 |
| after | **3.175 ms** | **90.1** |
| nvenc h265, same GPU | 4.356 ms | 90.0 |

Encode time is linear in pixel count either way. No other SPS/PPS knob moves it: transform hierarchy depth, transform skip, SAO, AMP, strong intra smoothing, merge candidates, bit depth and encode quality level are all within noise.

h264 is unaffected — fixed 16x16 macroblocks, no minimum CU.

### Reproducing

Needs the benchmark harness from #1094

```bash
tools/perfetto/encoder_bench.py bootstrap
BENCH=~/.cache/wivrn-encoder-bench/bin/encoder_bench.py

$BENCH run --encoder vulkan --codec h265   # before
git cherry-pick <this commit>
$BENCH run --encoder vulkan --codec h265   # after
$BENCH report                              # both, with the delta
```

### Not covered

- Compression efficiency is unmeasured. Dropping the 8x8 partition costs quality at a fixed bitrate; the old configuration was dropping frames instead.
- Tested on NVIDIA only.

### DIFF
--- a/server/encoder/video_encoder_vulkan_h265.cpp
+++ b/server/encoder/video_encoder_vulkan_h265.cpp
@@ -134,7 +134,7 @@ wivrn::video_encoder_vulkan_h265::video_encoder_vulkan_h265(
                 .bit_depth_luma_minus8 = static_cast<uint8_t>(settings.bit_depth - 8),
                 .bit_depth_chroma_minus8 = static_cast<uint8_t>(settings.bit_depth - 8),
                 .log2_max_pic_order_cnt_lsb_minus4 = 4,      // arbitrary
-                .log2_min_luma_coding_block_size_minus3 = 0, // keep it 0, related values are filled in create()
+                .log2_min_luma_coding_block_size_minus3 = 0, // set in create(), with the related values
                 .num_short_term_ref_pic_sets = 0,
                 .num_long_term_ref_pics_sps = 0,
                 .pcm_sample_bit_depth_luma_minus1 = 0,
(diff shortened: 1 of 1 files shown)


xytovl:
That's very interesting, have you checked if it affects decoding times?

galister:
Will see if I can test this on AMD tomorrow.

xytovl:
On RDNA3.5/Quest 3 there isn't any significant difference.

damaestro:
## Decoder-side impact: 16x16 vs 8x8 minimum CTB

Captured on my Vive Focus 3, `perfetto` + `hal`/`video`/`gfx` atrace categories. I'm still getting used to what to capture and how to interpret the HMD side stuff. So far I've just been :eyes: on the WebUI tracing random stuff. Here is a 1s slice with the 8x8 on the left and the 16x16 on the right:

screenshot-perfetto-hmd-traces-decoder-20260906


This looks nearly identical to me.

Admittedly I didn't push the video stream with anything complex. I used `hello_xr -g Vulkan2` and just made sure I was looking around and moving the controllers for the 15 second period.

I'll open a PR with a perfetto config that can be ran via `adb` and some basic analysis tools sometime soon. On the production firmware we don't get all of the tracepoints that give you really detailed data. Maybe this will give me enough of a reason to add some to the client code :smile: 

| Metric | 8x8 (before) | 16x16 (after) | Δ |
|---|---:|---:|---:|
| `media.codec` total Running time | 1532.6 ms | 1509.9 ms | **-1.5%** |
| `VideoDecMsgThre` (HAL decode worker) Running time | 585.9 ms | 574.0 ms | -2.0% |
| `emptyBuffer` HIDL call latency (mean) | 0.1652 ms | 0.1634 ms | -1.1% |
| `fillBuffer` HIDL call latency (mean) | 0.1582 ms | 0.1532 ms | -3.1% |
| Decode call count (`emptyBuffer` n) | 2686 | 2684 | ~identical |

client/application.cpp
```diff
@@ -1031,7 +1031,7 @@ void application::initialize_vulkan()
 	                          .instance = *vk_instance,
 	                  },
 #ifndef NDEBUG
-	                  debug_utils_found
+	                  debug_utils_found and vk_device.getDispatcher()->vkSetDebugUtilsObjectNameEXT
```
xytovl: I don't know what to say about that.

We already have https://github.com/WiVRn/WiVRn/blob/116e9e5454eb36b459bbb819548b90c0df722f47/client/hmd_traits.cpp#L177 for Quest 1

Generic https://github.com/WiVRn/WiVRn/blob/116e9e5454eb36b459bbb819548b90c0df722f47/client/hmd_traits.h#L60

The missing function while the extension is reported is a vulkan violation, so it kind of make sense to use a quirk mechanism. Probing the pointer is less hardcoding and will fix itself if HTC fixes the runtime. It also risk causing crashes if `vkSetDebugUtilsObjectNameEXT` is called without checking the allocator's property.

damaestro: Yeah -- I'll pull this from this PR and ask you fix it the best way :smile: A trait for this hardware makes sense to me.

I was unable to use the debug build from this PR without this guard.

damaestro: https://github.com/WiVRn/WiVRn/pull/1098

Issue #873, opened 2026-04-13
[Bug]: Recent compositor changes introduced severe performance issues on devices equipped with NVIDIA GPUs.


ZokuTe:
### Describe the bug

> This issue is similar to: #865 

Specifically, the manifestations include:

- Random occurrence of past frames
- Random screen tearing
- Strange input latency
- Motion-to-photon latency appears normal, but feels very high
- Frame rate data appears normal, but the actual frame rate is very low
- Causes severe motion sickness
- Random see-through frames when Alpha channel is disabled
- Minor visual corruption when Alpha channel is enabled
- Incorrect cropping edges when Alpha channel is enabled
- The output from the Vulkan encoder is green and completely unusable.

### Steps to Reproduce

normal use of the latest master branch.

### WiVRn Version

#1f6e237

### Installation method

built from sources.

### Server Config

```json
{
  "application": [
    "wayvr"
  ],
  "debug-gui": false,
  "encoder": {
    "codec": "h265",
    "encoder": "nvenc"
  },
  "hid-forwarding": false,
  "openvr-compat-path": "/opt/xrizer",
  "use-steamvr-lh": false
}
```

### Server Logs

```log
WiVRn 1f6e237f starting
Service published: zokute-l
Client connected
Server started, PID 26602
 INFO [ipc_server_main_common] WiVRn '1f6e237f' starting up...
Launching "wayvr"
 INFO [comp_vulkan_formats_log] Supported formats:
	VK_FORMAT_R16G16B16A16_UNORM: true
	VK_FORMAT_R16G16B16A16_SFLOAT: true
	VK_FORMAT_R16G16B16_UNORM: false
	VK_FORMAT_R16G16B16_SFLOAT: false
	VK_FORMAT_R8G8B8A8_SRGB: true
	VK_FORMAT_B8G8R8A8_SRGB: true
	VK_FORMAT_R8G8B8_SRGB: false
	VK_FORMAT_R8G8B8A8_UNORM: true
	VK_FORMAT_B8G8R8A8_UNORM: true
	VK_FORMAT_R8G8B8_UNORM: false
	VK_FORMAT_B8G8R8_UNORM: false
	VK_FORMAT_R5G6B5_UNORM_PACK16: true
	VK_FORMAT_R32_SFLOAT: true
	VK_FORMAT_D32_SFLOAT: true
	VK_FORMAT_D16_UNORM: true
	VK_FORMAT_X8_D24_UNORM_PACK32: true
[...]

PR #1096, opened 2026-09-06
app_pacer: add completion headroom for streaming compositors

azorlu80:
## Problem

Fixes the periodic flashing/stutter on network-streamed sessions (#873; severe on high-pipeline-latency / NVIDIA encode setups).

The app pacer wakes the app at:

```
wake = display - (cpu_time + gpu_time + compositor_time + min_margin)
```

`compositor_time` is Monado's `extra_ns` (the compositor picks the frame up ~`extra_ns` before display), so the `compositor_time` terms cancel and only `min_margin` (~1.5 ms) of slack is left between the app finishing a frame and the compositor picking it up.

On a **native** compositor `extra_ns` is vsync-stable and 1.5 ms is fine. On a **streaming** compositor `extra_ns` is the full pipeline latency (encode + network + decode, ~50 ms on my Wi-Fi link) and it drifts frame-to-frame (phase adjustment, latency re-estimation), routinely eating the 1.5 ms. The frame then misses its pickup → a stale frame is shown → the next completed frame drops the missed one (`wait_for_scheduled_free`, "Dropping old missed frame") → periodic flash/stutter.

## Fix

Add up to one display period of completion headroom, phased in only for the part of `extra_ns` that exceeds one period:

```
safety = clamp(compositor_time - period, 0, period)
```

- **Native** (`extra_ns <= period`) ⇒ `safety == 0` ⇒ behaviour unchanged, zero risk.
- **Streaming** ⇒ completion margin grows from ~1.5 ms to ~15 ms. The term is added symmetrically to both the `min_ready` display placement and the wake-up lead, so it does **not** add display latency — it only moves the wake-up earlier by the same amount it reserves.

The "is the app itself too slow to hit its slot" test is kept on a separate app-limited estimate that caps `compositor_time` at one period, so it does not false-fire on every streamed frame (where `extra_ns > period` always).

## Evidence (RTX 4050 mobile / Quest 3, 72 Hz Wi-Fi stream)

Raw per-frame pacer log, 14.8k frames:

- `gpu_time` avg **0.12 ms** (never > period), `cpu_time` avg **5.3 ms** — app side is healthy.
- `compositor_time` avg **51 ms** (range 8–82 ms), 99.8% > period — this is the pipeline latency, not app work.

Completion-vs-pickup margin (does the app finish before the compositor's pickup deadline?):

| build | frames finishing **after** pickup deadline (dropped) |
|-------|------|
| stock  | **11.9%** |
| with fix | **0.3%** |

On-head dropped frames: **~7/s → ~1/s** (the remaining ~1/s is app-side stalls, unrelated to pacing). Frame-completion lateness averages **~8 ms** — no added latency.

### DIFF
--- a/server/driver/app_pacer.cpp
+++ b/server/driver/app_pacer.cpp
@@ -133,13 +133,39 @@ void app_pacer::predict(int64_t now_ns,
 	get_frame(frame_id) = {.frame_id = frame_id};
 
 	// no need to lock, called in same thread as writes
-	auto min_ready = now_ns + cpu_time + gpu_time + compositor_time;
+	const int64_t margin_ns = int64_t(U_TIME_1MS_IN_NS * min_margin_ms.val);
+
+	// compositor_time is Monado's extra_ns: the gap from the compositor main loop waking
+	// up to display. The compositor picks this app's frame up ~extra_ns before display,
+	// and the wake-up below (display - (cpu + gpu + compositor_time + margin)) makes the
+	// compositor_time terms cancel, leaving only min_margin (~1.5 ms) of slack between
+	// the app finishing a frame and the compositor picking it up. On a native compositor
+	// extra_ns is vsync-stable and 1.5 ms is enough; on a network-streaming compositor
+	// extra_ns is the pipeline latency (encode + network + decode) and it drifts while a
+	// frame is in flight (phase adjustment, latency re-estimation), routinely eating the
+	// 1.5 ms. The frame then misses its pickup, a stale frame is shown, and the next
+	// completed frame drops the missed one -> periodic flashing/stutter.
+	//
+	// Add up to one display period of completion headroom, phased in only for the part of
+	// extra_ns that exceeds one period. On a native compositor extra_ns <= period, so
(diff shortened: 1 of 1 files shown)


xytovl:
Can you please explain it in your own words, LLMs just write plenty of garbage.

azorlu80:
Closing this. The completion-margin change reduces the compositor 'Dropping old missed frame' rate in an A/B test, but it did not resolve the flashing/stutter on my hardware and caused an image-quality regression on a full session. Needs more validation before it's a real fix for #873. Withdrawing until then.

xytovl:
Just to let you know, I prefer a good explanation of the mechanism that causes the issue than an AI coded pull request.

LLMs tend to write a lot of text and are never straight to the point. Maintaining such a code becomes impossible as those comments become out of date and clutter the code. It's possible to have context for the fix in commit messages and pull request description.

app pacing is tricky because there's feedback loops, interactions between GPU, CPU scheduler. On NVIDIA https://github.com/WiVRn/WiVRn/pull/1095 (using vulkan video encode) might help fix the issue if it makes compositor times more stable.

PR #1101, opened 2026-09-09
po: Add Azerbaijani language

jamalkamaladdin:
This PR adds an Azerbaijani (az) translation.

New file locale/az/wivrn.po: 271 strings, all filled in.
New file locale/az/wivrn-dashboard.po: 169 strings, all filled in, including 5 plural entries.

PR #1102, opened 2026-09-09
po: fill in the az translator header

jamalkamaladdin:
The Azerbaijani catalogs added in #1101 carry a placeholder in the Last-Translator header.

locale/az/wivrn.po and locale/az/wivrn-dashboard.po: Last-Translator set to the translator name and address.


xytovl:
It's fine to set them, but nobody cares, there's git log/blame to know about this

PR #1109, opened 2026-09-13
Fix typos

antonkesy:
Fixes typos in comments and string literals.
Should not affect behavior (except for fixed stdout)


dashboard/qml/SettingsPage.qml
```diff
@@ -282,7 +282,7 @@ Kirigami.ScrollablePage {
                     Layout.fillWidth: true
                 }
                 Controls.Button {
-                    text: i18nc("browse to choose the OpenVR compatility to use", "Browse")
+                    text: i18nc("browse to choose the OpenVR compatibility to use", "Browse")
```
xytovl, line 285: Can you update the translation files too? The context is used as a key

PR #1117, opened 2026-09-16
Update monado

galister:
saw that there were some big changes lately and didn't want the work to pile up later on

1 commit per breaking change, as per usual

there were some major changes to `steamvr_lh_create_devices` upstream so i went & tested lh devices just to be sure.


server/driver/wivrn_session.cpp
```diff
@@ -154,41 +153,41 @@ wivrn::wivrn_session::wivrn_session(std::unique_ptr<wivrn_connection> connection
 		U_LOG_W("Disregard lighthousedb / chaperone related error messages from the lighthouse driver. These are irrelevant in case of WiVRn.");
 		U_LOG_W("If getting a SIGSEGV right after this, you are likely using an unsupported SteamVR version!");
 		U_LOG_W("=====================");
-		if (steamvr_lh_create_devices(nullptr, &lhdevs) == XRT_SUCCESS)
+		int original_count = static_xdev_count;
+		if (steamvr_lh_create_devices(nullptr, this) == XRT_SUCCESS)
```
xytovl, line 157: Are we certain this won't create an hmd device?

galister: it attempts to create a hmd device, if there is one actually plugged in. does not assign it any role, though.

same behavior as before, from what I can tell. on the wivrn side we just ignore that XRT_DEVICE_TYPE_ so nothing happens to it

xytovl (approved):

PR #1121, opened 2026-09-17
feat(client): add per-posture player height offset setting

moi952:
Hi,

I've been using WiVRn for quite a while now and it's honestly great, but one thing that always bugged me is that some games just don't give you any way to adjust your height — annoying when the floor calibration is a bit off, or when you want to play seated but don't want the game to think you shrunk.

So I added a height adjustment setting:

- Two separate presets, one for standing and one for seated, each remembering its own offset in cm so you don't have to redo it every time you switch
- A posture toggle — set your default (used whenever a stream connects), or just override it for the current session if you want to switch mid-game without messing with your usual setup
- A slider to fine-tune it live while streaming, no need to disconnect
- Everything's applied client-side on the poses sent to the PC, so no protocol changes, nothing to touch server-side

First time digging into the tracking code so if something doesn't fit how you'd want it done, happy to adjust. Thanks for the project either way, it's been a staple for me.


galister:
wayvr and motoc are both utilities that let you do this btw

PR #1123, opened 2026-09-18
client: support XR_META_foveation_eye_tracked

galister:
our current eye-tracked foveation impl uses `XR_EXT_eye_gaze_interaction` which produces a processed gaze pose to be used for interaction. the spec explicitly allows smoothing this. on later version of Horizon OS, this has been an issue as Meta has added smoothing to the poses.

since we only get a single "aim" pose, we also have to calculate eye convergence at a fixed distance.

`XR_META_foveation_eye_tracked` on the other hand, while far from our ideal use-case, seems to be gradually becoming the industry standard with Pico and AndroidXR as well as Steam Frame supporting it. That leaves only HTC (not supported) and Play-for-Dream (unknown) from the well-known brands.

what we need to do in order to receive updates to the `foveationCenter`:
- enable required Vulkan & OpenXR extensions
- apply the foveation profile to the xrSwapchain
- call xrUpdateSwapchainFB after applying the profile
- when calling xrEnumerateSwapchainImages, also enumerate the fragment density map

the fragment density map is then ignored, and we render full res to the swapchain's color images as per usual

the old `XR_EXT_eye_gaze_interaction` path is still used on HMDs that don't support this.

tested on Quest Pro


client/scenes/stream_tracking.cpp
```diff
@@ -451,6 +453,16 @@ void scenes::stream::tracking()
...
+							if (forward_foveation_center)
+							{
+								auto latest = latest_foveation_center.lock();
+								if (latest->dirty)
+								{
+									tracking.foveation_center = latest->center;
+									latest->dirty = false;
+								}
+							}
+
```
xytovl, line 465: If all eye-tracked headsets have the meta extension, we should remove the eye_gaze from foveation path and just use it as normal interaction.
tracking_control will then be able to enable/disable it according to application use.

foveation center updates would be forwarded unconditionally and be the only source of eye-tracked foveation on server side.
Foveation center is not tied to display times, the last received version is used, it might even be out of the tracking thread itself, as it will be produced by the render thread

galister: i would leave it in because while the meta extension is widely supported on standalone headsets that target gaming use, there are definitely headsets that either do not support it due to them being HTC, or do not support it because they don't care about games. however, virtually all eye-tracked standalone headsets support eye gaze, and meta is the only case that i know where it gets smoothed. i think there's no harm leaving it in for now, especially because we know that it works. even when it's smoothed, the user may have to switch from 70% to 50% but it's still better than fixed foveation.

client/xr/session.cpp
```diff
@@ -60,6 +60,93 @@ xr::session::session(xr::instance & inst, xr::system & sys, vk::raii::Instance &
...
+		xrGetFoveationEyeTrackedStateMETA = inst.get_proc<PFN_xrGetFoveationEyeTrackedStateMETA>("xrGetFoveationEyeTrackedStateMETA");
+	}
+}
+
+bool xr::session::supports_foveation_center() const
+{
+	return xrCreateFoveationProfileFB and xrDestroyFoveationProfileFB and xrUpdateSwapchainFB and xrGetFoveationEyeTrackedStateMETA;
+}
+
+std::optional<std::array<XrVector2f, 2>> xr::session::get_foveation_center(XrSwapchain swapchain) const
```
xytovl, line 84: any reason to have this method on the session and not the swapchain?

xytovl: It looks like we could have the foveation profile on the render swapchain (as it was supported until f32d6f47109d5f896d59c9425023b02da6544a56)

server/compositor/foveation.cpp
```diff
@@ -369,9 +369,20 @@ void foveation::compute_params()
...
 		if (foveated_size.width < extent_w)
 		{
-			auto distance = manual_foveation.enabled ? manual_foveation.distance : convergence_distance;
-			auto angle_x = convergence_angle(distance, eye_x[i], -e.x);
-			auto center = angles_to_center(angle_x, fov.angle_left, fov.angle_right);
+			float center;
+			if (use_foveation_center and not manual_foveation.enabled)
+			{
+				if (has_foveation_center)
+					center = std::clamp(foveation_center[i].x * 2 - 1, -1.0f, 1.0f);
```
xytovl, line 376: What happens with cropped fov? (when using passthrough on wayvr)

galister: did not notice anything different just by using wayvr in passthrough, any edge cases that you can think of that I should test?

client/xr/session.cpp
galister: not sure how that would play out, since we'd be submitting the frames at a different foveation center than what we're querying due to the delay

galister: i can move it to the swapchain if that makes more sense. though it's not really doing any actual swapchain work and all.

server/compositor/foveation.cpp
xytovl: If we don't render the full fov, the center must be calculated on the actual fov. How could this work?

client/xr/session.cpp
xytovl: We have calls to `xrCreateFoveationProfileFB` (and destroy) each time we query the center, this sort of function may not be trivial and I would prefer having a handle somewhere that we query when we want the data.

client/scenes/stream_tracking.cpp
xytovl: If we have 2 code path where one is only used on HTC, we know it will bit rot.

Even if we maintain the 2 sources of data, now that the client knows about eye-tracked foveation, the eye_gaze pose should use the normal scheduling mechanism in tracking_control.
Either have the server side decide if it should request it, or inject it in stream_tracking when we must use a fallback mechanism.
This would remove the eye_gaze polling when not actually used

server/compositor/foveation.cpp
galister: let me test again

client/scenes/stream_tracking.cpp
galister: did we ever get an extension list from the PlayForDream people? that's also a headset that might not be prominent in LVRA but it's been selling out over and over since it came out, and it's a quite new headset as well.

server/compositor/foveation.cpp
galister: ok right i didn't notice it because the screen covers most of my view, but it does throw the foveation center off noticeably.

client/scenes/stream_tracking.cpp
xytovl: The way we might do it: switch the network protocol to foveation center only, on headset have code to synthesise center from gaze.
We can then test with the extension blacklist mechanism in hmd_traits.

This makes the paths rejoin much earlier in the pipeline, and the server doesn't need to know about it

server/compositor/foveation.cpp
xytovl: The `XR_META_foveation_eye_tracked` extensions states
> foveationCenter is the center of the foveal region defined in NDC space in the range of -1 to 1 for both eyes.

I suppose this assumes the FoV is what `xrLocateViews` returns, and it could vary (for instance with ipd adjustment). We have to recompute the center knowing both field of views.

galister:
resolved all prior convos due to major rework.

we now forward the view angles in radians. this is unified for both eye gaze, meta foveation as well as fixed. update is only sent if dirty. this is used for foveation only, poses for xr client apps still use forwarded pose from device::EYE_GAZE which is backed by eye_gaze_interaction.

foveation overrides are now also handled on the client.

~~however i'm still struggling to get the cropped FOV logic to work. no matter how I look at it, what we're sending in the packet should match the values of angle_x and angle_y on master, and the fact that non-cropped works flawlessly makes me think that the client-side math is fine. anything that you can think of here?~~

galister:
found the issues with the FOV cropping, there seems to be a number of factors coming together in an unfortunate :hankey: :tornado: 

① the foveationCenter that Meta gives us is not relative to the HMD's full FOV range but relative to the FOV range of the last projection layers that were submitted for that eye.
② meta also clamps the foveationCenter to [-0.5, 0.5] which makes some sense for full FOV, but this means that when we're submitting fov-cropped projection layers, the foveationCenter won't reach the edges properly but stop before reaching the edge.

illustration of hankey tornado

there are 2 solutions i can think of right off the bat, though neither of them are great:
Ⓐ include some padding when cropping the FOV
Ⓑ reproject the cropped layer to a full-size one and submit that instead

edit: I'm not sure if other implementers also handle it like ① or not. the spec is vague here, just mentioning that "foveationCenter is a NDC" which is not really defined formally, just mentioned in the motion vectors part of the spec. Ⓑ might be a safer solution with less surprises down the road.

xytovl:
Option B has a major drawback, we would need a massive swapchain if the useful part is only a small portion of the total.

The specification doesn't state what the foveation center actually is, every runtime may have their own interpretation.

galister:
as Ⓒ we could have a full-size layer backed by a tiny swapchain for the purposes of sampling the NDC. it would only need to be active when in cropped mode. and this I would expect to have the same portability advantages as Ⓑ

galister:
## Behavior as per current state of PR

It behaves perfectly until we run into the clamping action around the edges.

https://github.com/user-attachments/assets/4c5dc201-443c-4931-8753-eb1cb8957579

- tint is applied to max res area via server's foveation.comp
- mouse is driven by XR_EXT_eye_gaze_interaction

## Viability of solution Ⓑ - oversized full-fov swapchain when cropped

I expect this solution to be portable between vendors because we're always at full FOV. When switching to cropped, the swapchain gets re-created at a higher resolution.

the CPU+GPU times heavily depended on how much of the FOV is covered by the cropped layer. I tried my best to record the worst values that I was seeing.

| Test | Passthru | CPU | GPU | Latency | Framerate |
|------|----------|-----|-----|---------|-----------|
| wayvr+helloxr (no crop, regular res) | :x: | 1.5ms | 3.5ms | 43–50ms | stable |
| wayvr+skybox (no crop, regular res) | :x: | 1.5ms | 3.5ms | 43–50ms | stable |
| wayvr (crop, 180% res @ 3240×3564) | :x: | 1.5ms | 6ms | 49–55ms | stable |
| wayvr (crop, 200% res @ 3600×3960) | :x: | 1.5ms | 9ms | 52–60ms | stable |
| wayvr (crop, 240% res @ 4320×4752) | :x: | 3–7ms | 9–10ms | 55–64ms | unstable |
| wayvr (control, old code) | :x: | 1.5ms | 3.5ms | 43–50ms | stable |
| wayvr (crop, 180% res @ 3240×3564) | :heavy_check_mark: | 14ms | 14ms | 66ms | unstable |
| wayvr (crop, 200% res @ 3600×3960) | :heavy_check_mark: | 17ms | 17ms | 68–74ms | unstable |
| wayvr (crop, 240% res @ 4320×4752) | :heavy_check_mark: | 26ms | 24ms | 76–82ms | unstable |
| wayvr (control, old code) | :heavy_check_mark: | 4–7ms | 8–9ms | 59–66ms | unstable |

there is minimal visual difference between 180% and 240%, but even 240% was noticeably fuzzier than control.

## Viability of solution Ⓒ - dummy swapchain

I also expect this to be portable between vendors.

Requirements on Meta hardware:
- swapchain must be actually submitted every frame, else `foveationCenter` won't update
- swapchain must be large enough (64x64 didn't work)


| Test | Passthru | CPU | GPU | Latency | Framerate |
|------|----------|-----|-----|---------|-----------|
| No dummy swapchain | :x: | 1.5ms | 3.5ms | 43–50ms | stable |
| 512² dummy swapchain | :x: | 2ms | 7ms | 50–55ms | stable |
| No dummy swapchain | :heavy_check_mark: | 3~7ms | 8~9ms | 56–60ms | unstable |
| 512² dummy swapchain | :heavy_check_mark: | 3~9ms | 8~12ms | 53–62ms | unstable |

On the Quest Pro, enabling passthru will generally completely wreck performance. due to system crashes, Meta has reduced the passthru video feed from ~60fps down to ~15. it doesn't cause the whole headset system to crash anymore, but it sure is still laggy.

This is solution is probably less invasive than I've initially thought. The performance impact is quite large when not using passthru (no fast-path, probably?). However the impact is not as bad when passthru is enabled. Passthru itself seems to be a way bigger contributor to slowness.

This is seems acceptable for our use-case, because either:
- In a scene app → don't need dummy swapchain → no impact
- Not in a scene app, passthru is on → impact is small enough to not matter
- Not in a scene app, passthru is off → impact on GPU time, but still stable @ max supported refresh rate

## Test environment & params

<table>
  <tr><th>Headset</th><td>Quest Pro</td></tr>
  <tr><th>Refresh rate</th><td>90fps</td></tr>
  <tr><th>Stream resolution</th><td>140%</td></tr>
  <tr><th>Foveation</th><td>70%</td></tr>
  <tr><th>Video</th><td>80mbit HEVC 10-bit</td></tr>
  <tr><th>Encoder</th><td>Vulkan</td></tr>
  <tr><th>High-power mode</th><td>OFF</td></tr>
</table>

client/xr/session.h
```diff
@@ -21,6 +21,7 @@
 
 #include "utils/handle.h"
 #include "utils/thread_safe.h"
+#include <array>
```
xytovl, line 24: this is no longer needed

client/foveation.cpp
```diff
@@ -0,0 +1,286 @@
...
+		        .type = XR_TYPE_FOVEATION_PROFILE_CREATE_INFO_FB,
+		        .next = &level_info,
+		};
+
+		XrFoveationProfileFB raw_profile = XR_NULL_HANDLE;
+		XrResult profile_result = xrCreateFoveationProfileFB(session, &profile_info, &raw_profile);
+		if (XR_FAILED(profile_result))
+			return {};
+
+		foveation_profile.emplace(raw_profile, xrDestroyFoveationProfileFB);
```
xytovl, line 123: I don't think we need lazy creation of the profile, just move that to the constructor

client/foveation.cpp
```diff
@@ -0,0 +1,286 @@
...
+	if (XR_FAILED(state_result))
+		return {};
+
+	if (!(state.flags & XR_FOVEATION_EYE_TRACKED_STATE_VALID_BIT_META))
+		return {};
+
+	return std::array<XrVector2f, 2>{state.foveationCenter[0], state.foveationCenter[1]};
+}
+
+angles center_to_angles(const std::array<XrVector2f, 2> & center, std::span<const XrFovf> fovs)
```
xytovl, line 152: This function may be static/in an anonymous namespace, and the get_foveation_center one return angles directly.

common/wivrn_packets.h
```diff
@@ -405,6 +406,7 @@ struct tracking
 
 	std::array<view, 2> views;
 	std::vector<pose> device_poses;
+	std::optional<std::array<XrVector2f, 2>> foveation_angles; // Per-eye x/y view angles in radians
```
xytovl, line 409: The nominal flow now is to have the updated value in the render thread, once per frame.

I think it will be easier to move the foveation center update to a separate packet and push the update from the place where it is polled directly.
This would remove thread synchronization, and the last remaining interaction between tracking and foveation.

For eye gaze based foveation, we can just poll the pose where we would poll the foveation center

